PR #25, opened 2024-07-16
Update README.md

rachel-fenichel:


BeksOmega (approved):
Can you fix the last line in the file while you're here?

PR #26, opened 2024-07-17
feat: add p5 blocks and execution

rachel-fenichel:
This PR adds blocks and an execution environment for drawing with p5.js. The goal is to make the test page easier to use for questions like "how would you change the color of the circle"?

The updated test page, with a drawing area on the left

Code can be run with the Shift + R keyboard shortcut.

Blocks and some other code were copied from https://github.com/google/blockly-samples/pull/2085/files

I suggest reviewing commit-wise because I ran the formatter in the middle and it touched a lot of files.


src/blocks/p5_blocks.js
```diff
@@ -0,0 +1,177 @@
+import * as Blockly from 'blockly/core';
```
BeksOmega, line 1: License here and in any other files.

src/blocks/toolbox.js
```diff
@@ -0,0 +1,135 @@
+
+export const toolbox = {
```
BeksOmega, line 1: Can we use the normal toolbox and just add another category for the P5 blocks?

test/index.ts
```diff
@@ -5,8 +5,35 @@
...
+// @ts-ignore
+import {save, load} from './serialization';
+
+function runCode(workspace: Blockly.WorkspaceSvg) : void {
+  const codeDiv = document.getElementById('generatedCode')!.firstChild;
+  const code = javascriptGenerator.workspaceToCode(workspace);
+  (codeDiv as any).innerText = code;
+  const p5outputDiv = document.getElementById('p5output');
+  p5outputDiv!.innerHTML = '';
+  console.log('running code');
```
BeksOmega: Do you need this log still?

test/serialization.js
```diff
@@ -0,0 +1,60 @@
...
+  const data = Blockly.serialization.workspaces.save(workspace);
+  window.localStorage?.setItem(storageKey, JSON.stringify(data));
+};
+
+const defaultData = {
+  'blocks': {
+    'languageVersion': 0,
+    'blocks': [
+      {
+        'type': 'p5_setup',
```
BeksOmega: Is it actually helpful to have these things by default? Or does that mess with testing? I think it would be better to have a button that loads some state rather than saying you must always have this state. This makes it kind of inflexible for researchers.

test/serialization.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param {Blockly.Workspace} workspace Blockly workspace to load into.
+ * @returns
+ */
+export const load = function(workspace) {
+  const data = window.localStorage?.getItem(storageKey) ||
+  JSON.stringify(defaultData);
+
+  // Don't emit events during loading.
+  Blockly.Events.disable();
+  Blockly.serialization.workspaces.load(JSON.parse(data), workspace, false);
```
BeksOmega: Similarly, don't know that we necessarily want to persist the state between refreshes because that might mess with testing.

test/index.ts
```diff
@@ -20,19 +20,23 @@ import {javascriptGenerator} from 'blockly/javascript';
...
-    eval(code);
-  }, p5outputDiv!);
+  if (p5outputDiv) {
+    // Clear the old canvas.
+    p5outputDiv.innerHTML = '';
+    // Run P5 in instance mode. The name 'sketch' matches the name used
+    // in the generator for all of the p5 blocks.
+    // eslint-disable-next-line new-cap
+    new p5((sketch) => {
+      console.log(sketch);
```
BeksOmega: Do you still need this log?

src/blocks/toolbox.js
rachel-fenichel: I would like to stick to a non-category toolbox while we try to figure out how to do basic navigation within a stack and within a workspace. I think the workflow for categories in toolboxes is fairly standard and adding it in just adds complexity to the basic test scenarios.

I'm targeting the simple use case (up to ten blocks on a workspace, up to three stacks, and no categories in the toolbox) for the intro sequences that will use this first.

test/index.ts
rachel-fenichel: Removed.

test/serialization.js
rachel-fenichel: I actually removed this already, but there is still state persistence because we're using `createPlayground` from `@blockly/dev-tools`.

I could remove `createPlayground` entirely, since we don't want to use the drawer full of options at this point. That would also remove the state persistence and a dependency on dev-tools. What do you think?

test/index.ts
rachel-fenichel: Removed.

test/serialization.js
rachel-fenichel: I think it should start with *some* state by default, because it reduces the work needed for a test user to start using Blockly--no one has to come through and set up the blocks for them.

I am interested in having buttons that let you load a few different scenarios, with instructions on how to interact with those scenarios (adding blocks vs editing blocks, etc). I thought that could be a separate feature.

test/serialization.js
BeksOmega: Sgtm, but can be done as a follow up.

BeksOmega (approved):

PR #33, opened 2024-07-23
feat: add application role to Blockly workspace div

### LINKED ISSUE #27: Make it possible to jump to the workspace using only the keyboard
## Expected Behavior

When a user loads the page they should be able to get to the workspace using only the keyboard.

## Actual Behavior

The user has to click on the workspace, at which point they can't start using keyboard navigation.

## Proposed solution

* [x] #428
* [x] #71
* [ ] Document how to set focus on Blockly workspace compatibly with tabbing behaviour.

rachel-fenichel:
Fixes #27 

This makes a few changes:
- Add the application role and a relevant label to the injection div.
- Set the tabindex to 0 to ensure that the workspace is in the focus ring.
- Add a handler to focus the workspace when tab focus hits it.

I tested this by deploying it to my own github pages site, and then using NVDA on windows to move around the page.

### DIFF
--- a/src/extended_navigation_controller.js
+++ b/src/extended_navigation_controller.js
@@ -301,6 +301,17 @@ export class ExtendedNavigationController extends NavigationController {
   }
 }
 
+function addAriaInfo(workspace, navController) {
+  const injectionDiv = workspace.getInjectionDiv();
+  injectionDiv.setAttribute('role', 'application');
+  injectionDiv.setAttribute('aria-label', 'Blockly workspace');
+  injectionDiv.setAttribute('tabindex', '0');
+  // Set focus when entering the region.
+  injectionDiv.addEventListener('focusin', (event) => {
+    navController.navigation.focusWorkspace(workspace);
+  })
+}
+
 export function installNavController(workspace) {
   const navigationController = new ExtendedNavigationController();
   navigationController.init();
@@ -308,4 +319,6 @@ export function installNavController(workspace) {
   // Turns on keyboard navigation.
   navigationController.enable(workspace);
   navigationController.listShortcuts();
+
+  addAriaInfo(workspace, navigationController);
 }


src/extended_navigation_controller.js
```diff
@@ -301,11 +301,24 @@ export class ExtendedNavigationController extends NavigationController {
   }
 }
 
+function addAriaInfo(workspace, navController) {
+  const injectionDiv = workspace.getInjectionDiv();
+  injectionDiv.setAttribute('role', 'application');
+  injectionDiv.setAttribute('aria-label', 'Blockly workspace');
```
BeksOmega, line 307: We probably don't want to say Blockly here if this is user-facing. (Maybe "Code workspace" instead, or ask Josh what they call the code area in CSFirst) You don't need to change it now but maybe file an issue to fix it before releasing.

src/extended_navigation_controller.js
```diff
@@ -301,11 +301,24 @@ export class ExtendedNavigationController extends NavigationController {
...
+  injectionDiv.setAttribute('tabindex', '0');
+  // Set focus when entering the region.
+  injectionDiv.addEventListener('focusin', (event) => {
```
BeksOmega, line 310: Just a Q: How does this interact with 1) Clicking on the workspace (does that fire a focus event) 2) modifying text input fields, since the text input is outside the workspace div?

BeksOmega (approved):

src/extended_navigation_controller.js
rachel-fenichel: 1) clicking inside the workspace also fires a focus event, so it flips the cursor back to the top level of the workspace. I think this is part of the discussion in https://github.com/google/blockly-keyboard-experimentation/discussions/11

src/extended_navigation_controller.js
rachel-fenichel: Good point. Do you want that to be a bigger thing about how to customize details like that, or just a simple issue for now and we can return to it as we pile up more related configuration details?

src/extended_navigation_controller.js
rachel-fenichel: 2. This does not do what I want. Back to the drawing board.

src/extended_navigation_controller.js
BeksOmega: I think just a simple issue is good!

PR #36, opened 2024-07-25
fix: copy over keyboard navigation code and remove plugin dependency

### LINKED ISSUE #35: Copy over keyboard navigation plugin code and remove plugin dependency
When I started this project I extended the classes in the keyboard-navigation plugin, for simplicity. But that also means that the code is split across two repositories, which makes it hard to modify. 

The relevant parts of the keyboard navigation source should be copied in, and the dependency should be deleted.

rachel-fenichel:
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/35

Should be reviewed commitwise for readability. What this PR does:
- I started by copying over the basic keyboard navigation code. Then I moved code from my `extended_navigation_controller` and `extended_line_cursor` into the base classes, cleaning up as I went. This avoids the churn of creating shortcuts and then reassigning them.
- Once all of the code was consolidated, I moved all of the base files up to the top level of `src`.
- I created an `announcer` class for interacting with the announcer region on the test page.
- Deleted field handler monkeypatching that was only needed because we were using WASD instead of arrow keys.
- Did cleanup (linting, formatting, and adding jsdoc).

As a follow-up I want to convert this code to TypeScript so we get the type hinting benefits.


src/base_keyboard_nav/navigation_controller.js
```diff
@@ -12,7 +12,8 @@
 
 import './gesture_monkey_patch';
 
-import * as Blockly from 'blockly/core';
+import {ASTNode, ShortcutRegistry} from 'blockly';
```
BeksOmega: I'm curious why you changed things over to use named imports. Was it to save characters?

Either way you should import from `blockly/core` not `blockly`! The `blockly` export includes the JSGenerator and bloats the package.

src/navigation_controller.js
```diff
@@ -61,14 +61,6 @@ export class NavigationController {
    * @protected
    */
   addShortcutHandlers() {
-    if (Blockly.FieldColour) {
```
BeksOmega: Why are these unnecessary?

BeksOmega (approved):

src/navigation_controller.js
rachel-fenichel: These fields already have code for handling arrow keys, and it all does the correct things. The reason we needed extra code was because we were using something _other than_ arrow keys (WASD).

rachel-fenichel: We may need something like this if/when we allow remapping to other key sets. I think trying to design for that right now would be premature. Since it would also require monkey-patching, I'd rather just delete this code.

src/base_keyboard_nav/navigation_controller.js
rachel-fenichel: I used named imports from the beginning in the code that I wrote, and changed things over in this PR for consistency across the code.

The main motivation was to save characters in code like this:
```
    ShortcutRegistry.registry.register(previousShortcut);
    ShortcutRegistry.registry.addKeyMapping(
      BlocklyUtils.KeyCodes.UP,
      previousShortcut.name,
    );
    ```

rachel-fenichel: I'll fix the imports!

Issue #37, opened 2024-07-25
Convert code to typescript


rachel-fenichel:
Our current code is a mix of js and ts. Converting to TS will give us better type information and editor support.

PR #41, opened 2024-08-08
chore: convert navigation controller to typescript

rachel-fenichel:
Part of #37 

Continues converting code to typescript and identifying errors exposed by adding types.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -13,7 +13,7 @@
 import './gesture_monkey_patch';
 
 import * as Blockly from 'blockly/core';
-import {ASTNode, ShortcutRegistry} from 'blockly/core';
+import {ASTNode, ShortcutRegistry, BlockSvg, WorkspaceSvg} from 'blockly/core';
 import {utils as BlocklyUtils} from 'blockly/core';
 
 import * as Constants from './constants';
@@ -30,7 +30,7 @@ export class NavigationController {
   copyData: BlockCopyData | null = null;
 
   /** The workspace a copy or cut keyboard shortcut happened in. */
-  copyWorkspace: Blockly.WorkspaceSvg | null = null;
+  copyWorkspace: WorkspaceSvg | null = null;
   navigation: Navigation;
   announcer: Announcer;
 
@@ -111,7 +111,7 @@ export class NavigationController {
    * @param workspace The workspace to add keyboard
    *     navigation to.
    */
-  addWorkspace(workspace: Blockly.WorkspaceSvg) {
+  addWorkspace(workspace: WorkspaceSvg) {
     this.navigation.addWorkspace(workspace);
   }
 
@@ -122,7 +122,7 @@ export class NavigationController {
    * @param workspace The workspace to remove keyboard
    *     navigation from.
    */
-  removeWorkspace(workspace: Blockly.WorkspaceSvg) {
+  removeWorkspace(workspace: WorkspaceSvg) {
     this.navigation.removeWorkspace(workspace);
   }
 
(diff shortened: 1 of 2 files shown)


src/navigation_controller.ts
```diff
@@ -12,43 +12,43 @@
...
-import {ASTNode, ShortcutRegistry} from 'blockly/core';
+import * as Blockly from 'blockly/core';
+import {ASTNode, ShortcutRegistry, BlockSvg, WorkspaceSvg} from 'blockly/core';
 import {utils as BlocklyUtils} from 'blockly/core';
 
 import * as Constants from './constants';
 import {Navigation} from './navigation';
 import {Announcer} from './announcer';
+import {LineCursor} from './line_cursor';
+import {BlockCopyData} from 'node_modules/blockly/core/clipboard/block_paster';
```
BeksOmega, line 24: Can you add a TODO with a tag for the issue you filed?

src/navigation_controller.ts
```diff
@@ -12,43 +12,43 @@
...
-  constructor(optNavigation) {
+  constructor() {
     /**
      * Handles any keyboard navigation shortcuts.
-     * @type {!Navigation}
-     * @public
      */
-    this.navigation = optNavigation || new Navigation();
+    this.navigation = new Navigation();
     this.announcer = new Announcer();
```
BeksOmega: Can these be constructed in the property declarations (instead of the constructor) actually?

src/navigation_controller.ts
```diff
@@ -69,37 +68,36 @@ export class NavigationController {
...
+    this: Blockly.Toolbox,
+    shortcut: ShortcutRegistry.KeyboardShortcut,
+  ): boolean {
     if (!this.selectedItem_) {
       return false;
     }
     switch (shortcut.name) {
       case Constants.SHORTCUT_NAMES.PREVIOUS:
-        return this.selectPrevious_();
+        return (this as any).selectPrevious_();
```
BeksOmega, line 82: Might have missed this in previous PRs, but do you want to set up an `AnyDuringMigration` alias? Not sure if there are places we actually want to be using `any` vs places where we're just doing it temporarily.

src/navigation_controller.ts
```diff
@@ -109,96 +107,95 @@ export class NavigationController {
...
     const cursor = workspace.getCursor();
     if (!cursor || !cursor.getCurNode()) {
       return false;
     }
     const curNode = cursor.getCurNode();
     if (curNode.getType() === ASTNode.types.FIELD) {
-      return /** @type {!Blockly.Field} */ (curNode.getLocation()).onShortcut(
-        shortcut,
-      );
+      return (curNode.getLocation() as Blockly.Field).onShortcut(shortcut);
```
BeksOmega, line 156: In the future we could probably switch this to use a type guard so you don't have to cast. But might be better to just have these be actual subclasses or generics.

BeksOmega (approved):

src/navigation_controller.ts
rachel-fenichel: The linter warns about uses of `any` so I'm going to rely on that instead of an `AnyDuringMigration` until I'm convinced that we have legitimate uses of `any` instead of just patching over issues.

src/navigation_controller.ts
rachel-fenichel: Yeah, this problem is what made me consider pulling over all of the AST code as well. I think there are typescript-y ways to get the types to behave right on this without casts, and it's a common pattern.

src/navigation_controller.ts
rachel-fenichel: Done

src/navigation_controller.ts
rachel-fenichel: Done

PR #51, opened 2024-09-06
feat: add draw_emoji block

rachel-fenichel:
For testing purposes.

Simple block with a dropdown that draws an emoji in the middle-ish of the canvas.

image


cpcallen (approved):
🐻

PR #61, opened 2024-09-30
refactor: Refactor `KeyboardShortucts` to simplify `NavigationController`

cpcallen:
A few changes:

* Make use of feature allowing you to add a key mapping at the time a `KeyboardShortcut` is registered.
* Put all shortcuts in a dictionary object, to simplify registration and unregistration.  In future, different dictionaries can be used for different modes.
* Make JSDoc comments more concise (and in several cases more accurate).
* Preserve the original `Toolbox.prototype.onShortcut` method when monkeypatching.
* Use `Array.prototype.includes` instead of `.indexOf` when approriate.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -178,11 +178,20 @@ export class NavigationController {
   }
 
   /**
-   * Keyboard shortcut to go to the previous location when in keyboard
-   * navigation mode.
+   * List all the currently registered shortcuts.
    */
-  protected registerPrevious() {
-    const previousShortcut: ShortcutRegistry.KeyboardShortcut = {
+  listShortcuts() {
+    this.announcer.listShortcuts();
+  }
+
+  /**
+   * Dictionary of KeyboardShortcuts.
+   */
+  protected shortcuts: {
+    [name: string]: ShortcutRegistry.KeyboardShortcut;
+  } = {
+    /** Go to the previous location. */
+    previous:  {
       name: Constants.SHORTCUT_NAMES.PREVIOUS,
       preconditionFn: (workspace) => {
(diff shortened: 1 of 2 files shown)


src/navigation_controller.ts
```diff
@@ -44,6 +45,14 @@ export class NavigationController {
   navigation: Navigation = new Navigation();
   announcer: Announcer = new Announcer();
 
+  /**
+   * Original Toolbox.prototype.onShortcut method, saved by
+   * addShortcutHandlers.
+   */
+  private origToolboxOnShortcut:
+    | typeof Blockly.Toolbox.prototype.onShortcut
```
rachel-fenichel, line 53: Why the leading `|`?

src/navigation_controller.ts
```diff
@@ -53,23 +62,24 @@ export class NavigationController {
   }
 
   /**
-   * Adds methods to core Blockly components that allows them to handle keyboard
-   * shortcuts when in keyboard navigation mode.
+   * Monkeypatches core Blockly components to add methods that allow
```
rachel-fenichel, line 65: Historical note: this used to also monkeypatch some fields. I deleted that during cleanup, leaving only the toolbox monkeypatching.

rachel-fenichel (approved):
Improved readability ftw!

src/navigation_controller.ts
cpcallen: Thanks for this info.

src/navigation_controller.ts
cpcallen: No idea.  Ask the Prettier developers.  ¯\\\_(ツ)_/¯

PR #70, opened 2024-10-16
feat: make draw code only run once

### LINKED ISSUE #69: Confusing behaviour with run button and random colour block
## Expected Behaviour

Either colour changes once each time I click run button, or the code runs periodically but I don't have to press run at all.

## Actual Behaviour

When you hit run the colour starts changing rapidly.

1.
1.
1.

rachel-fenichel:
Fixes #69 

Uses [`noLoop`](https://p5js.org/reference/p5/noLoop/) to make `draw` only run once. This keeps the generated code in a standard p5 structure, which means it's easy to change it back to running repeatedly for animations.

It would be simple to just move all of the code into the `setup` block instead of separating into `setup` and `draw`, but even simple micro:bit programs often have multiple block stacks.


BenHenning (approved):
Seems like a reasonable fix for #69. Thanks!

PR #72, opened 2024-10-18
feat: make Blockly cooperative with tab navigation

### LINKED ISSUE #71: Users should be able to navigate to/away and within/outside Blockly using tabs
## Expected Behavior
Users should be able to focus Blockly by navigating to it via tabs and otherwise it shouldn't receive keyboard inputs. Similarly, when Blockly has focus, users shouldn't be able to navigate to other elements.

As discussed in #63 the initial idea is to:
- Disable keyboard navigation until the user navigates to Blockly via tab.
- Require explicit user input to enable keyboard navigation via the 'enter' key.
- Disallow tab navigation while keyboard navigation is enabled.
- Allow the user to exit keyboard navigation using 'escape'.

## Actual Behavior
Currently, users are able to navigate independently of whether Blockly shortcuts are being used. This is inherently confusing: a user can focus the drop-down on the test page while simultaneously trying to use arrow keys to navigate Blockly elements (since the plugin always takes all keyboard input).

1. Load the the latest keyboard plugin test environment (e.g. ``npm start run`` on ad213f5a6094a82bd344352acf83001d5386165f).
2. (Optionally) enable a screen reader like Chromevox (not necessary since one can observe element highlighting for current tab focus).
3. Navigate a local browser tab to the test environment (e.g. ``http://localhost:8080/``).
4. Try to navigate around the screen with tab while also interacting with the Blockly environment using tab navigation.

BenHenning:
Fixes #71

This introduces more intentional tab navigation support in Blockly when using the keyboard navigation plugin, specifically by:
- Ensuring the top-level container is only in the tab index once. Without these changes, Blockly's top div and top group are both in the root tab index (which means both are tabbed to when trying to use tab navigation which can be confusing to the user since there's no feedback that one of the tabs is going 'nowhere').
- Introducing explicit 'enter' and 'escape' flows for giving Blockly keyboard focus (and taking away focus from the browser), otherwise the browser retains tab-based navigation.

See the following videos for demonstrations on the before/after behavior with these changes:

Before changes:

[before_without_reader.webm](https://github.com/user-attachments/assets/b4e8ea46-7f54-4237-9cf5-23841e6ca455)

After changes:

[after_without_reader.webm](https://github.com/user-attachments/assets/89df98f0-bd9a-4f6e-997b-e16ad933b044)

Note that there's still a "double tab navigation" issue for the main Blockly injection area with a screen reader (but not for normal navigation). This will require additional investigation outside this PR. The before/after behavior with the screen reader isn't included above since it's hard to see the results in the recording (the highlights and output of the reader aren't being captured).

Note also that this PR also includes a few cleanups:
- Some syntax simplification when using lambdas.
- Some constant renames/additions for consistency (and correctness in disposing of registered shortcuts).

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -28,7 +28,7 @@ export enum SHORTCUT_NAMES {
   IN = 'in',
   OUT = 'out',
   INSERT = 'insert',
-  MARK = 'mark',
+  ENTER_OR_MARK = 'enter_or_mark',
   DISCONNECT = 'disconnect',
   TOOLBOX = 'toolbox',
   EXIT = 'exit',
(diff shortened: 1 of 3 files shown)


BenHenning:
PTAL @rachel-fenichel.

rachel-fenichel (approved):

BenHenning (commented):
Self-reviewed changes post merge. @rachel-fenichel can you PTAL just for a quick glance?

rachel-fenichel (approved):

Issue #67, opened 2024-10-08
No focus indication (cursor) when focused on colour field


cpcallen:
## Expected Behaviour

When you move the cursor to a colour picker field, the cursor remains visible.

## Actual Behaviour

The cursor disappears when you move it to a colour picker field.

## Steps to Reproduce the Problem

1.
1.
1.

microbit-lucy:
It would be great to have the same orange border around the colour field to show it's selected. Do you think this will be possible in time for 28th October?

microbit-matt-hillsdon:
I think this may have been fixed for some time. RaspberryPiFoundation/blockly-samples#2722 tracks a related contrast issue that's particularly prominent with the colour block.

PR #79, opened 2024-10-28
fix: put highlight on shadow block when a colour field is selected

rachel-fenichel:
Works on [#https://github.com/google/blockly-keyboard-experimentation/issues/67](https://github.com/google/blockly-keyboard-experimentation/issues/67)

New: 
image

Old:
image

### Implementation notes
- If the cursor is on a block, hide the drawer and select the block.
- If the cursor is on a colour field, hide the drawer and call `addSelect` on the colour field's source block. Calling `setSelected` does not work in this case because selecting a shadow block automatically selects the parent.
- If the cursor was previously on a colour field, call `removeSelect` on the old colour field.
- If the cursor was previously on a block and is no longer on a block, call `setSelected(null)` to remove the old highlight. If it's on a block, calling `setSelected(block)` automatically removes the old selection.

### Notes

- I am not proud of this code. It requires cleanup.
- I'm running into an issue where the second colour field will let me open the editor, but won't respond to keypresses to select within the editor. I think this is unrelated.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -13,6 +13,7 @@
  * @author aschmiedt@google.com (Abby Schmiedt)
  */
 
+import { FieldColour } from '@blockly/field-colour';
 import * as Blockly from 'blockly/core';
 import {ASTNode, Marker} from 'blockly/core';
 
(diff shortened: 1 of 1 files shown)


src/line_cursor.ts
```diff
@@ -424,15 +425,44 @@ export class LineCursor extends Marker {
     const oldNode = (this as any).curNode;
     (this as any).curNode = newNode;
     const drawer = (this as any).drawer;
+    if (!drawer) {
+      console.error('could not find a drawer');
+      return;
+    }
+
+    const newNodeIsFieldColour = newNode?.getType() == ASTNode.types.FIELD &&
```
cpcallen, line 433: `===`?

cpcallen (approved):

PR #82, opened 2024-10-29
fix: Remove extra tab stop on XML group

### LINKED ISSUE #80: Extra tab stop before Blockly workspace
In the demo page, there is an extra tab stop between the scenario pop-up and the tab stop on which pressing enter allows you to begin keyboard navigation of Blockly.  It's unclear to me what the purpose of this tab stop is, and it appears to be causing unnecessary difficulty for our research participants.  Can we remove it?

cpcallen:
Ben's PR #72 removed the tabindex on the enclosing DIV, but there's also a tabindex on the XML element that encloses the G that gets the focus/blur event listeners.  Remove it.

An alternative approach would be to add the event listeners to the XML element instead of the G element, and this might be preferable overall but breaks the visual indicator that you have tabbed to the workspace (and I don't know how to fix it) so I'm going with this approach for now.

Fixes #80

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,8 +32,9 @@ export class KeyboardNavigation {
     this.setGlowTheme();
     installCursor(workspace.getMarkerManager());
 
-    // Ensure that only the root SVG group has a tab index.
+    // Ensure that only the root SVG G (group) has a tab index.
     workspace.getInjectionDiv().removeAttribute('tabindex');
+    workspace.getParentSvg().removeAttribute('tabindex');
 
     workspace.getSvgGroup().addEventListener('focus', () => {
       navigationController.setHasFocus(true);


cpcallen:
I'm going to merge this TBR @rachel-fenichel and @BenHenning.

I'd be particularly interested in understanding why #72 added the event listeners to the `<g>` element rather than the `<xml>` element.

BenHenning (commented):
Good find @cpcallen. I definitely missed this. When I was investigating the duplicate tab indexes, I didn't realize that there was another present (or, more precisely, I realized there was a second tab but didn't find it). For some reason, during my testing this only affected navigation when I had a screenreader enabled which is why I didn't try to fix it.

Thanks for the quick fix.

Issue #81, opened 2024-10-28
Make visual indication that blockly workspace is selected more obvious


cpcallen:
When you tab to the Blockly workspace, the outline around the workspace gets slightly more prominent, but some users do not notice this.  Is it possible to make it more prominent?

bmxedd:
I found a solution to adjust the thickness of the highlight via CSS. However I noted that the workspace focus does not work on Firefox with the previous removal of the tabindex on the SVG element. The only elements that are focused are the form elements. Whatever technique that makes Chrome make <g> element a focus target doesn't seem to work.

cpcallen:
This is probably a good one for @BenHenning and @bmxedd to coordinate on, since I know Ben is working on implementing roving tabindex, and Ed is good at CSS.

Maybe we add a class to the Blockly workspace to make it look focused, even though the actual focus will be on a particular block / menu item / etc.?

cpcallen:
Following testing done by Micro:Bit / MakeCode: we should test fixes for this issue against both the simple toolbox and the categories toolbox.

PR #83, opened 2024-10-29
feat: Make workspace focus highlight more visible

cpcallen:
A small hack (kludge?) from @bmxedd to make it a bit more obvious when the workspace is focused.

@bmxedd notes:

    "Many CSS styles don't work on internal SVG elements. This
    at least allows to the outline to be shown and not be
    obscured by the flyout.

    There is something going on with the width of the outline
    to the top and left side that I couldn't figure out."

A temporary, not-quite-adequate fix for #81.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -26,6 +26,18 @@
         width: 100%;
         max-height: 100%;
         position: relative;
+        --outline-width: 3px;
+      }
+
+      .blocklyMainBackground {
+        height: calc(100% - var(--outline-width));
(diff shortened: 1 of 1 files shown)


cpcallen:
I am going to merge this TBR as it should be helpful for getting better feedback from the remainder of today's UX testing sessions.

BenHenning:
@cpcallen might you have a before/after screenshot that you can add here or to the issue? That might be nice to have in order to better illustrate the change.

Issue #102, opened 2024-11-08
Make placing blocks using the marker more intuitive


cpcallen:
This bug concerns the "marker workflow", where the users places a marker somewhere in an existing program, and then goes to the toolbox to select a block to be inserted at the marked location.

In testing, it was observed that users had considerable difficulty using the marker workflow to place blocks:

* They often forgot (or did not know) that they needed to place the marker before choosing a block from the toolbox.
* They expected that pressing `t` to move to the toolbox would (in our terminology) place the marker automatically at their current location.
* They often placed the marker unintentionally while trying to do something else (e.g., place marker on a connection while trying to open a colour picker on the block attached to that connection.
* They did not find the depiction of the marker intuitive, and suggested instead that an insertion marker or whitespace gap should be shown instead.
* They were confused by, or did not understand the significance of having the marker left "lying around" on the workspace, and were not sure how to deactivate it.

Additionally, there are some other less-than-desirable behaviours:

* Selecting a block that can't be placed at the marked location will cause it to be placed as a top-level block at an arbitrary place on the workspace.

I therefore propose the following changes to how the marker is used and behaves (in order of priority to implement):

* [x] Pressing enter on a connection should place the marker there and move the cursor to the toolbox.
* [x] The marker should be transient.
  * It should disappear after a block is placed, or as soon as the cursor is moved to a different part of the workspace.
* [ ] #211
  * the cursor will only visit blocks suitable blocks, or
  * the cursor will visit all blocks, but attempting to select an unsuitable block will display an error message explaining why that block can't be placed at the marked location.
* [x] #212
  * Depending on how this is implemented it might be that we can actually remove the existing marker code completely.

Visiting the toolbox by any other means than pressing enter while on a  connection (e.g. by pressing `t`, or by tabbing to it—see #101) will still allow any block to selected from the toolbox and placed as a new top-level block on the workspace.

kmcnaught:
> Visiting the toolbox by any other means than pressing enter while on a connection (e.g. by pressing t, or by tabbing to it—see https://github.com/google/blockly-keyboard-experimentation/issues/101) will still allow any block to selected from the toolbox and placed as a new top-level block on the workspace.

[...]

PR #115, opened 2024-11-11
feat: open flyout or toolbox when pressing enter on the workspace

rachel-fenichel:
Fixes one part of #102 

Pressing enter when on a connection now marks that connection and opens the toolbox or flyout and moves the cursor there. The user can then navigate to a block in the flyout and press enter again to insert it.

After insertion, the marker still sticks around (even though it should be made transient) so fixing that is a follow up, already tracked in #102.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1233,6 +1233,11 @@ export class Navigation {
       nodeType == Blockly.ASTNode.types.WORKSPACE
     ) {
       this.markAtCursor(workspace);
+      if (!workspace.getToolbox()) {
+        this.focusFlyout(workspace);
+      } else {
+        this.focusToolbox(workspace);
+      }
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -1233,6 +1233,11 @@ export class Navigation {
       nodeType == Blockly.ASTNode.types.WORKSPACE
     ) {
       this.markAtCursor(workspace);
+      if (!workspace.getToolbox()) {
+        this.focusFlyout(workspace);
+      } else {
+        this.focusToolbox(workspace);
+      }
```
cpcallen: Why not:

```suggestion
      if (workspace.getToolbox()) {
        this.focusToolbox(workspace);
      } else {
        this.focusFlyout(workspace);
      }
```

cpcallen (approved):

Issue #103, opened 2024-11-08
Make it possible to place blocks using copy-and-paste


cpcallen:
It should be possible to:
* navigate to the toolbox (e.g. by pressing `t` or in future `tab`)
* use the cut or copy shortcut to choose a block to place,
* navigate to the desired location on the workspace, and
* use the paste shortcut to paste the chosen block at that location.

rachel-fenichel:
This works in the main workspace but does not work when copying from the flyout, which suggests that the handler for copy needs to be added to the flyout workspace.

The handlers for cut and paste do not apply in the flyout workspace, so they should just be ignored.

rachel-fenichel:
Actually, copy also fires on the flyout but this precondition likely fails: 
```
      preconditionFn: (workspace) => {
        if (
          workspace.keyboardAccessibilityMode &&
          !workspace.options.readOnly
        ) {
          const curNode = workspace.getCursor()?.getCurNode();
          if (curNode && curNode.getSourceBlock()) {
            const sourceBlock = curNode.getSourceBlock();
            return !!(
              !Blockly.Gesture.inProgress() &&
              sourceBlock &&
              sourceBlock.isDeletable() &&
              sourceBlock.isMovable()
            );
          }
        }
        return false;
      },
```

rachel-fenichel:
After https://github.com/google/blockly-keyboard-experimentation/pull/120 it is possible to copy a block from the flyout and paste it to the workspace immediately and without moving focus back to the workspace. It pastes in the most recent cursor location, which was marked when opening the toolbox.

@kmcnaught what do you think the behaviour should be here?

kmcnaught:
I think that pasting in the most recent cursor location is good behaviour - it is clear and not surprising, and if it's not what you wanted then you can presumably navigate the cursor elsewhere before pasting. I think this behaviour should work regardless if you copied from the flyout or the toolbox. Is there another flow I'm missing?

kmcnaught:
I note @cpcallen's comment on #120 about paste from the flyout failing - let me chat with the others here tomorrow. 

Regarding @rachel-fenichel's comment about pasting at the old marker position - I presume this is unintended behaviour that won't be a problem once we lose the concept of marker and cursor being in different places.

kmcnaught:
[...]

PR #120, opened 2024-11-12
feat: make it possible to copy a block from the flyout and paste it into the workspace

rachel-fenichel:
Fixes part of https://github.com/google/blockly-keyboard-experimentation/issues/103

The issue is that the shortcut is always passed the main workspace, but that it needs to get the cursor on the flyout workspace (not the main workspace) if that's currently active. I added code to query the correct workspace. Also, blocks in the flyout are not movable, so the other condition failed as well.

I changed paste to check if hte source workspace was a flyout and if so, paste to the main workspace.

This still pastes at the location of the marker, not the cursor. The marker location is set when entering the flyout. If the user copies in the flyout, returns to the main workspace, navigates, and pastes, they will still end up pasting at the site of the marker.

The fix to that issue is part of the general plan to remove or significantly reduce the marker, by making it transient and only active in specific situations. That's outside the scope of this change.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -533,22 +533,40 @@ export class NavigationController {
       name: Constants.SHORTCUT_NAMES.COPY,
       preconditionFn: (workspace) => {
         if (this.canCurrentlyEdit(workspace)) {
-          const curNode = workspace.getCursor()?.getCurNode();
-          if (curNode && curNode.getSourceBlock()) {
-            const sourceBlock = curNode.getSourceBlock();
-            return !!(
-              !Blockly.Gesture.inProgress() &&
-              sourceBlock &&
-              sourceBlock.isDeletable() &&
-              sourceBlock.isMovable()
-            );
+          switch (this.navigation.getState(workspace)) {
+            case Constants.STATE.WORKSPACE:
+              const curNode = workspace?.getCursor()?.getCurNode();
+              if (curNode && curNode.getSourceBlock()) {
+                const sourceBlock = curNode.getSourceBlock();
+                return !!(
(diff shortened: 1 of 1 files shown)


src/navigation_controller.ts
```diff
@@ -557,22 +557,40 @@ export class NavigationController {
...
+              if (curNode && curNode.getSourceBlock()) {
+                const sourceBlock = curNode.getSourceBlock();
+                return !!(
+                  !Blockly.Gesture.inProgress() &&
+                  sourceBlock &&
+                  sourceBlock.isDeletable() &&
+                  sourceBlock.isMovable()
+                );
+              }
+              return false;
```
cpcallen:
```suggestion
              const sourceBlock = curNode?.getSourceBlock();
              return !!(
                sourceBlock?.isDeletable() &&
                sourceBlock?.isMovable() &&
                !Blockly.Gesture.inProgress()
              );
```
`sourceBlock` might be null/undefined but I _think_ the final result will be the same in all cases.  I reordered the clauses in the final expression to match the order in the following case (but maybe it would be better to reorder that one instead).

cpcallen (approved):
Code seems fine, but I wonder about the expected behaviour when pasting with focus still on the flyout.  I'd sort of expect it to fail, rather than paste to the main workspace.  The later behaviour might be useful but is at least slightly surprising.

I don't think that needs to be addressed in this PR but I think the question deserves at least a note on #103 or one of the marker-related issues, so we don't forget to actually consider it properly.

src/navigation_controller.ts
rachel-fenichel: Applied manually.

Issue #101, opened 2024-11-08
Make tab key behaviour more intuitive


cpcallen:
This bug tracks several proposed changes to how the tab key and keyboard navigation interact.

* [x] Make keyboard navigation active any time the workspace (or any part of it) has focus.
  * At the moment it is necessary to tab to the workspace and then press enter to being interacting with it, and then press esc to deactivate keyboard navigation before it is possible to tab away from the workspace.
  * Instead, tabbing to the workspace should suffice to activate navigation, and tabbing away should deactivate it (and be possible at any time without first pressing esc).
* [x] Make it possible to tab to the toolbox.
* [ ] Implement [roving tab index](https://opensource.adobe.com/spectrum-web-components/tools/roving-tab-index/), for better compatibility with screen readers.

Instead, keyboard navigation should be active any time the workspace (or any part of it) has focus.

See also:

* #62
* #63
* #89
* #92
* #96

BenHenning:
So one question: how should navigation to toolbox work? The natural flow per the page layout, I think, would be that the toolbox is tabbed to before the workspace. However, this may not be an ideal flow since the toolbox is only sometimes needed by the user so to me it seems clearer to navigate first to the workspace. Or, are there other thoughts on how the toolbox should be tab-navigable?

rachel-fenichel:
I [summarized our discussion here](https://github.com/google/blockly-keyboard-experimentation/discussions/62#discussioncomment-11268064).

In testing, the kids did not at first know about tabbing, but picked it up relatively quickly and expected it to move logically through the page. I think it's fine for them to have to go to the toolbox and then to the block editor, and that skipping it would make it harder for them to realize that the toolbox is reachable through tabbing.

rachel-fenichel:
Closed because this large-scale feature is now tracked in smaller issues and feature requests.

PR #126, opened 2024-11-15
feat: Simplify workspace navigation flow

### LINKED ISSUE #89: Cache cursor location when exiting workspace and resume when reentering
When the user tabs out of the workspace (and turns off Blockly keyboard nav), stash their location and hide the cursor. When they reenter, pop the location and show the cursor.

This may be @BenHenning for the code that handles entering and exiting the workspace and @cpcallen for the code that keeps track of the cursor.

BenHenning:
Fixes part of #101
Fixes #89

This change removes the need for using enter and escape to enable keyboard navigation when the workspace is focused.

Instead, simply focusing the workspace is sufficient to allow keyboard navigation. Note that this PR goes back to the previous behavior of always enabling keyboard navigation on page load and instead limits the preconditions of navigation keys to check focus rather than feature enabled state. This allows for #89 to be fully mitigated. Since keyboard navigation is never disabled, the cursor position is never lost (and thus doesn't need to be cached). The removal of enter/escape fixes the first part of #101.

This PR also performs some other minor cleanup work in ``navigation_controller.ts``.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -47,7 +47,6 @@ export class NavigationController {
   announcer: Announcer = new Announcer();
   shortcutDialog: ShortcutDialog = new ShortcutDialog();
 
-  isAutoNavigationEnabled: boolean = false;
   hasNavigationFocus: boolean = false;
 
   /**
@@ -139,22 +138,6 @@ export class NavigationController {
   }
 
   /**
-   * Sets whether the navigation controller should automatically intercept tab,
-   * enter, and escape navigation events to ensure that the user needs to
-   * explicitly enter and exit the Blockly environment for keyboard shortcuts to
-   * be enabled.
-   *
-   * This mode has no effect unless setHasFocus is correctly used to communicate
-   * focus gained and lost contexts.
-   *
-   * @param autoNavEnabled whether the controller should automatically
-   *     intercept navigation events.
-   */
-  setHasAutoNavigationEnabled(autoNavEnabled: boolean) {
-    this.isAutoNavigationEnabled = autoNavEnabled;
-  }
-
-  /**
    * Sets whether the navigation controller has focus. This has no side effect
    * unless auto navigation is enabled via setHasAutoNavigationEnabled.
    *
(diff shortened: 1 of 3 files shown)


BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @gonfunko or reroute if needed.

FYI @cpcallen and @rachel-fenichel.

src/navigation_controller.ts
```diff
@@ -139,29 +138,43 @@ export class NavigationController {
...
+   * can support keyboard navigation.
    *
-   * @param autoNavEnabled whether the controller should automatically
-   *     intercept navigation events.
+   * @param workspace the workspace in which keyboard navigation may be allowed.
+   * @returns whether keyboard navigation is currently allowed.
    */
-  setHasAutoNavigationEnabled(autoNavEnabled: boolean) {
-    this.isAutoNavigationEnabled = autoNavEnabled;
+  private isKeyboardNavigationCurrentlyAllowed(workspace: WorkspaceSvg) {
```
cpcallen: TIL you not only have [Java](https://www.reddit.com/r/ProgrammerHumor/comments/xb18vc/java_naming_conventions/?rdt=49176) readability but you are a Java readability reviewer…

Would `canNavigate` or, if you think that too inspecific, `canCurrentlyNavigate` suffice?

src/navigation_controller.ts
```diff
@@ -139,29 +138,43 @@ export class NavigationController {
...
-   * @param isFocused whether the environment has browser focus.
+   * For the criteria on keyboard navigability, see
+   * isKeyboardNavigationCurrentlyAllowed.
+   *
+   * @param workspace the workspace in which keyboard editing may be allowed.
+   * @returns whether keyboard navigation and editing is currently allowed.
    */
-  setHasFocus(isFocused: boolean) {
-    this.hasNavigationFocus = isFocused;
+  private isWorkspaceCurrentlyKeyboardEditable(workspace: WorkspaceSvg) {
```
cpcallen: `canEdit` or `canCurrentlyEdit`?

cpcallen (approved):

src/navigation_controller.ts
BenHenning: I like the suggestion for 'can' since it's more deliberately tied to the input when read out. Will make this change in a follow-up.

I'm (unfortunately) using 'currently' to capture the entire context of whether the accessibility mode is currently enabled. One alternative is to use 'and' but I'm usually not a fan of that (since it feels like an obvious violation of single responsibility even though it's just a name--I try to find the common single behavior). I don't have a better idea than 'currently' at the moment. :) I lean toward ``canCurrentlyEdit``.

src/navigation_controller.ts
BenHenning: I went into more details on 'currently' and 'can' in my other comment, but I think I lean toward ``canCurrentlyNavigate`` unless you have any suggestions for clearer names?

BenHenning:
I'll address the suggested changes & submit this early next week (I want to make sure I actually retest it since there isn't any CI to verify that I don't break the plugin if I try making changes on GitHub directly (: ).

BenHenning (commented):
Self-reviewed changes, and also tested them locally (though I didn't test the shortcuts exhaustively). Things should be in good order, though.

BenHenning:
Thanks @cpcallen! Going to go ahead and merge this since it's a significant navigation change that would be good to get in front of users for the next testing round. Please note that I resolved your comment threads--if you have any follow-up suggestions please let me know and I'll be happy to address them in a follow-up PR.

PR #135, opened 2024-12-06
Make workspace focus outline more prominent

### LINKED ISSUE #81: Make visual indication that blockly workspace is selected more obvious

bmxedd:
Fixes #81 

This fixes the previous hack to get the focus outline to display more prominently on the workspace.

I've opted to use the *Highlight* colour for better cross browser support as opposed to *--webkit-focus-ring-color*. This shows up as a different colour on webkit browsers. 

Note: in my tests, Firefox does not show any focus on the workspace after the tabindex was explicitly removed from the container elements. 

Apologies for the other empty commits which were previously merged. It's a state I cannot figure out.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -110,60 +110,66 @@ export class LineCursor extends Marker {
   }
 
   /**
-   * Decides if the previous and next methods should traverse the given node.
-   * The previous and next method only traverse previous connections, next
-   * connections and blocks.
+   * Returns true iff the given node represents the "beginning of a
+   * new line of code" (and thus can be visited by pressing the
+   * up/down arrow keys).  Specifically, if the node is for:
+   *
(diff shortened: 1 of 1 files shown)


bmxedd:
Friendly ping!

cpcallen (changes requested):
Hi Ed.  Sorry for not looking at this earlier.

The changes to `index.html` look fine (though I am very much not a CSS expert), but I see a bunch of (seemingly unrelated) changes to the shortcuts dialog.  Did you intend for these to be included in this PR?  If so, please update the description; if not then please rebase to remove them.  (Let me know if you need a hand with this!)

PR #141, opened 2025-01-09
feat: Only init the navigation cursor when needed

### LINKED ISSUE #140: Unexpected initial state with flashing cursor after workspace focus PR
https://github.com/google/blockly-keyboard-experimentation/pull/126 has changed the initial state in [the demo](https://google.github.io/blockly-keyboard-experimentation/) to put a flashing cursor on the workspace. This seems wrong as the focus is not yet on the workspace.

It is hard to see though as it's behind the default demo blocks! [Try a change like this](https://github.com/google/blockly-keyboard-experimentation/compare/main...microbit-matt-hillsdon:blockly-keyboard-experimentation:offset-200) to move them out of the way (demo'd [on this deployment](https://offset-200.blockly-keyboard-experimentation.pages.dev/)).

Image

I think this happens because the plugin now does much more at construction time than previously. The call to `navigationController.enable(workspace)` in the plugin init does more than you might expect.

We've also found it triggers a "marker_move" event (because it adds the cursor?) via the workspace change listener. This caused confusion in our MakeCode integration because it happened earlier than MakeCode was expecting to get events. I'm hopeful that a fix for the cursor will also address that but can split out if needed.

BenHenning:
Fixes #140

This introduces improved initialization logic around the navigation cursor whereby it's only initialized upon focus (with previous position being retained when navigating back to the workspace after tabbing out of it) rather than upon enabling the controller (since the cursor doesn't actually need to be active when navigation isn't enabled).

Behaviorally, this change results in the blinking cursor mentioned in #140 not existing. It also has the nice benefit of highlighting the first top-level block upon tabbing to the workspace (which will affect readout with a screenreader enabled, but I think in a way that's consistent with other a11y environments--often the first interactive element is enabled by default).

Separately, this PR also fixes a small documentation inconsistency.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,10 +36,10 @@ export class KeyboardNavigation {
     workspace.getParentSvg().removeAttribute('tabindex');
 
     workspace.getSvgGroup().addEventListener('focus', () => {
-      navigationController.setHasFocus(true);
+      navigationController.setHasFocus(workspace, true);
     });
     workspace.getSvgGroup().addEventListener('blur', () => {
-      navigationController.setHasFocus(false);
+      navigationController.setHasFocus(workspace, false);
     });
   }
 
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -535,14 +535,15 @@ export class Navigation {
    * top block on a workspace or to the workspace.
    *
    * @param workspace The workspace to focus on.
+   * @param keepCursorPosition Whether to retain the cursor's previous position.
    */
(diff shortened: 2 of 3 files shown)


BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @cpcallen or @rachel-fenichel (though feel free to review if you want to @gonfunko).

rachel-fenichel (approved):

BenHenning:
Thanks @rachel-fenichel!

Going ahead and merging this so that enable downstream testing/verification of the original issue.

Issue #136, opened 2025-01-03
Navigation after field edit broken by recent change


microbit-matt-hillsdon:
1. Go to https://google.github.io/blockly-keyboard-experimentation/
2. Use keyboard navigation to edit the canvas width
3. Attempt to continue keyboard navigation and find you're stuck in the field

You can rescue it with shift-tab tab.

This regressed in #126. 

CC @BenHenning

microbit-matt-hillsdon:
We think the problem is that when editing fields, focus moves to the widget or dropdown divs and then never moves back to the workspace. Workspace focus is now a precondition for the key events. We need to return the focus to the workspace (or later, perhaps the block) after the field edit.

This issue makes it difficult to test/explore MakeCode's field editors with the latest version of the plugin.

BenHenning:
Thanks for filing this @microbit-matt-hillsdon.

The ultimate issue here is that we now (as of #126) require focus of the Blockly workspace in order for keyboard navigation to be enabled, and the field input editors live as singletons outside of the workspace which means focus to the workspace is lost when those editors gain focus. The latter is an existing issue in core Blockly that has been around I think since the start of Blockly.

The reason this wasn't an issue before is because technically the keyboard events are global and thus it was possible to operate keyboard navigation regardless of focus (original plugin behavior) or based on an internal logical flag (the behavior before #126).

@gonfunko is considering a broader core Blockly change that would bring singletons like the input field editors into the injection div which would, indirectly, fix this issue. The alternative is to try and restore focus after input editing. From investigation it seems that this isn't something we can do via the plugin and would also require a core Blockly change. We're probably going to implement something like this farther in the future, anyway, since we need to introduce the concept of dynamic focus in order for keyboard navigation to correctly trigger readout from an active screenreader. With @gonfunko's change, however, we can possibly wait on that exact part of dynamic focus until later when we bring it from the plugin (where it will be initially implemented) into core Blockly.

rachel-fenichel:
@BenHenning can you restore focus after input editing with a monkeypatch on core in order to unblock some testing? The long-term solution would need core to be aware of the navigation code and that's not happening in the next short while, so a monkeypatch is appropraite.

rachel-fenichel:
[...]

Issue #87, opened 2024-11-05
firefox can't focus code editor


kmcnaught:
The current code (13bc49401) had some last minute hacks to (a) remove an extraneous tab stop before the editor and (b) change the focus indicator for tabbing to the editor. I presume (a) is buggy (i.e. 3100903e6) because I can't tab to the code editor on Firefox. I'm not sure what the behaviour was before those 2 commits. 

Here's a [30 second loom video](https://www.loom.com/share/229d67cc07ad4aea8babea1d9a11c05e) comparing the two behaviours

Firefox 132.0 (64-bit) on Windows 11.

cpcallen:
Tentatively assigning this to Ben as it relates to the work he is doing fixing tabbing more generally.

microbit-matt-hillsdon:
This also applies to Safari. I think the difference is that Chrome will focus elements with a focus listener that it otherwise would not, but for other browsers you need tabindex. I experimented with this [in this sandbox](https://codesandbox.io/p/sandbox/nvdp22).

The extra tab stop appears to be back for the same reason since #156.

microbit-matt-hillsdon:
Recent focus work has helped here but there are still serious issues and some Firefox/Safari review is needed.

The most glaring issues 

### Move mode appears to be broken with a logged message "Trying to focus a node that can't be focused."

This affects insert and 'M'.

### Clicking on a block doesn't correctly enable keyboard navigation (e.g. can't press 't' to open toolbox). Repro:

1. Open local build of demo in Safari (I tested against e34a9690eddf4512f682aa80c697211123326a7c of blockly)
2. Click on "create canvas" block
3. Try to navigate or open toolbox
4. Click on a different block, note same issue
5. Click back to "create canvas", note keyboard nav works now

There are also default styles for focus that I'll look at on #511,.

I think these suggest fundamental issues and someone looking at focus changes should take a look soon to unblock further Firefox/Safari testing.

BenHenning:
Thanks @microbit-matt-hillsdon that's very helpful. I can confirm that in Firefox tabbing to the workspace works fine, but the exact same problems that you see in Safari also seem to be affecting Firefox. These aren't completely new failures as we had versions of them with Chrome in the past, so I'll dig and figure out what's happening here.

BenHenning:
[...]

PR #156, opened 2025-01-16
fix: add a focus listener on the parent svg

rachel-fenichel:
This works around the issue reported in #136 to unblock testing, but does not fix the root cause.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -124,6 +124,12 @@ export class NavigationController {
    */
   addWorkspace(workspace: WorkspaceSvg) {
     this.navigation.addWorkspace(workspace);
+
+    // Temporary workaround for #136.
+    // TODO(#136): fix in core.
+    workspace.getParentSvg().addEventListener('focus', () => {
+      this.setHasFocus(workspace, true);
+    });
   }
 
   /**
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -124,12 +124,6 @@ export class NavigationController {
    */
   addWorkspace(workspace: WorkspaceSvg) {
     this.navigation.addWorkspace(workspace);
-
-    // Temporary workaround for #136.
-    // TODO(#136): fix in core.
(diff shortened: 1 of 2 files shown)


src/navigation_controller.ts
```diff
@@ -124,6 +124,12 @@ export class NavigationController {
    */
   addWorkspace(workspace: WorkspaceSvg) {
     this.navigation.addWorkspace(workspace);
+
+    // Temporary workaround for #136.
+    // TODO(#136): fix in core.
+    workspace.getParentSvg().addEventListener('focus', () => {
```
BenHenning: Might this be better placed near the existing location where we register focus:

https://github.com/google/blockly-keyboard-experimentation/blob/0402a4d0dae7eeaa24cdd3cfca387af94f87c38d/src/index.ts#L38-L43

That way we can keep all focus/blur management in one location.

In terms of focus/blur symmetry, I suppose we should probably just switch that logic over to using parent SVG instead of the SVG group. I thought that I had tried this before and ran into issues, but I may be misremembering.

BenHenning (commented):
Thanks @rachel-fenichel! Just had one comment--PTAL.

rachel-fenichel:
Updated. Not sure this is the right thing to do long-term, but I want to unblock testing. This also solves a parallel issue with dismissing the context menu, which also lives in the widget div.

src/navigation_controller.ts
BenHenning: Addressed in latest.

BenHenning (approved):
Thanks @rachel-fenichel!

microbit-matt-hillsdon:
One issue with this approach is that Chrome seems to take the focus listener to mean the parent SVG is now a tab position so the invisible tab stop between the select and the focus-outlined workspace is back. Other browsers have more significant issues because of this difference—I've also commented on #87.

Joe1the2Creator3 (commented):
Apply approved

PR #157, opened 2025-01-17
fix: use valid colour in second colour picker

### LINKED ISSUE #68: Unable to pick colour in second colour picker in initial test scenario #2
## Actual Behavior

I could open the colour-picker pop-up, but was unable to select any colour—cursor keys did nothing.

1. Run `npm start` to start keyboard nav demo page.
2. Press `tab` three times to navigate to Blockly worksapce.
3. Press `enter` to begin navigating within Blockly.
4. Press `↓` 8 times (or as needed) to navigate to the "draw cirlce" block.
5. Press `→` 2 times to navigate to the colour block.
6. Press `enter` to open colour picker.
7. Press arrow keys; note that colour picker does not respond.

Compare to behaviour of colour block in "set background colour" block, which works as expected.

rachel-fenichel:
Patches over #68 by using a valid colour as the starting value.

### DIFF
--- a/test/loadTestBlocks.js
+++ b/test/loadTestBlocks.js
@@ -236,7 +236,7 @@ const simpleCircle = {
                     'type': 'colour_picker',
                     'id': 'gq(POne}j:hVw%C3t{vx',
                     'fields': {
-                      'COLOUR': '#f59b42',
+                      'COLOUR': '#ffff00',
                     },
                   },
                 },


cpcallen (approved):

cpcallen:
I'm going to mark #68 as fixed but open a bug in samples to replace it.

Issue #130, opened 2024-11-15
Keyboard navigable context menus


cpcallen:
* [x] When the cursor is visiting a block (or equivalently, when a block has focus) or visiting a connection (which, per #129, will only be possible when that connection does not have a connected block), pressing enter should open that block's context menu and give it focus.
* [x] The context menu should be able to display keyboard shortcuts for each item.
* [x] When opening the context menu using the keyboard, the first item in the menu should be selected by default.
* [x] The user should be able to navigate the context menu using the up and down arrow keys.
* [x] The user should be able  and select the current item by using enter, or any item with a shortcut by pressing that shortcut; in either case the menu should close.
* [x] The user should be able to close the menu without selecting an item by pressing escape.
* [x] When the menu closes, focus should return to its previous location (i.e., the cursor's location)

Combined with proposals to add new context menu options (#TBD...), this facilitates discovering all relevant block-related keyboard shortcuts, including the possibility discovering one can navigate to a block's fields and inputs.

cpcallen:
Updated proposal (and issue title) to reflect the need to be able to display context menus for connections, to support user discovery of the ability to place blocks using the revised marker workflow (see #102).

cpcallen:
Questions for discussion:

When I wrote up this issue, I wrote (edited, emphasis added):

> * [ ] When the cursor is visiting a block (or equivalently, when a block has focus) **or visiting a[n unconnected] connection** … pressing enter should open that block's context menu and give it focus.

My thinking was that we need a way to access shortcuts like "insert a block here" via a menu (i.e., discoverably) even when there is not currently a block there.  I think that is still desirable, but additionally now:

* I think the same should apply to the workspace, for placing new top-level blocks.
* I think it seems kind of odd to have special context menus, that can only be accessed by the keyboard and only for unconnected connections.  Should it be possible to open such context menus via the mouse, and/or on all connections, not just unconnected ones?
  * We could avoid this issue by preventing navigation to any connections, even disconnected ones—but that would prevent users from placing blocks on open connections using the "marker workflow" (i.e., navigate to desired location then choose block, rather than choosing block first then moving it to desired location).

kmcnaught:
Hi @cpallen, to clarify, we are expecting to lose input/output connections (unless there's nothing on them) but we want previous/next connections to always be available for inserting a block.

[...]

PR #158, opened 2025-01-17
feat: Press enter on block to open action menu

cpcallen:
Part of #130.

Introduce the "action menu", which uses the `Blockly.ContextMenu` machinery to display a custom menu of shortcut actions on a block when you press enter.

For now the action menu contents are identical to the normal context menu contents, but it is created in a way that will allow us to control the contents of the action menu independently of the block's custom context menu (or request not to have a context menu at all).

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1231,13 +1231,15 @@ export class Navigation {
    */
   handleEnterForWS(workspace: Blockly.WorkspaceSvg) {
     const cursor = workspace.getCursor();
-    if (!cursor) {
-      return;
-    }
+    if (!cursor) return;
     const curNode = cursor.getCurNode();
     const nodeType = curNode.getType();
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):

src/navigation.ts
```diff
@@ -1248,13 +1248,63 @@ export class Navigation {
...
+  /**
+   * Show the action menu for a given node.
+   *
+   * The action menu will contain entries for relevant actions for the
+   * node's location.  If the location is a block, this will include
+   * the contents of the block's context menu (if any).
+   */
+  openActionMenu(node: Blockly.ASTNode) {
+    const fakeEvent = fakeEventForNode(node);
+    (node.getLocation() as Blockly.BlockSvg).showContextMenu(fakeEvent);
```
cpcallen, line 1265: What on earth was I doing?  This does not belong here!

PR #159, opened 2025-01-18
feat: keyboard navigation delete option in context menu

rachel-fenichel:
This bodges together the keyboard shortcut item and the context menu item to register a "Keyboard navigation: delete" item into the context menu and show it at the appropriate time.

This is an example of how to transform a keyboard shortcut definition into a context menu item.

The added context menu item differs from the original context menu item because its precondition function updates the location of the cursor after deletion, rather than leaving it in an invalid state.

Parts of the code:
- Make the keyboard shortcut precondition and callback functions into methods on `NavigationController`.
- Cache the original context menu item to use its precondition function.
- Register a context menu item that mashes these functions together.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -229,7 +229,7 @@ export class NavigationController {
    * Precondition function for deleting a block from keyboard
    * navigation. This precondition is shared between keyboard shortcuts
    * and context menu items.
-   * 
+   *
    * FIXME: This should be better encapsulated.
    *
    * @param workspace The `WorkspaceSvg` where the shortcut was
@@ -251,7 +251,7 @@ export class NavigationController {
    * Callback function for deleting a block from keyboard
    * navigation. This callback is shared between keyboard shortcuts
    * and context menu items.
-   * 
+   *
    * FIXME: This should be better encapsulated.
    *
    * @param workspace The `WorkspaceSvg` where the shortcut was
@@ -260,7 +260,7 @@ export class NavigationController {
    *     if called from a context menu.
    * @returns True if this function successfully handled deletion.
    */
-  protected deleteCallbackFn(workspace: WorkspaceSvg, e: Event|null) {
+  protected deleteCallbackFn(workspace: WorkspaceSvg, e: Event | null) {
     const cursor = workspace.getCursor();
     if (!cursor) {
       return false;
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@cpcallen fyi. This isn't ready to merge, but it does work.

src/navigation_controller.ts
```diff
@@ -224,6 +225,65 @@ export class NavigationController {
...
+   */
+  protected deletePreconditionFn(workspace: WorkspaceSvg) {
+    if (this.canCurrentlyEdit(workspace)) {
+      const curNode = workspace.getCursor()?.getCurNode();
+      if (curNode && curNode.getSourceBlock()) {
+        const sourceBlock = curNode.getSourceBlock();
+        return !!(sourceBlock && sourceBlock.isDeletable());
+      }
+    }
+    return false;
```
cpcallen:
```suggestion
    if (!this.canCurrentlyEdit(workspace)) return false;
    const sourceBlock = workspace.getCursor()?.getCurNode().getSourceBlock();
    return !!(sourceBlock?.isDeletable());
```

src/navigation_controller.ts
```diff
@@ -224,6 +225,65 @@ export class NavigationController {
...
+   *     invoked.
+   * @param e The originating event for a keyboard shortcut, or null
+   *     if called from a context menu.
+   * @returns True if this function successfully handled deletion.
+   */
+  protected deleteCallbackFn(workspace: WorkspaceSvg, e: Event | null) {
+    const cursor = workspace.getCursor();
+    if (!cursor) {
+      return false;
+    }
```
cpcallen:
```suggestion
    if (!cursor) return false;
```

src/navigation_controller.ts
```diff
@@ -224,6 +225,65 @@ export class NavigationController {
...
+    if (e) {
+      // Stop the browser from going back to the previous page.
+      // Do this first to prevent an error in the delete code from resulting
+      // in data loss.
+      e.preventDefault();
+    }
+    // Don't delete while dragging.  Jeez.
+    if (Blockly.Gesture.inProgress()) {
+      return false;
+    }
```
cpcallen:
```suggestion
    if (Blockly.Gesture.inProgress()) false;
```

cpcallen (approved):

rachel-fenichel:
I accepted the updates and tested it locally before merging.

PR #161, opened 2025-01-20
feat: add a category toolbox to the test page

### LINKED ISSUE #160: Add toolbox categories and more blocks to test page
For improved testing.

rachel-fenichel:
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/160
image

Includes the p5 blocks as a category--they are now exported from `src/blocks/toolbox.js` as `p5CategoryContents`.

This means that it is possible to create non-runnable code, but provides an expanded set of test cases for development.

### DIFF
--- a/test/toolboxCategories.js
+++ b/test/toolboxCategories.js
@@ -1,0 +1,798 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {p5CategoryContents} from '../src/blocks/toolbox.js';
+
+export default {
+  kind: 'categoryToolbox',
+  contents: [
+    {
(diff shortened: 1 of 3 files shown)


test/toolboxCategories.js
cpcallen, line 1: Wouldn't it be tidier to import the standard toolbox from `@blockly/dev-tools` (needed only as a dev dependency, since this is in `test/`) and `push` the extra category on?

rachel-fenichel: I specifically avoided pulling in dev-tools because it's easy to get into a bad dependency state when updating this to use a beta without updating dev-tools.

cpcallen: Fair point.

cpcallen (approved):

PR #167, opened 2025-01-21
fix: paste at active cursor location instead of at marker location

### LINKED ISSUE #165: Paste at active cursor location
Currently paste puts the block at the marker location (if any) and onto the workspace otherwise. It should put the block at the cursor location instead of the marker location, so that the user does not need to explicitly create a marker.

rachel-fenichel:
Fixes #165 

If the cursor location is not valid, pastes on the workspace (same behaviour as before).

The first commit cleans up a helper function to refer to "nodes" instead of "marker" and "cursor". The second commit actually changes the behaviour.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -537,7 +537,10 @@ export class Navigation {
    * @param workspace The workspace to focus on.
    * @param keepCursorPosition Whether to retain the cursor's previous position.
    */
-  focusWorkspace(workspace: Blockly.WorkspaceSvg, keepCursorPosition: boolean = false) {
+  focusWorkspace(
+    workspace: Blockly.WorkspaceSvg,
+    keepCursorPosition: boolean = false,
+  ) {
     workspace.hideChaff();
     const reset = !!workspace.getToolbox();
 
@@ -554,7 +557,10 @@ export class Navigation {
    * @param workspace The main Blockly workspace.
    * @param keepPosition Whether to retain the cursor's previous position.
    */
-  setCursorOnWorkspaceFocus(workspace: Blockly.WorkspaceSvg, keepPosition: boolean) {
+  setCursorOnWorkspaceFocus(
+    workspace: Blockly.WorkspaceSvg,
+    keepPosition: boolean,
+  ) {
     const topBlocks = workspace.getTopBlocks(true);
     const cursor = workspace.getCursor();
     if (!cursor) {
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -687,57 +687,54 @@ export class Navigation {
...
     workspace: Blockly.WorkspaceSvg,
-    markerNode: Blockly.ASTNode,
-    cursorNode: Blockly.ASTNode,
+    firstNode: Blockly.ASTNode,
+    secondNode: Blockly.ASTNode,
   ): boolean {
-    if (!this.logConnectionWarning(markerNode, cursorNode)) {
+    if (!this.logConnectionWarning(firstNode, secondNode)) {
       return false;
     }
```
cpcallen:
```suggestion
    if (!this.logConnectionWarning(firstNode, secondNode)) return false;
```

src/navigation.ts
```diff
@@ -1343,9 +1340,9 @@ export class Navigation {
     block: Blockly.BlockSvg,
   ): boolean {
     let isHandled = false;
-    const markedNode = workspace.getMarker(this.MARKER_NAME)?.getCurNode();
+    const markedNode = workspace.getCursor()?.getCurNode();
```
cpcallen: Maybe `cursorNode` or `focusedNode` instead of `markedNode`?

src/navigation.ts
```diff
@@ -687,57 +687,54 @@ export class Navigation {
...
   /**
-   * Tries to connect the given marker and cursor node.
+   * Tries to intelligently connect the blocks or connections
+   * represented by the given nodes, based on node types and locations.
    *
    * @param workspace The main workspace.
-   * @param markerNode The node to try to connect to.
-   * @param cursorNode The node to connect to the markerNode.
+   * @param firstNode The first node to connect.
+   * @param secondNode The second node to connect.
```
cpcallen: If I understand correctly `secondNode` is the thing being attached, and `firstNode` is the location to attach it to.  Perhaps `newLocation` and `newItem`, or `parent` and `child`, or similar (instead of `first` and `second`)?

cpcallen (commented):
I think the actual implementation of `tryToConnectNodes` should be revised so that it has the same behaviour regardless of whether `firstNode` is a block or that block's parent connection—but I think that can be done in a later PR, especially as this is a good, straightforward refactor.

src/navigation.ts
rachel-fenichel: Renamed to `targetNode`.

src/navigation.ts
rachel-fenichel: Will that still apply if this function is made to actually be properly symmetrical, though?

cpcallen: Hmm.  I agree that my proposed change would be silly if the function was commutative—but I am not sure I understand why we would want that.

~~When moving blocks (or adding new blocks from the toolbox), there is always one group of blocks (the one that is moving / new)that's being connected by its root (prev/output) connection, and one group of blocks that is being connected by a child (next/input) connection.~~

Oh, scratch that (no pun intended): I've been spending too much time looking at event-driven code, because I'd forgotten you can connect blocks to the top of a stack as well as the bottom, and to the output of a top-level value block.  So the parent/child naming scheme might not be sensible—but is a moving blocks vs. stationary blocks distinction also not still useful?  Hmm.  🤔

Let's chat about this tomorrow.  I want to understand why you would like to make this function symmetric.

rachel-fenichel: Updated to change names. PTAL.

cpcallen (approved):

PR #166, opened 2025-01-21
feat: add block copy as context menu item

rachel-fenichel:
Part of https://github.com/google/blockly-keyboard-experimentation/issues/132

As discussed with @cpcallen on 20 Jan 2025, this adds a context menu item for an existing keyboard shortcut (copy). I intend to apply this transformation to several keyboard shortcuts by making changes in `navigation_controller.ts`, then find a general pattern for creating `Action`s that can be registered as both keyboard shortcuts and context menu items. In that second stage I will likely move these definitions out of the `navigation_controller.ts` and into separate per-action files, to improve code cleanliness/readability.

This change is related to https://github.com/google/blockly-keyboard-experimentation/pull/159

PR #171, opened 2025-01-21
feat: contextmenu action to insert above a block

rachel-fenichel:
Part of #130 

Related to #166

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -951,6 +951,40 @@ export class NavigationController {
   }
 
   /**
+   * Register the action for inserting above a block.
+   */
+  protected registerInsertAction() {
+    const insertAboveAction: ContextMenuRegistry.RegistryItem = {
+      displayText: (scope) => 'Keyboard Navigation: insert above',
+      preconditionFn: (scope) => {
+        const ws = scope.block?.workspace;
+        if (!ws) return 'hidden';
+
+        if (!scope.block?.previousConnection) return 'hidden';
+
+        return this.canCurrentlyEdit(ws) ? 'enabled' : 'hidden';
+      },
+      callback: (scope) => {
+        const ws = scope.block?.workspace;
+        if (!ws) return false;
+
+        if (this.navigation.getState(ws) == Constants.STATE.WORKSPACE) {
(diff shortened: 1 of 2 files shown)


cpcallen:
I think it would be better to avoid moving the cursor, and instead [make the block insertion happen at the parent connection](https://github.com/google/blockly-keyboard-experimentation/pull/167#pullrequestreview-2564321707):

* If the user cancels the insert it will leave the cursor in the wrong place.
* This code will break if the connection isn't a valid cursor location.

rachel-fenichel:
Updated and rebased.

src/navigation_controller.ts
```diff
@@ -950,6 +950,37 @@ export class NavigationController {
     ContextMenuRegistry.registry.register(copyAction);
   }
 
+  /**
+   * Register the action for inserting above a block.
+   */
+  protected registerInsertAction() {
+    const insertAboveAction: ContextMenuRegistry.RegistryItem = {
+      displayText: (scope) => 'Keyboard Navigation: insert above',
```
cpcallen: We can probably discuss preferred wording (maybe it should change depending on the cursor location), but I'd suggest simply "Insert block" for now.

src/navigation_controller.ts
```diff
@@ -950,6 +950,37 @@ export class NavigationController {
...
+      preconditionFn: (scope) => {
+        const ws = scope.block?.workspace;
+        if (!ws) return 'hidden';
+
+        if (!scope.block?.previousConnection) return 'hidden';
```
cpcallen, line 963: Is it disabled on the top block of the stack because tryToConnectNodes can't handle that case yet?

cpcallen (approved):

src/navigation_controller.ts
rachel-fenichel: I'm making it "insert block above" because I still think that insert meaning before is non-obvious to non-programmers.

src/navigation_controller.ts
rachel-fenichel: This is the case where there's no previous connection, so there's no way to insert above. This comes up for the event blocks.

PR #176, opened 2025-01-24
fix: copy/delete shortcuts must bind extracted functions

microbit-matt-hillsdon:
Otherwise delete and copy shortcuts error when accessing fields on `this`.


microbit-matt-hillsdon:
@rachel-fenichel please can you help with the CLA check? We signed the corporate one yesterday. As I understand it, it's using a "microbit-google-contributors" Google group that contains my email which I also just made public on my GitHub account prior to opening this PR.

Issue #97, opened 2024-11-08
Clicking on a block selects it but does not move cursor


cpcallen:
When you use the mouse to click on a block it is selected, and any other selected block (including the one denoting the cursor location) is deselected, but the cursor is not moved to the selected block, so using keyboard navigation controls causes the cursor (to appear to) jump unexpectedly to nearby to its previous location.

PR #178, opened 2025-01-24
fix: Move cursor when user selects a block with mouse

cpcallen:
Hack to temporarily address #97.

Overrides `Marker.prototype.getCurNode` in `LineCursor` to add a hack that attempts to detect if the user has moved focus by selecting a block and, if so, update the cursor location (and any highlighting) to match.

This works reasonably well but has some glitches, most notably that if the cursor is not on a block (e.g. it is on a connection or the workspace) then it will remain visible in its previous location until a cursor key is pressed.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -403,7 +403,7 @@ export class LineCursor extends Marker {
   }
 
   /**
-   * Get the current location of the marker.
+   * Get the current location of the cursor.
    *
    * Overrides superclass implementation to add a hack that attempts
    * to detect if the user has moved focus by selecting a block and,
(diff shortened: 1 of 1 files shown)


src/line_cursor.ts
```diff
@@ -402,6 +402,49 @@ export class LineCursor extends Marker {
     return this.getRightMostChild(newNode);
   }
 
+  /**
+   * Get the current location of the marker.
```
rachel-fenichel: Current location of the _cursor_, not the marker, here and on line 421.

rachel-fenichel (approved):

src/line_cursor.ts
cpcallen: Hah: description copied from superclass, but fair enough!

Issue #181, opened 2025-01-28
Remove any Marker usage other than for insert


cpcallen:
The insert action should be the only way to place the marker, and when using that action the marker should be transient.

rachel-fenichel:
#187 addresses making the marker ephemeral.

If the user has the cursor on a block and hits T to open the toolbox, I think the marker should be placed at an appropriate connection and the block selected from the toolbox should be connected if possible. 

For instance, since the default for a stack block is to insert before, the marker should be placed before the block to show where the insertion will occur.

What do you think?

rachel-fenichel:
Fixed: there is no longer a shortcut to add the marker at an arbitrary location. The marker is added and then removed when the toolbox/flyout is opened and then closed.

microbit-matt-hillsdon:
I can reproduce a persistent marker with the cursor in a next/previous position, pressing 't' or 'enter' and then escape. I'm not sure it gets used for anything now, but it is confusing.

<img width="345" alt="Image" src="https://github.com/user-attachments/assets/f409b233-7cb3-4bd0-8af1-42740e0bf2b4" />

(please let me know if you prefer notes like this as new issues)

rachel-fenichel:
Normally I'd like those as new issues, but that one is tracked by https://github.com/google/blockly-keyboard-experimentation/issues/168

The issue there is that the _cursor_ is in the wrong place or rendering poorly, which I'm still trying to track down, in order to get the marker into the right place and get good reset behaviour.

Issue #168, opened 2025-01-21
Hitting Escape in toolbox should return cursor to marker location and delete marker


rachel-fenichel:
Currently you can get to the toolbox by hitting "I" for insertion. This adds a (temporary) marker to the workspace and moves the cursor to the toolbox.

If the user selects a block in the flyout and hits enter, that block will be placed at the marker location and the marker deleted.

### Current behaviour
If the user hits Escape to leave the flyout, the marker is left in place and the cursor is moved to the top level of the workspace. 

### Expected behaviour
Hitting Escape in the flyout or toolbox puts the cursor back at the marker location on the workspace, and deletes the temporary marker.

PR #187, opened 2025-02-05
feat: make the marker ephemeral

rachel-fenichel:
Part of https://github.com/google/blockly-keyboard-experimentation/issues/168
Part of https://github.com/google/blockly-keyboard-experimentation/issues/181

This change gets rid of the `Marker` object that we were using to mark where the cursor was before a jump to the toolbox or flyout.

That `Marker` was a variant on a `Cursor`, and there was no way to hide it once it had been shown.

In November and January we decided that the marker workflow was confusing, and that the marker should only be visible while actively inserting a block from the toolbox/flyout. In this context, the marker shows where the selected block will be inserted. The marker is then expected to disappear after the block is inserted.

Before this change, the marker stuck around after the new block was inserted. Calling `hide` on the marker didn't make it go away, because core Blockly redrew it during the next render pass.

### Changes in this PR
- Stop using the marker from core and its associated renderer.
- Start calling the marked node "passively focused". The "active focus" is the cursor. The "passive focus" shows up when we move the cursor to a new context but want to show where it will return to.
- Store the location of the passive focus in `navigation.ts`.
- Add a new class, `PassiveFocus`, that renders passive focus at a specific location. This new class only keeps track of it location to let it hide later.
- Implement passive focus show/hide functions for blocks and next connections.
- Correctly show and hide the passive focus indicator when opening the toolbox/flyout.
- Correctly hide the cursor when opening the flyout or toolbox for insertion.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -88,7 +88,15 @@ export const SHORTCUT_CATEGORIES = {
     'toolbox',
     'disconnect',
   ],
-  'Editing': ['cut', 'copy', 'paste', 'undo', 'redo', 'enter_or_mark', 'insert'],
+  'Editing': [
+    'cut',
+    'copy',
+    'paste',
+    'undo',
+    'redo',
+    'enter_or_mark',
+    'insert',
+  ],
   'Code navigation': [
     'previous',
     'next',
@@ -104,6 +112,6 @@ export const SHORTCUT_CATEGORIES = {
     'workspace_up',
     'workspace_right',
     'clean_up_workspace',
-    'intercept_tab_navigation'
+    'intercept_tab_navigation',
   ],
 };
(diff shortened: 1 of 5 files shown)


gonfunko (approved):

rachel-fenichel:
Implications for testing changes to the marker and its rendering:
- You can modify the appearance of the next connection indicator by applying CSS--the class is `passiveNextIndicator`.
- You can change the shape of the next connection indicator by modifying the `createNextIndicator` function.
- You can add indicators for other types of nodes by adding cases to `passiveFocus.show` and `passiveFocus.hide` that are parallel to the cases for the next indicator.
  - This can include creating new DOM elements for other cases. See `createNextIndicator` for an example.

PR #190, opened 2025-02-06
feat: Allow disposing of a KeyboardNavigation instance.

### LINKED ISSUE #186: Add a dispose function to the KeyboardNavigation class
The existing Blockly NavigationController has a dispose() function that we use regularly to keep from getting errors on multiple initialization calls. It would be great if there were an equivalent in the new experiment, whether it is written in the NavigationController or the KeyboardNavigation class itself.

gonfunko:
This PR adds a `dispose()` method to `KeyboardNavigation`, which can be used to disable keyboard navigation for the workspace. A  new `KeyboardNavigation` instance can subsequently be instantiated in order to reenable it. Fixes #186

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,24 @@ export class KeyboardNavigation {
   /** The workspace. */
   protected workspace: Blockly.WorkspaceSvg;
 
+  /** Event handler run when the workspace gains focus. */
+  private focusListener: () => void;
+
+  /** Event handler run when the workspace loses focus. */
+  private blurListener: () => void;
+
+  /** Keyboard navigation controller instance for the workspace. */
+  private navigationController: NavigationController;
+
+  /**
+   * These fields are used to preserve the workspace's initial state to restore
+   * it when/if keyboard navigation is disabled.
+   */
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):

src/index.ts
```diff
@@ -22,33 +40,80 @@ export class KeyboardNavigation {
...
+    this.workspace.getSvgGroup().removeEventListener('blur', this.blurListener);
+    this.workspace
+      .getSvgGroup()
+      .removeEventListener('focus', this.focusListener);
+
+    if (this.workspaceParentTabIndex) {
+      this.workspace
+        .getParentSvg()
+        .setAttribute('tabindex', this.workspaceParentTabIndex);
+    }
```
cpcallen, line 101: This looks suspect to me: after all, the most common tabindex value is `0`, _absence_ of a tabindex attribute is also an important state that might need to be restored?

(Oh: I guess `tabindex="0"` probably gets you the value `"0"` which is truthy—but the `null` case should probably still be dealt with correctly.)

Here and similarly for several similar situations below.

Issue #184, opened 2025-01-29
Context menu at connection location


rachel-fenichel:
Regardless of our decisions about how to unify (or not unify) navigable locations on single-field blocks, we need to be able to display a context menu when the cursor is at a connection. This comes up when navigating to the last connection on a stack to insert a block.

This issue is for the technical work involved:
- Trigger when appropriate
- Get a list of context menu items to show

kmcnaught:
Also: 
- offset context menu so you can still see the block in question (Matt's demo has this already, but the current implementation here doesn't)

rachel-fenichel:
Status update: after #209 the action menu can be triggered on connections. 

Remaining work: position appropriately relative to the connection it is attached to.

PR #191, opened 2025-02-07
feat: show some block for context when opening action menu

microbit-matt-hillsdon:
Previously it was shown over the top left, which obscured useful context.  I considered positioning it below the block but it can be a long way down in some cases. That can feel unexpected and makes it more likely that the menu is repositioned to keep it on screen.

Mentioned on https://github.com/google/blockly-keyboard-experimentation/issues/184#issuecomment-2623882452

This is a less arbitrary version of the offset I had in "Demo A" (that just used 50px which looked fine for zelos styling but not great in geras).

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1406,14 +1406,27 @@ function fakeEventForNode(node: Blockly.ASTNode): PointerEvent {
   // Get the location of the top-left corner of the block in
   // screen coordinates.
   const block = node.getLocation() as Blockly.BlockSvg;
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
Deployment of this PR's branch: https://position-menu.blockly-keyboard-experimentation.pages.dev/

rachel-fenichel (approved):
I like the positioning you chose, as it addressed one of my concerns about being too far from where the user is looking.

During testing we'll need to look at students using onscreen magnifiers to make sure that enough block and context menu is in view even at a high level of magnification.

Issue #155, opened 2025-01-16
Design unified object for keyboard shortcuts and context menu actions


rachel-fenichel:
As part of #130, we need to register all keyboard shortcuts (that should be accessible from context menus) as context menu items, with appropriate callbacks and preconditions.

Since the shortcuts are all defined in as an array of `ShortcutRegistry.KeyboardShortcut` items in `navigation_controller.ts`, it would be best to use that as the source of truth and write a helper to do the mapping, which is called once.

Later when we add a different scope to the context menu system we can update this helper to add the correct scope.

rachel-fenichel:
- [ ] Where a shortcut is intended to be on the meta key, this unified object should take in that information and map to the appropriate key for the platform, rather than requiring the creator of the shortcut to enumerate all options.

- [ ] The shortcut key information should be shown in the context menu to assist in discovery of new shortcuts.

rachel-fenichel:
More requirements:
- Add a new scope type, `Connection`, to contextmenu scopes.

rachel-fenichel:
- [ ] Both the precondition and the callback functions should receive information about whether the context menu was opened through the mouse or the keyboard.

cpcallen:
To be integrated into google/blockly#8788 and closed.

PR #192, opened 2025-02-07
feat: enter now edits full field blocks

microbit-matt-hillsdon:
- Skip the field positions for those blocks
- Move the action menu to Ctrl/Cmd+Enter

This is a tidied up subset of what we've discussed as Demo A.

The alert is not intended to be the final UX but to indicate that there's a point where we need to offer the user a hint. We can drop it from this PR and add it after more UX discussion if preferred.

I've assumed that we're OK with calling isFullBlockField for now (protected in core).

Deployment of this PR: https://enter-to-edit.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -552,6 +552,34 @@ export class NavigationController {
       keyCodes: [KeyCodes.ENTER],
     },
 
+    /**
+     * Cmd/Ctrl/Alt+Enter key:
+     *
+     * Shows the action menu.
+     */
+    menu: {
+      name: Constants.SHORTCUT_NAMES.MENU,
+      preconditionFn: (workspace) => this.canCurrentlyNavigate(workspace),
+      callback: (workspace) => {
+        switch (this.navigation.getState(workspace)) {
+          case Constants.STATE.WORKSPACE: {
+            const node = workspace.getCursor()?.getCurNode();
+            if (node?.getType() === Blockly.ASTNode.types.BLOCK)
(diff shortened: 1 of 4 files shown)


src/navigation_controller.ts
```diff
@@ -552,6 +552,34 @@ export class NavigationController {
...
+              this.navigation.openActionMenu(node);
+
+            return true;
+          }
+          default:
+            return false;
+        }
+      },
+      keyCodes: [
+        createSerializedKey(KeyCodes.ENTER, [KeyCodes.CTRL]),
```
microbit-matt-hillsdon, line 577: I've followed the pattern for cut/copy/paste here, but alternatively we could check the platform at this point. Or the shortcut mechanism could have a way of defining this kind of platform specific shortcut. Would any platform use Alt in this way?

rachel-fenichel: Noted in https://github.com/google/blockly-keyboard-experimentation/issues/155#issuecomment-2643431314 to centralize responsibility for choosing the appropriate meta key.

In my (windows + linux) experience, Alt operates on windows.

rachel-fenichel (approved):

src/navigation_controller.ts
microbit-matt-hillsdon: I'm not convinced Alt+C and similar on Windows are real. Raised https://github.com/google/blockly-keyboard-experimentation/issues/258

PR #194, opened 2025-02-07
fix: Update cursor location in response to `SELECT` events

### LINKED ISSUE #97: Clicking on a block selects it but does not move cursor

cpcallen:
This adds an event listener for `SELECT` events to `LineCursor`.  This solves an issue where, when the cursor was not on a block, clicking on a block to select it would not immediately cause the cursor to disappear from its previous location, though it would eventually update when a cursor key was pressed.

Fixes #97

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -27,6 +27,7 @@ export class LineCursor extends Marker {
    */
   constructor(public readonly workspace: Blockly.WorkspaceSvg) {
     super();
+    workspace.addChangeListener(this.selectListener.bind(this));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/line_cursor.ts
```diff
@@ -511,6 +548,13 @@ export class LineCursor extends Marker {
 
     drawer.draw(oldNode, newNode);
   }
+
+  private selectListener(event: Blockly.Events.Abstract) {
```
rachel-fenichel, line 556: Add a comment: "Event listener that syncs the cursor location to the selected block on SELECTED events."

src/line_cursor.ts
```diff
@@ -410,10 +448,14 @@ export class LineCursor extends Marker {
...
+   * Doing this only when getCurNode would naturally be called works
+   * reasonably well but has some glitches, most notably that if the
+   * cursor was not on a block (e.g. it was on a connection or the
+   * workspace) when the user selected a block then it will remain
+   * visible in its previous location until some keyboard navigation occurs.
+   *
+   * To ameliorate this, the LineCursor constructor adds an event
+   * listener that calls getCurNode in response to SELECTED events.
    *
    * TODO(#97): Remove this hack once Blockly is modified to update
```
rachel-fenichel: Delete "TODO(#97)". This fixes the issue stated in 97 and we'll rewrite much more than just this function when modifying core Blockly for this.

rachel-fenichel (approved):

PR #204, opened 2025-02-14
fix: use correct target node for paste

### LINKED ISSUE #165: Paste at active cursor location

microbit-matt-hillsdon:
This was working as originally implemented but broken by the selection/focus workaround. We now take care to grab the target node before it changes due to the paste.

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/165

Based on @rachel-fenichel's diagnosis of the root cause on #165. I've collapsed `insertPastedBlock` into `paste` as it seemed important to have the getCurrNode call and the paste call near each other so the sequencing was clear.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1356,45 +1356,27 @@ export class Navigation {
    * @returns True if the paste was sucessful, false otherwise.
    */
   paste(copyData: Blockly.ICopyData, workspace: Blockly.WorkspaceSvg): boolean {
-    let isHandled = false;
+    // Do this before clipoard.paste due to cursor/focus workaround in getCurNode.
+    const targetNode = workspace.getCursor()?.getCurNode();
+
     Blockly.Events.setGroup(true);
     const block = Blockly.clipboard.paste(
       copyData,
       workspace,
     ) as Blockly.BlockSvg;
     if (block) {
-      isHandled = this.insertPastedBlock(workspace, block);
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -1356,45 +1356,27 @@ export class Navigation {
...
+    // Do this before clipoard.paste due to cursor/focus workaround in getCurNode.
+    const targetNode = workspace.getCursor()?.getCurNode();
+
     Blockly.Events.setGroup(true);
     const block = Blockly.clipboard.paste(
       copyData,
       workspace,
     ) as Blockly.BlockSvg;
     if (block) {
-      isHandled = this.insertPastedBlock(workspace, block);
```
microbit-matt-hillsdon, line 1366: I think this isHandled bit was wrong as it returns false when insertPastedBlock didn't manage to connect the blocks. But if Blockly.clipboard.paste succeeded then we did handle the action at least by putting the block on the workspace.

rachel-fenichel: Agree since we don't currently do anything different if the block went onto the workspace rather than being connectable; we can always add logic back later if we want to do anything extra to handle that case.

rachel-fenichel (approved):

PR #206, opened 2025-02-17
fix: check isSimpleReporter to be sure of full block field

microbit-matt-hillsdon:
isFullBlockField can be true in scenarios where it is not the only field and so needs visiting. This happens at least for MakeCode if block dummy input clickable image fields (add remove icons).

Screenshot2025_02_14_165155

The MakeCode block has quite a lot of logic but I think the part that relates to this field is simply:

```javascript
  appendDummyInput('ADDBUTTON')
            .appendField(
                new Blockly.FieldImage(this.ADD_IMAGE_DATAURI, 24, 24, "*", addElseIf, false))
```

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -245,9 +245,14 @@ export class LineCursor extends Marker {
         return true;
       case ASTNode.types.INPUT:
         return !(location as Blockly.Connection).isConnected();
-      case ASTNode.types.FIELD:
-        // @ts-expect-error isFullBlockField is a protected method.
-        return !(location as Blockly.Field).isFullBlockField();
+      case ASTNode.types.FIELD: {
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):
Makes sense--it's hitting `isFullBlockField` in core, which looks like this:

```
  /**
   * Defines whether this field should take up the full block or not.
   *
   * Be cautious when overriding this function. It may not work as you expect /
   * intend because the behavior was kind of hacked in. If you are thinking
   * about overriding this function, post on the forum with your intended
   * behavior to see if there's another approach.
   */
  protected isFullBlockField(): boolean {
    return !this.borderRect_;
  }
```

and is perhaps better understood as "can be a full block field".

PR #209, opened 2025-02-18
feat: Display action menu, containing insert action, for stack connections

cpcallen:
Have the context menu shortcut display an action menu, containing for now only an insert action, on stack `NEXT`/`PREVIOUS` connection nodes.

* The `blockInsertAbove` action is renamed `insert` and modified so it will work with a `{connection: Connection}` scope.
* The action menu cribs some code from the shortcut registry to run the `preconditionFn` and `displayText` methods.
* For now the action menu is positioned in the same place as its parent block's context menu would be—to be improved in a following PR.

Part of #184.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1294,33 +1294,34 @@ export class Navigation {
   }
 
   /**
-   * Show the action menu for a given node.
+   * Show the action menu for the current node.
    *
    * The action menu will contain entries for relevant actions for the
    * node's location.  If the location is a block, this will include
    * the contents of the block's context menu (if any).
+   *
+   * Returns true if it is possible to open the action menu in the
+   * current location, even if the menu was not opened due there being
+   * no applicable menu items.
    */
-  openActionMenu(node: Blockly.ASTNode) {
-    const fakeEvent = fakeEventForNode(node);
-    (node.getLocation() as Blockly.BlockSvg).showContextMenu(fakeEvent);
-
+  openActionMenu(workspace: Blockly.WorkspaceSvg): boolean {
     let menuOptions: Array<
       | Blockly.ContextMenuRegistry.ContextMenuOption
       | Blockly.ContextMenuRegistry.LegacyContextMenuOption
     > | null = null;
     let rtl: boolean;
-    let workspace: Blockly.WorkspaceSvg;
 
+    const cursor = workspace.getCursor();
+    if (!cursor) throw new Error('workspace has no cursor');
+    const node = cursor.getCurNode();
     const nodeType = node.getType();
     switch (nodeType) {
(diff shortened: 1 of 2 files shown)


src/navigation.ts
```diff
@@ -1449,6 +1501,36 @@ function fakeEventForNode(node: Blockly.ASTNode): PointerEvent {
...
+ * Create a fake PointerEvent for opening the action menu for the
+ * given ASTNode of type NEXT or PREVIOUS.
+ *
+ * For now this just puts the action menu in the same place as the
+ * context menu for the source block.
+ *
+ * @param node The node to open the action menu for.
+ * @returns A synthetic pointerdown PointerEvent.
+ */
+function fakeEventForStackNode(node: Blockly.ASTNode): PointerEvent {
```
rachel-fenichel, line 1514: Why is this `fakeEventForStackNode` and not `fakeEventForConnectionNode`?

src/navigation_controller.ts
```diff
@@ -983,27 +987,33 @@ export class NavigationController {
...
+          : scope.connection
+            ? 'Insert block here'
+            : 'Insert',
+      preconditionFn: (scope: Scope) => {
+        const block = scope.block ?? scope.connection?.getSourceBlock();
+        const ws = block?.workspace as WorkspaceSvg | null;
         if (!ws) return 'hidden';
 
-        if (!scope.block?.previousConnection) return 'hidden';
+        // if (!scope.block?.previousConnection) return 'hidden';
```
rachel-fenichel: Delete or reenable.

src/navigation.ts
```diff
@@ -1294,55 +1294,90 @@ export class Navigation {
...
         // End reimplement.
         break;
+
+      // case Blockly.ASTNode.types.INPUT:
+      case Blockly.ASTNode.types.NEXT:
+      case Blockly.ASTNode.types.PREVIOUS:
+        const connection = node.getLocation() as Blockly.Connection;
+        rtl = connection.getSourceBlock().RTL;
+
+        // Slightly hacky: get insert action from registry.  Hacky
```
rachel-fenichel, line 1343: I added a note in https://github.com/google/blockly-keyboard-experimentation/issues/155 to add connection as a scope type.

src/navigation.ts
cpcallen: Because there needs to be an equivalent one for input nodes.  (Maybe the two will end up being folded together, but I didn't want to assume that from the outset.)

rachel-fenichel: I'd rather have one method to handle all of the connections, rather than using "stack" (which has had a different meaning in other places in this code).

Issue #222, opened 2025-02-20
Flyout auto-closes after creating a variable


rachel-fenichel:
When making a variable with the “make a variable button” in the flyout, the flyout closes–so you have to reopen it to go get the block, even though you obviously want to go select the variable block immediately.
Probably triggered by an overly-eager `hideChaff` call

BenHenning:
I need to check again, but I think #240 may fix this.

Issue #142, opened 2025-01-10
Use focus to denote cursor location


cpcallen:
## TL;DR

Moving the cursor should should move (browser) focus, and vice versa.

### History

This plugin uses `LinceCursor`, which is a kind of `Marker`, to denote the location the user has navigated to.  Originally it always used the marker's dawer* (a [`MakerSvg`](https://github.com/google/blockly/blob/a42c2d15082d3261a643f205bae1c7860ba48416/core/renderers/common/marker_svg.ts#L45) instance) to denote the location of the cursor.

When marking a block, `MarkerSvg` draws a rectangle around the entire block _and all it's child blocks_.  This behaviour seems unintuitive, especially when navigating stacks of statement blocks—we'd like just the individual block to be marked.  In preparation for user testing on October 28 2024 we decided to fix this, but because changing `MarkerSvg` was non-trivial @rachel-fenichel opted to make a quick, kludgy fix #76 to use the existing block selection code to denote the cursor, when the cursor was on a block.

This was successful, insofar as the students participating in user testing generally did not get confused about the cursor location, and in post-testing discussions we realised that we probably _do_ in most cases want navigating to a block to select that block: it is natural to navigate to a block and then use the delete/cut/copy shortcuts, in expectation that they will apply to that block (without the need to first place the marker or do something else to explicitly select the block).

\* N.B.: that's draw-er—as in something that draws—not [drawer](https://en.wikipedia.org/wiki/Drawer)!

### Focus vs. Selection

It is helpful to [distinguish focus from selection](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusvsselectionandtheperceptionofdualfocus)—in short, in the abstract:

- Only one item can have focus at a time; there should be a visible focus indicator at all times.
- Zero, one, or multiple items can be selected; there should be a distinct visual indicator to indicate which item(s) are selected.
- Moving focus may or may not change what is selected, e.g.:
  - In a typical spreadsheet, moving focus to a different cell selects that cell and unselects any previously-selected ones.
  - In a multiselect list box, moving focus does not change which items are selected.

Currently core Blockly does not have a distinct concept of "focus", and restricts selection to at most one block at a time.  This plugin conflates focus and selection by using selection to denote the location of the cursor.

### Introduce focus; unify with cursor

We should introduce formal support for focus, separately from selection.
[...]

PR #225, opened 2025-02-21
feat: Make toolbox & flyout properly focusable.

### LINKED ISSUE #149: Toolbox with categories takes focus but doesn't indicate it or have working nav
https://category-toolbox.blockly-keyboard-experimentation.pages.dev/ is this plugin [but with the toolbox modified to have categories](https://github.com/google/blockly-keyboard-experimentation/compare/main...microbit-matt-hillsdon:blockly-keyboard-experimentation:category-toolbox?expand=1).

The toolbox is currently an invisible tab position before the workspace. When focussed keyboard navigation doesn't work (as the workspace is not focussed). The navigation works when starting the interaction via the workspace.

I think this should be a real tab position with a focus indicator and working keyboard navigation when focussed, but would be good to get @kmcnaught's input here too. Complicated for now because focus has to be on the workspace for keyboard navigation to work in the toolbox.

Please can we also consider enhancing the demo to have a toolbox variant and more varied blocks? Or is there some other testing environment we can use to test the toolbox nav?

### LINKED ISSUE #180: Remove spurious tab stop before workspace
At the moment there's an extra tab stop between the the rest of the page and the tab stop at which keyboard navigation works.  This needs to be removed.

BenHenning:
Fixes #180
Fixes #149
Fixes part of #101

The toolbox already had a tab stop, but it wouldn't become properly enabled for keyboard navigation. Now it is properly enabled and works when focused both via tab navigation and with the 'T' shortcut.

This has fairly substantial behavioral differences now:
- Focus has been reworked to logically support _either_ the toolbox _or_ the workspace currently having DOM focus in order to support the toolbox properly having keyboard navigation support when enabled.
- Keyboard navigation provided by core Blockly for toolbox has been monkey patched out since it would otherwise interfere with the toolbox navigation (and results in items being skipped when navigating through the categories).
- The navigation controller now has support for a 'nowhere' state to support blurring the toolbox.
- It's possible for the toolbox to have keyboard input with or without DOM focus which can result in some quirkiness when tabbing between the two--this may be something to iterate on if it introduces any user confusion.

From a long-term perspective, we may want to:
- Limit some shortcuts to only cases when the workspace is focused (there's a behavior difference with either approach compared to the currently implementation).
- Figure out a good way to ensure that these behaviors are properly captured in core Blockly as this plugin is moved over.

Separately, note that:
- Per feedback from @rachel-fenichel this only handles toolbox focus and ignores flyouts altogether. Proper flyout support can be added later as needed.
- A subcategory was added to the demo toolbox categories as part of the analysis from #149 (see the second video below).

The new behavior can be seen here:

[Screen recording 2025-03-11 4.59.00 PM.webm](https://github.com/user-attachments/assets/b123d90d-69fb-42a7-b3f6-3ba173392264)

Note that the specific behaviors to observe are around how tabbing to the toolbox allows it to be focused and have keyboard navigation enabled. The video also demonstrates some of the awkward behaviors of switching focus between the toolbox and workspace when the other one is actually focused (such as when inserting a block from a focused toolbox or using the 't' command from an active workspace).

Note that this video also shows how subcategories are currently a bit buggy with left/right navigation (an existing issue now being tracked in #241):

https://github.com/user-attachments/assets/90b89b0d-0242-4863-8414-b4383b68c8e5


BenHenning:
@RoboErikG PTAL for approval.

@rachel-fenichel could you PTAL for the behavior in the video to make sure it looks good? I plan to ask @microbit-matt-hillsdon to take a look at this once it's merged, as well, to ensure it has good compatibility with MakeCode's toolbox.

BenHenning:
TODO: Self, double check if `||`'s should be replaced with `??`'s.

microbit-matt-hillsdon:
Interestingly this changes the behaviour for #222 and the flyout now switches from Variables to the first category after creating a variable.

We're a little worried this one might be hard for us to integrate ahead of user testing due to the toolbox interactions and we don't want to lose the ability to take other changes. We'll explore that today and get back to you.

microbit-matt-hillsdon:
> We're a little worried this one might be hard for us to integrate ahead of user testing due to the toolbox interactions and we don't want to lose the ability to take other changes. We'll explore that today and get back to you.

We've since done a test merge of this change and it's neutral from the perspective of the MakeCode toolbox.

I was hopeful that if the flyout learnt to take focus we could simplify some of what we've done there. At the moment when we hand off from the MakeCode toolbox to the flyout (right arrow) we DOM focus the workspace and call navigation.focusFlyout so that we have working keyboard nav in the flyout. It feels weird to me to see a focus outline on the workspace due to that. Long term I was hoping that could just be us calling focus on the flyout. I think it could be useful for us to discuss the MakeCode toolbox hack we have now in the light of the plan on #142.

RoboErikG:
I'm not sure I understand the keyboard navigation pieces enough yet to do a quick review. Reassigning in case there's time pressure to get this in.

src/index.ts
```diff
@@ -82,14 +82,26 @@ export class KeyboardNavigation {
...
     };
     this.blurListener = () => {
-      this.navigationController.setHasFocus(workspace, false);
+      this.navigationController.updateWorkspaceFocus(workspace, false);
     };
 
     workspace.getSvgGroup().addEventListener('focus', this.focusListener);
     workspace.getSvgGroup().addEventListener('blur', this.blurListener);
+
+    const toolbox = workspace.getToolbox();
```
rachel-fenichel, line 107: Also remove these in `dispose` below.

src/navigation_controller.ts
```diff
@@ -148,33 +159,65 @@ export class NavigationController {
...
-   * @param workspace the workspace that now has input focus.
+   * @param workspace the workspace that now has toolbox/flyout input focus.
    * @param isFocused whether the environment has browser focus.
    */
-  setHasFocus(workspace: WorkspaceSvg, isFocused: boolean) {
-    this.hasNavigationFocus = isFocused;
+  updateToolboxOrFlyoutFocus(workspace: WorkspaceSvg, isFocused: boolean) {
     if (isFocused) {
-      this.navigation.focusWorkspace(workspace, true);
+      if (!workspace.getToolbox()) {
```
rachel-fenichel: nit: I would flip this to
```
if (workspace.getToolbox()) {
  this.navigation.focusToolbox(workspace);
} else {
  this.navigation.focusFlyout(workspace);
}
```

Fortunately none of the workspace we're testing with have neither a flyout nor a toolbox.

src/navigation_controller.ts
```diff
@@ -148,33 +159,65 @@ export class NavigationController {
...
+   * may be reset if it hasn't already been positioned in the workspace.
+   *
+   * @param workspace the workspace that now has workspace input focus.
+   * @param isFocused whether the environment has browser focus.
+   */
+  updateWorkspaceFocus(workspace: WorkspaceSvg, isFocused: boolean) {
+    if (isFocused) {
+      this.navigation.focusWorkspace(workspace, true);
+      this.navigationFocus = NAVIGATION_FOCUS_MODE.WORKSPACE;
+    } else this.navigationFocus = NAVIGATION_FOCUS_MODE.NONE;
```
rachel-fenichel: Style: Please use brackets even for a one-liner. We only skip them for `if (x) return;`

src/navigation.ts
```diff
@@ -537,14 +537,34 @@ export class Navigation {
...
+  blurToolboxAndFlyout(workspace: Blockly.WorkspaceSvg) {
+    workspace.hideChaff();
+    const reset = !!workspace.getToolbox();
+
+    this.resetFlyout(workspace, reset);
+    switch (this.getState(workspace)) {
+      case Constants.STATE.FLYOUT:
+      case Constants.STATE.TOOLBOX:
+        // Clear state since neither the flyout nor toolbox are focused anymore.
+        this.setState(workspace, Constants.STATE.NOWHERE);
```
rachel-fenichel, line 501: Will this ever be called when state needs to move directly to the workspace, instead of to `NOWHERE`?

src/navigation_controller.ts
```diff
@@ -44,6 +45,16 @@ interface Scope {
...
 }
 
+/** Represents the current focus mode of the navigation controller. */
+enum NAVIGATION_FOCUS_MODE {
+  /** Indicates that no interactive elements of Blockly currently have focus. */
+  NONE = 'none',
+  /** Indicates either the toolbox or a flyout has focus. */
+  TOOLBOX_OR_FLYOUT = 'toolbox_or_flyout',
+  /** Indicates that the main workspace currently has focus. */
+  WORKSPACE = 'workspace',
```
rachel-fenichel, line 47: Handy since there's a chance we'll need to add `modal` or `widgetdiv` or similar as options.

src/index.ts
```diff
@@ -82,14 +82,26 @@ export class KeyboardNavigation {
...
+    if (toolbox != null && toolbox instanceof Blockly.Toolbox) {
+      const contentsDiv = toolbox.HtmlDiv?.querySelector('.blocklyToolboxContents');
+      contentsDiv?.addEventListener('focus', () => {
```
rachel-fenichel: You've tried to generalize this to handle either toolbox or flyout, but does this selector actually work in the case where there is a flyout with no toolbox?

We're only testing with category toolboxes so I'm also fine with this code just setting focus on the toolbox and not trying to handle flyouts--would rather have that instead of generalized code that doesn't work in one of two cases.

src/navigation_controller.ts
rachel-fenichel: And see comment in `src/index.ts` about potentially just making this handle the toolbox case and skipping the flyout case if you haven't tested it.

rachel-fenichel (commented):
This makes sense to me logically.

In the screencast I see that the text on all blocks in the flyout is highlighted as though selected. Please fix.

src/index.ts
BenHenning: Done.

src/index.ts
BenHenning: I was mostly aiming to mimic existing logic around how the toolbox was handled elsewhere (such as in `openToolboxOrFlyout`), but I have not tested the case of a flyout with no toolbox. I think it makes sense to simplify the solution here to keep it specific to toolboxes--done.

src/navigation.ts
BenHenning: Not directly from what I've observed. The logic here is largely relying on the focus/blur events to be handled in the correct order at the DOM level. When tab navigating from a focused toolbox to the workspace this method will be called to reset state to nowhere, then the workspace will receive focus and this state will immediately be updated to be workspace.

We needed an intermediary state because the assumption that workspace is always 'next' is also false (when shift tabbing from the toolbox).

I'm open to suggestions on alternate designs or ways of making this flow a bit clearer since I recognize that it is a bit obscured currently.

src/navigation_controller.ts
BenHenning: Flipped the logic at first, but then changed it a bit more due to the other comment and removing explicit flyout support.

src/navigation_controller.ts
BenHenning: Ah, I was unsure on this case. :) Fixed.

BenHenning:
> > We're a little worried this one might be hard for us to integrate ahead of user testing due to the toolbox interactions and we don't want to lose the ability to take other changes. We'll explore that today and get back to you.
> 
> We've since done a test merge of this change and it's neutral from the perspective of the MakeCode toolbox.
> 
> I was hopeful that if the flyout learnt to take focus we could simplify some of what we've done there. At the moment when we hand off from the MakeCode toolbox to the flyout (right arrow) we DOM focus the workspace and call navigation.focusFlyout so that we have working keyboard nav in the flyout. It feels weird to me to see a focus outline on the workspace due to that. Long term I was hoping that could just be us calling focus on the flyout. I think it could be useful for us to discuss the MakeCode toolbox hack we have now in the light of the plan on #142.

Thanks for checking @microbit-matt-hillsdon. Just to clarify: is there anything else you specifically need here ahead of the user test, or do you think the changes here are sufficient & compatible?

Re: the focus outline, I agree completely. It's confusing that the toolbox can hold focus but the workspace be receiving input and vice versa (both of which states are now made possible with this PR). The planned changes in #142 should fix that, though there will need to be a fair amount of reworking of the plugin to rely more on focus and less on presuming focus (i.e. Blockly elements need to actually have focus not just the indicator), but that's expected as part of the broader focus work.

BenHenning (commented):
Self-reviewed latest changes.

Thanks @rachel-fenichel. PTAL.

src/navigation.ts
rachel-fenichel: > The logic here is largely relying on the focus/blur events to be handled in the correct order at the DOM level. When tab navigating from a focused toolbox to the workspace this method will be called to reset state to nowhere, then the workspace will receive focus and this state will immediately be updated to be workspace.

Got it. Good to keep in mind as a constraint, in case we later find that the events are coming in in a different order in some configurations, but this explanation makes sense to me.

src/toolbox_monkey_patch.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+Blockly.Toolbox.prototype.onKeyDown_ = function () {
```
rachel-fenichel, line 9: In this PR or in a followup (file an issue for it), you should make this only happen on install/uninstall instead of happening on load.

rachel-fenichel (approved):
LGTM. Thanks for simplifying out the flyout case for now--please also file a bug to follow up on the flyout case, if we don't already have one.

microbit-matt-hillsdon:
> Thanks for checking @microbit-matt-hillsdon. Just to clarify: is there anything else you specifically need here ahead of the user test, or do you think the changes here are sufficient & compatible?

There's nothing further we need here, thanks. If it is going to be merged before UT then the sooner the better from our point of view. We can live without it if needed just because the MakeCode toolbox already had these behaviours (and we've aligned them where this PR differed).

src/toolbox_monkey_patch.js
BenHenning: Done, though just to check installation is a bit awkward since it needs to happen even before NavigationController is created--does this seem reasonable?

BenHenning:
Thanks @rachel-fenichel. I followed up with a fix for the monkey patch installation--any concerns? It seems a bit awkward to me due to how early in the application lifecycle it needs to be 'installed'.

Re two earlier comments:
- "please also file a bug to follow up on the flyout case, if we don't already have one." -- which flyout case do you mean?
- "In the screencast I see that the text on all blocks in the flyout is highlighted as though selected." -- is this the flyout case? Regardless, I'm not sure that I fully understand what's observed as broken. Approximately which part in the video do you see the broken behavior (I admittedly should have slowed it down originally since it navigates through parts of the UI a bit quickly).

rachel-fenichel:
The text on the blocks (if, do, etc) looks like it has been selected with the mouse on all blocks in the flyout. It looks like that's true for all categories.

blocks in the flyout with text highlighted

Also I shouldn't have merged my change before yours--sorry about the merge conflict >.<

test/index.ts
```diff
@@ -83,6 +84,7 @@ function createWorkspace(): Blockly.WorkspaceSvg {
     renderer,
   };
   const blocklyDiv = document.getElementById('blocklyDiv')!;
+  ToolboxMonkeyPatch.install();
```
rachel-fenichel: Oh, I didn't mean to move this all the way up into `test/index.ts`. Instead, call `install` from `NavigationController.init()` and `uninstall` from `NavigationController.dispose()`

rachel-fenichel: However, I think that I'm distracting from the point of this PR by focusing on where that code goes, and it's fine to merge as is.

rachel-fenichel (dismissed):

rachel-fenichel:
"the flyout case" meaning the case that there is a flyout with no toolbox.

test/index.ts
BenHenning: Reverted the logic back to the way it was before since installation is messy here, and the monkey patches can go away soon (once v12 is available).

BenHenning (commented):
Self-reviewed latest changes.

@rachel-fenichel thanks for the extra details on the text issue. It seems that it mysteriously went away after syncing to latest, so that's fun. :) If you could, PTAL at the latest changes just to make sure everything seems reasonable. I've enabled auto-merge so feel free to merge this, as well, if there are no follow-ups. The first video in the description has also been updated accordingly.

BenHenning:
Hmm GitHub is very mysterious. I apparently don't need to get re-approval after dismissing approval to merge. Well, please go ahead and merge this upon approval if it looks good @rachel-fenichel otherwise I will tomorrow AM.

rachel-fenichel (approved):

rachel-fenichel:
Changes LGTM.

Issue #212, opened 2025-02-19
Passive focus rendering on blocks and connections


rachel-fenichel:
The current placeholder rendering is a solid blue line for a next indicator, which is controlled by `passive_focus.ts`, and the addition of selection on a block (orange outline).

![blue next connection marker on a block stack](https://github.com/user-attachments/assets/8b783f0b-42ec-4136-9323-feb907104b44)

![orange block marker on a block stack](https://github.com/user-attachments/assets/7dce5843-8592-47fa-9789-fb5bb71746c6)

The new focus/selection code allows CSS changes to indicate active and passive focus, but does not allow changes in the block size and shape that would trigger a render. 

@cpcallen proposed that "The marker should be depicted by opening a small gap in the stack (for next statement connections), or temporarily displacing the attached value block (for value inputs), or by an insertion marker", but that requires rendering.

rachel-fenichel:
Rendering on a block is controlled by [`showAtBlock`](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/passive_focus.ts#L95) and [`hideAtBlock`](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/passive_focus.ts#L107) in `passive_focus.ts`.

rachel-fenichel:
There is currently no passive focus indicator for fields, but it can be implemented in a parallel way to #236 and by adding logic to [`show`](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/passive_focus.ts#L69) in `passive_focus.ts`.

maribethb:
We now have better passive focus, and because it's css host applications can change it if they don't like ours.

PR #226, opened 2025-02-21
fix: Ensure visual indicating for focus loss.

### LINKED ISSUE #92: No visual indication when keyboard navigation disabled
At the moment, when you press 'enter' to enable keyboard navigation the cursor appears, but there is no corresponding visual indication when keyboard navigation is disabled (i.e. you press 'esc').

There should be a clear visible change when keyboard navigation is disabled; at minimum, the cursor should be hidden.

Currently we are using block selection as the main visual indication the cursor is on a block.  If we continue to do that we should have an additional visual indicator (e.g. a small arrow) to denote focus; this additional indicator can be hidden even if the block remains selected (and would additionally distinguish the focused block in the event that multiselect is supported).

In the event that we continue to oblige use of enter/esc to enable and disable keyboard navigation, pressing esc should not result in the workspace focus indicator (blue / black outline) disappearing—in fact, perhaps it should _reappear_ when esc is pressed, having previous disappeared (in favour of the cursor) when enter was pressed.

BenHenning:
Fixes #92

If the workspace loses focus, any existing block selection or cursor visuals should be hidden to represent that the focus has been lost. Instead, the passive indicator should be shown so that the user has context on where they are returning when navigating back to the workspace.

This changes the hide behavior for the line cursor to also clear any block selection that may be present (since the cursor is using selection as a means of visually highlighting blocks today).

There was an issue restoring selection due to the passive focus indicator interfering with block selection (since it was using selection). This was updated to use non-selection CSS to keep the indicator without affecting the selection state. This pulls in the changes from #236.

See video for the new state:

[Screen recording 2025-03-11 4.15.41 PM.webm](https://github.com/user-attachments/assets/2b82f8a2-4e16-471f-b579-e0908d8dc6df)

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1139,7 +1139,7 @@ export class Navigation {
     // Although it seems like this should never happen, the typings are wrong
     // in the base Marker class and this can therefore be null.
     if (this.markedNode) {
-      // this.passiveFocusIndicator.show(this.markedNode);
+      this.passiveFocusIndicator.show(this.markedNode);
     }
   }
 
@@ -1149,7 +1149,7 @@ export class Navigation {
    * @param workspace The workspace.
    */
   removeMark(workspace: Blockly.WorkspaceSvg) {
-    // this.passiveFocusIndicator.hide();
+    this.passiveFocusIndicator.hide();
     this.markedNode = null;
   }
 
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -544,7 +544,7 @@ export class Navigation {
    *  - Resume editing by returning the cursor to its previous location, if any.
    *  - Move the cursor to the top connection point on on the first top block.
    *  - Move the cursor to the default location on the workspace.
-   * 
+   *
    * @param workspace The main Blockly workspace.
    * @param keepPosition Whether to retain the cursor's previous position.
    */
@@ -1139,7 +1139,7 @@ export class Navigation {
     // Although it seems like this should never happen, the typings are wrong
     // in the base Marker class and this can therefore be null.
     if (this.markedNode) {
-      this.passiveFocusIndicator.show(this.markedNode);
+      // this.passiveFocusIndicator.show(this.markedNode);
     }
   }
 
@@ -1149,7 +1149,7 @@ export class Navigation {
    * @param workspace The workspace.
    */
   removeMark(workspace: Blockly.WorkspaceSvg) {
-    this.passiveFocusIndicator.hide();
+    // this.passiveFocusIndicator.hide();
     this.markedNode = null;
   }
 
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -559,6 +559,15 @@ export class Navigation {
     }
 
     if (this.markedNode) {
+      // Note that this hide happens twice, one before setCurNode() and once in
+      // removeMark. The latter is actually a logical no-op because setCurNode()
+      // will trigger a selection update of the currently marked node (if it's a
+      // block) and that, in turn, clones the underlying block's
+      // pathObject.svgPath. Since svgPath is updated to remove any passive
+      // focus indicator after selection clones it, the effect of removing the
+      // indicator doesn't do anything (hence it needs to be done *before*
+      // selection is added in order to immediately take effect).
+      this.passiveFocusIndicator.hide();
       cursor.setCurNode(this.markedNode);
       this.removeMark(workspace);
       return;
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -555,6 +555,22 @@ export class LineCursor extends Marker {
     this.updateFocusIndication(oldNode, newNode);
   }
 
+  override hide(): void {
+    super.hide();
+
+    // If there's a block currently selected, remove the selection since the
+    // cursor should now be hidden.
+    const curNode = this.getCurNode();
+    if (curNode.getType() === ASTNode.types.BLOCK) {
+      const block = curNode.getLocation() as Blockly.BlockSvg;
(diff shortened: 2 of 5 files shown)


BenHenning:
@rachel-fenichel would be nice to get your thoughts on the experience. From what I can currently tell, the passive indicator doesn't seem 100% consistent as to what shows when in the toolbox (or outside keyboard nav focus entirely), so it would be helpful to get clarity on whether the video above is ideal or, if not, what the ideal experience should be.

@cpcallen would like your thoughts on the passive indicator as well--I lack the context as to how it's meant to work. It confuses me that we're using the selection indicator for representing 'passive' focus at the moment since it's the same visual representation as active (and thus doesn't really communicate a difference to the user).

src/navigation.ts
```diff
@@ -1139,7 +1139,7 @@ export class Navigation {
     // Although it seems like this should never happen, the typings are wrong
     // in the base Marker class and this can therefore be null.
     if (this.markedNode) {
-      this.passiveFocusIndicator.show(this.markedNode);
+      // this.passiveFocusIndicator.show(this.markedNode);
```
BenHenning: TODO: Get resolution from Rachel and/or Christopher before finalizing this solution and moving PR to review.

microbit-matt-hillsdon:
I didn't expect us to lose the context of where a block would go when the flyout is opened. Previously there was a mixed approach to this, with the blue marker for next/previous and what looks like the normal cursor outline for other positions. But I think it did an important job.

The UI now looks the same in two scenarios: starting from the toolbox & starting from e.g. a next connection on the workspace, but inserting a block with Enter will behave differently.

Absent another proposal, I think we see active focus as outline around the area + the cursor/selection rendering, and passive focus as just the latter.

rachel-fenichel:
The passive focus indicator should show when the active focus moves to the flyout/toolbox. Agree that losing it seems like a problem. I'm fine with Matt's proposal that "we see active focus as outline around the area + the cursor/selection rendering, and passive focus as just the latter".

I implemented the passive focus indicator (as it currently exists today) and can give the context you asked for.

The passive focus indicator is different from the cursor in two ways: 
- The passive focus indicator can be null. That's not true for the cursor and wasn't true for the marker, which is why the marker would persist in annoying locations.
- The target (block or connection) doesn't know about it. That means that if the target block gets updated or rerendered, it doesn't know to force a rerender of the passive focus indicator. 

In implementing the current `passive_focus.ts` my primary goal was to allow it to be set to null. I implemented `showAtBlock` and `hideAtBlock` in a minimal way, by calling the block's `addSelect` and `removeSelect`. That's why it looks the same as the cursor selection, which I agree is confusing. It was never intended to be permanent: https://github.com/google/blockly-keyboard-experimentation/issues/212 tracks making it more distinctive.

I just added a draft PR https://github.com/google/blockly-keyboard-experimentation/pull/236 that makes the passive focus indicator distinctive on blocks, and makes both block and next connection indicators controllable with CSS. Please don't merge it with the current colours, but adding that code into your workspace may make it easier to see when you have passive focus and when you have active focus.

src/navigation.ts
cpcallen: I think we want to continue to display a passive focus indicator here.  I agree that is is suboptimal that it is not distinguishable from the active focus indicator—but I think the fix to that is to change how it looks, not to not display it at all.  @rachel-fenichel ?

BenHenning:
Currently stuck on trying to pull in #236 as it seems that the CSS style cannot always be reliably reset:

[Screen recording 2025-03-05 2.29.35 PM.webm](https://github.com/user-attachments/assets/9fe7f27f-6311-4ccf-9a15-6e6d211829d3)

Enabling the passive focus indicator does correctly show up with a dashed line, but navigating back to it (i.e. refocusing it) causes the dashed line to stay even with the selection indicator from core. What's confusing is that I've verified that the stroke dasharray has been correctly set to 'none' when selection returns, so I have no idea why it's not updating correctly.

src/navigation.ts
BenHenning: This has been fixed in the latest, PTAL.

BenHenning:
Thanks @rachel-fenichel @microbit-matt-hillsdon and @cpcallen. Please take a look at the latest video and follow up if you have any concerns.

Rachel: you mentioned in #236 that we should pick a different color. Might you (or anyone else) have any specific suggestions? I suppose we could use the same selection color now that the (current) passive indicator style is dashed, and maybe lower its stroke width slightly so it's thinner than the selection indicator.

rachel-fenichel:
> I suppose we could use the same selection color now that the (current) passive indicator style is dashed, and maybe lower its stroke width slightly so it's thinner than the selection indicator.

Works for me!

rachel-fenichel (approved):

microbit-matt-hillsdon:
I was initially surprised that switch to passive focus in the workspace when a context menu or field editor is opened. Logically it makes sense, but there's something a bit weird about the feeling that a value block gets "less selected" when you try to interact with it. Not arguing against this, just sharing in case others feel similarly.

image

image

image

[Deployment that I tested (as of 2463303619a2defa305e39cc11cacd85aa8b96c9)](https://indicate-when-workspace-focu.blockly-keyboard-experimentation.pages.dev/).

cpcallen:
@microbit-matt-hillsdon: I agree that this is slightly surprising, but actual UI will probably change once core Blockly has focus support: when the context menu opens, the block will go from active focus to passive focus, _but it should still remain selected_.

BenHenning:
Updated the code & demo video to use a better color (one that @rachel-fenichel suggested--it matches the fully filled indicator).

PR #230, opened 2025-02-21
feat: toasts for move/cut/copy and enter on a block

microbit-matt-hillsdon:
The enter on a block toast shows generic help and we might consider showing it earlier if e.g. the user tabs to the workspace or otherwise focusses it by keyboard.

This logically conflicts with https://github.com/google/blockly-keyboard-experimentation/pull/458 which moves some related text to Blockly.Msg but (obviously!) not the new text. We can merge in either order and clean up after or block one on the other.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -9,7 +9,6 @@ import {
   Events,
   ShortcutRegistry,
   utils as BlocklyUtils,
-  dialog,
 } from 'blockly/core';
 
 import type {
@@ -22,8 +21,12 @@ import type {
 
 import * as Constants from '../constants';
 import type {Navigation} from '../navigation';
-import {getShortActionShortcut} from '../shortcut_formatting';
 import {Mover} from './mover';
+import {
+  showConstrainedMovementHint,
+  showHelpHint,
+  showUnconstrainedMoveHint,
+} from '../hints';
 
 const KeyCodes = BlocklyUtils.KeyCodes;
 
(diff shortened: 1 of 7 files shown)


src/toast.ts
```diff
@@ -0,0 +1,144 @@
+import {WorkspaceSvg} from 'blockly';
```
microbit-matt-hillsdon: I have put no effort into doing things in a Blockly-like way here at this point, or even finding out what that might be.

rachel-fenichel:
@bmxedd do you have UX thoughts on best practices for toasts, particularly in this accessibility context?

microbit-matt-hillsdon:
Rebased on #408 for now.

microbit-robert:
Rebased on main and now uses the new Toast class in Blockly core.

Used in the following cases:
- Copy (cleared on paste)
- Cut (cleared on paste)
- Enter on a block (cleared on toggling help)
- First instance of inserting a block from the flyout with no constrained move positions
- First instance of inserting a block from the flyout with constrained move positions
- Attempting to move a block in constrained move mode when there are no constrained move positions

src/actions/enter.ts
```diff
@@ -150,6 +152,16 @@ export class EnterAction {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     workspace.getCursor()?.setCurNode(ASTNode.createBlockNode(newBlock)!);
     this.mover.startMove(workspace);
+
+    const isTopLevelBlock =
```
gonfunko: Nit: `isStartBlock` or the like might be clearer here; any block can be a top-level block, but this appears to be looking for start/"hat" blocks specifically?

gonfunko (approved):

src/actions/enter.ts
microbit-matt-hillsdon: Thanks, 40927bb6866cf8f5f0c2c186d157d211111cc44e

PR #231, opened 2025-02-21
feat: Scroll flyout to keep cursor in view

### LINKED ISSUE #90: Scroll toolbox flyout to follow cursor
When there are too many blocks in the flyout and you select a block at beyond the screen, the scroll bar will not scroll down automatically and you must use the mouse to scroll down.

```js
const flyout=workspace.getFlyout();
const cursor = flyout ? flyout.getWorkspace().getCursor() : null;
const drawer = cursor.getDrawer();
const blockSvg = drawer ? drawer.getSvgRoot() : null;
blockSvg.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'start' });
```

I tried to use the above code to solve it, but unfortunately it didn't work.

tashee:
Ensure FlyoutCursor is visible when edge scrolling.

Fixes #90

### DIFF
--- a/src/flyout_cursor.ts
+++ b/src/flyout_cursor.ts
@@ -31,13 +31,13 @@ export class FlyoutCursor extends Blockly.Cursor {
    */
   edgeScrollY(newNode: Blockly.ASTNode) {
     const wsHeight = this.flyout.getHeight();
-    const newNodeBlock: Blockly.Block | any = newNode.getSourceBlock();
-    const BlockBottomToTop =
-      newNodeBlock.getRelativeToSurfaceXY().y + newNodeBlock.height + 10;
+    const block = newNode.getSourceBlock() as Blockly.BlockSvg;
+    const blockHeight =
+      block.getRelativeToSurfaceXY().y + block.height + 10;
     const scrollY =
-      BlockBottomToTop > wsHeight
-        ? wsHeight - BlockBottomToTop
-        : BlockBottomToTop;
+    blockHeight > wsHeight
+        ? wsHeight - blockHeight
+        : blockHeight;
     this.flyout.getWorkspace().scroll(0, scrollY);
   }
 
--- a/src/flyout_cursor.ts
+++ b/src/flyout_cursor.ts
@@ -17,11 +17,28 @@ import * as Blockly from 'blockly/core';
  * This cursor only allows a user to go to the previous or next stack.
  */
 export class FlyoutCursor extends Blockly.Cursor {
+  private flyout: Blockly.IFlyout;
   /**
    * The constructor for the FlyoutCursor.
    */
-  constructor() {
+  constructor(private readonly workspace: Blockly.WorkspaceSvg | any) {
     super();
+    this.flyout = this.workspace.targetWorkspace.getFlyout();
+  }
+
+  /**
+   * Set correct height to enable edge scrolling
+   */
+  edgeScrollY(newNode: Blockly.ASTNode) {
+    const wsHeight = this.flyout.getHeight();
+    const newNodeBlock: Blockly.Block | any = newNode.getSourceBlock();
+    const BlockBottomToTop =
+      newNodeBlock.getRelativeToSurfaceXY().y + newNodeBlock.height + 10;
+    const scrollY =
+      BlockBottomToTop > wsHeight
+        ? wsHeight - BlockBottomToTop
+        : BlockBottomToTop;
+    this.flyout.getWorkspace().scroll(0, scrollY);
   }
 
   /**
@@ -38,6 +55,7 @@ export class FlyoutCursor extends Blockly.Cursor {
     const newNode = curNode.next();
 
     if (newNode) {
+      this.edgeScrollY(newNode);
       this.setCurNode(newNode);
     }
     return newNode;
@@ -48,8 +66,17 @@ export class FlyoutCursor extends Blockly.Cursor {
    *
    * @returns Always null.
    */
-  override in(): null {
-    return null;
+  override in(): Blockly.ASTNode | null {
+    const curNode = this.getCurNode();
+    if (!curNode) {
+      return null;
+    }
+    const newNode = curNode.in();
+    if (newNode) {
+      this.edgeScrollY(newNode);
+      this.setCurNode(newNode);
+    }
+    return newNode;
   }
 
   /**
@@ -64,8 +91,8 @@ export class FlyoutCursor extends Blockly.Cursor {
       return null;
     }
     const newNode = curNode.prev();
-
     if (newNode) {
+      this.edgeScrollY(newNode);
       this.setCurNode(newNode);
     }
     return newNode;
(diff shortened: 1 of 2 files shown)


src/flyout_cursor.ts
```diff
@@ -17,11 +17,28 @@ import * as Blockly from 'blockly/core';
  * This cursor only allows a user to go to the previous or next stack.
  */
 export class FlyoutCursor extends Blockly.Cursor {
+  private flyout: Blockly.IFlyout;
   /**
    * The constructor for the FlyoutCursor.
    */
-  constructor() {
+  constructor(private readonly workspace: Blockly.WorkspaceSvg | any) {
```
rachel-fenichel, line 24: Where is this passed into the constructor? I don't see that you updated the constructor call, so either this doesn't work or you don't actually need to pass in the workspace.

src/flyout_cursor.ts
```diff
@@ -17,11 +17,28 @@ import * as Blockly from 'blockly/core';
...
     super();
+    this.flyout = this.workspace.targetWorkspace.getFlyout();
+  }
+
+  /**
+   * Set correct height to enable edge scrolling
+   */
+  edgeScrollY(newNode: Blockly.ASTNode) {
+    const wsHeight = this.flyout.getHeight();
+    const newNodeBlock: Blockly.Block | any = newNode.getSourceBlock();
```
rachel-fenichel: `const newNodeBlock : newNode.getSourceBlock() as Blockly.Block;`

src/flyout_cursor.ts
```diff
@@ -17,11 +17,28 @@ import * as Blockly from 'blockly/core';
...
+    const BlockBottomToTop =
```
rachel-fenichel: Rename: `BlockBottomToTop` -> `blockHeight`

src/flyout_cursor.ts
tashee: unfortunately when rebasing one of the commits got dropped. fixing this.

tashee:
@rachel-fenichel  ready for re-review

microbit-matt-hillsdon:
Interestingly this causes an integration issue in MakeCode where the cursor constructor appears to be too soon to grab the flyout.

It may be that MakeCode is doing something unusual here, but in the short term at least I'd really appreciate deferring fetching the flyout to the point of use along these lines:

```patch
diff --git a/src/flyout_cursor.ts b/src/flyout_cursor.ts
index 7b7a638..b88e06b 100644
--- a/src/flyout_cursor.ts
+++ b/src/flyout_cursor.ts
@@ -17,25 +17,25 @@ import * as Blockly from 'blockly/core';
  * This cursor only allows a user to go to the previous or next stack.
  */
 export class FlyoutCursor extends Blockly.Cursor {
-  private flyout: Blockly.IFlyout;
   /**
    * The constructor for the FlyoutCursor.
    */
-  constructor(private readonly workspace: Blockly.WorkspaceSvg | any) {
+  constructor(private readonly workspace: Blockly.WorkspaceSvg) {
     super();
-    this.flyout = this.workspace.targetWorkspace.getFlyout();
   }
 
   /**
    * Set correct height to enable edge scrolling
    */
   edgeScrollY(newNode: Blockly.ASTNode) {
-    const wsHeight = this.flyout.getHeight();
+    const flyout = this.workspace?.targetWorkspace?.getFlyout();
+    if (!flyout) return;
+    const wsHeight = flyout.getHeight();
     const block = newNode.getSourceBlock() as Blockly.BlockSvg;
     const blockHeight = block.getRelativeToSurfaceXY().y + block.height + 10;
     const scrollY =
       blockHeight > wsHeight ? wsHeight - blockHeight : blockHeight;
-    this.flyout.getWorkspace().scroll(0, scrollY);
+    flyout.getWorkspace().scroll(0, scrollY);
   }
```

microbit-matt-hillsdon:
OK, it's worse than I thought! In the MakeCode case my suggested fix does not help much (it takes it from logging errors to doing nothing).

The cause is MakeCode's custom flyout has a dummy/fake workspace right up until when it's actually open. So grabbing it to create FlyoutCursorClass at init time breaks. I think it's reasonable to merge the approach on this PR and then I'll deal with it with a temporary workaround for user testing and look more carefully at what can be done for the longer term.

src/flyout_cursor.ts
```diff
@@ -17,11 +17,25 @@ import * as Blockly from 'blockly/core';
...
     super();
+    this.flyout = this.workspace.targetWorkspace.getFlyout();
+  }
+
+  /**
+   * Set correct height to enable edge scrolling
+   */
+  edgeScrollY(newNode: Blockly.ASTNode) {
+    const wsHeight = this.flyout.getHeight();
+    const block = newNode.getSourceBlock() as Blockly.BlockSvg;
```
microbit-matt-hillsdon, line 34: This can be null (e.g. label) but is type asserted to be Blockly.BlockSvg. 
Logs an error on the next line when attempting to navigate to a label.

microbit-matt-hillsdon: Presumably buttons and labels (FlyoutButton) should be scrolled into view in a similar way.
For user testing I think we'll be OK to skip this case so long as it does not error.

cpcallen: Good catch.  Issue (and/or PR) welcome as always.

src/flyout_cursor.ts
```diff
@@ -48,8 +63,17 @@ export class FlyoutCursor extends Blockly.Cursor {
    *
    * @returns Always null.
    */
-  override in(): null {
-    return null;
+  override in(): Blockly.ASTNode | null {
+    const curNode = this.getCurNode();
```
microbit-matt-hillsdon, line 67: I think I'm missing why in/out have been added here. I think they will never be called by `NavigationController` in `FLYOUT` state. They do no harm so far as I can tell.

cpcallen:
@microbit-matt-hillsdon:

> It may be that MakeCode is doing something unusual here…

Yes, it sounds like it may be.  In general we expect custom toolboxes / flyouts to be implemented by changing which class constructor is registered, and then allow `BlockSvg.prototype.addFlyout` to retrieve your custom constructor and `new` it.

But IIRC you said MakeCode's flyout isn't even an `IFlyout`, so presumably MakeCode does something completely different that we don't actually support.

As discussed on Tuesday, I anticipate that all cursors will eventually take their corresponding `Workspace` instance as a constructor parameter.  This shouldn't be a problem, because there should be a 1:1 relationship between workspaces and cursors, and (at least for the time being) the flyout cursor should be being created via `Navigation.prototype.addFlyout` (though this _also_ takes an `IFlyout`, so, uh, good luck I guess).

In the medium term making MakeCode's flyout an `IFlyout` seems like the correct course of action, but in the mean time maybe the not-too-awful kludge is to replace the `FlyoutCursor` instance created by `Navigation` for the flyout you don't use with a new one created for real flyout, once the latter is created?

microbit-matt-hillsdon:
For the moment I've removed the caching flyout implementation in MakeCode for the user testing setup so everything's working again. The flyout is an IFlyout (it's the toolbox that isn't an IToolbox) but not a very well behaved one. I need to explore the performance consequences of my change but I hope it's sufficient for user testing. I'll pick this up separately with a new issue/discussion once I have more context.

cpcallen:
> The flyout is an IFlyout (it's the toolbox that isn't an IToolbox) but not a very well behaved one.

Thanks for the correction.

> For the moment I've removed the caching flyout implementation in MakeCode for the user testing setup so everything's working again. I need to explore the performance consequences of my change but I hope it's sufficient for user testing. I'll pick this up separately with a new issue/discussion once I have more context.

Does the caching implementation work with #248?

Closing this as #248 has been merged.

Issue #233, opened 2025-02-21
Paste when flyout is open


rachel-fenichel:
It pastes at the marker! And leaves the flyout open. This is a bit unexpected.

microbit-matt-hillsdon:
Paste now always closes an auto close flyout.
It does paste at the location indicated by passive focus but I'm not sure what else it could do (nothing? - that would be OK I guess because copy moves focus to the workspace).

Issue #232, opened 2025-02-21
Copy behaviour when cursor is not on a copyable entity


rachel-fenichel:
When the cursor is on a block, copying copies that block (and children attached to value inputs).

When the cursor is on a connection, copying should do nothing.

When the cursor is on a field, should copying copy the field value? My inclination is no, because it's hard to show what was copied and because it will cause issues in the full-block-field case.

When the cursor is on a toolbox category, the workspace, of anywhere else that is not a block, copying should do nothing.

cpcallen:
Assigning to @rachel-fenichel to flesh out what a copyable entity is, and what the desirable behaviour is.

maribethb:
we have the behavior we expect

Issue #224, opened 2025-02-20
Organise context menu for user testing with shortcut information


microbit-matt-hillsdon:
## Status quo

In the demo, the context menu for a block contains

- Duplicate
- Add Comment
- Collapse Block
- Disable Block
- Delete N Blocks
- Keyboard Navigation: delete
- Keyboard Navigation: copy
- Insert block above

It's already quite a confusing sequence. Separate issues also plan to add "Cut", "Copy" and "Paste".

## Proposal

How about something like:

- Insert block above (I)  [or more suitable text depending on nature of insert action, if we go with I then depending on font there's potential for I vs L confusion here]
- Delete N Blocks (Del)
- Duplicate
- ---
- Cut  (Cmd + X)
- Copy (Cmd + C)
- Paste (Cmd+V)
- ---
- Add Comment 
- Collapse Block
- Disable Block

Organisation is core keyboard nav actions, copy/paste, less commonly used actions. Likely I've missed some. I don't have a strong view re Duplicate.

On an "Edit" menu it's common for delete to be below paste but I think it makes sense paired with insert here.

Not sure if there's support for separators or how pretty we can make the shortcuts, ideally right aligned like this:

![Image](https://github.com/user-attachments/assets/7c97d8d5-465d-4200-a161-7a48e5426850)

...but anything that exposes them is a bit step forward for user testing.

 Tagging @kmcnaught for input.

kmcnaught:
I'm generally very aligned with your approach @microbit-matt-hillsdon. I'm in favour of re-ordering to favour most common actions, and grouping with separators if possible. 

A few random thoughts..
1. Q: Will the context menu content be different depending if it's triggered from a keyboard or mouse? If so, do we need the same/comparable ordering, or can they be completely independent?
2. Q: Which blocks are intended to have "Collapse Block" ? Currently in MakeCode I've only seen it in top-level blocks, but in blockly currently it's always present. 
3. "Comment": probably needs a shortcut key, once comments are keyboard-navigable. I imagine many people don't comment, but once you start commenting, you probably do it quite a lot. 
[...]

PR #234, opened 2025-02-21
feat: delete and clipboard actions

### LINKED ISSUE #217: Cut option in context menu on blocks
Cut is already a valid keyboard shortcut, but is not exposed through the context menu. We want to expose it, with a hint about the keyboard shortcut, for user testing.

### LINKED ISSUE #218: Copy option in context menu on blocks
Copy is already a valid keyboard shortcut, but is not exposed through the context menu. We want to expose it, with a hint about the keyboard shortcut, for user testing.

### LINKED ISSUE #219: Paste option in context menu on blocks and connections
Paste is already a valid keyboard shortcut, but is not exposed through the context menu. We want to expose it, with a hint about the keyboard shortcut, for user testing.

Paste should be visible on both blocks and connections, unlike cut and copy.

rachel-fenichel:
### User facing changes

Fixes #217 by adding a cut option in the context menu on blocks
Fixes #218 by adding/updating the copy option in the context menu on blocks
Fixes #219 by adding a paste option in the context menu on blocks

Implements parts of #224 by organizing the context menu items and improving text
- Cut, Copy, Paste are in the correct relative order and have shortcut hints: `(Ctrl + X)` or similar
- Delete has a count and a shortcut hint: `(Del)`
- Removes text "`Keyboard Navigation: ` from copy and delete options
- Removes duplicate delete option
- Moves insert action above clipboard actions

### Additional bugs

While testing I found #233 and raised #232 for more complex copy behaviour. Current behaviour is not well-documented and probably not intentional.

### Structural changes

Moves the delete action to its own file, and the three clipboard actions to a shared file. Both classes have `install` and `uninstall` functions and keep track of state when necessary. This is work toward #155.

### DIFF
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -82,6 +82,100 @@ export class Clipboard {
     ShortcutRegistry.registry.unregister(Constants.SHORTCUT_NAMES.PASTE);
   }
 
+
+  /**
+   * Create and register the keyboard shortcut for the cut action.
+   */
+  private registerCutShortcut() {
+    const cutShortcut: ShortcutRegistry.KeyboardShortcut = {
+      name: Constants.SHORTCUT_NAMES.CUT,
+      preconditionFn: this.cutPrecondition.bind(this),
+      callback: this.cutCallback.bind(this),
+      keyCodes: [
+        createSerializedKey(KeyCodes.X, [KeyCodes.CTRL]),
+        createSerializedKey(KeyCodes.X, [KeyCodes.ALT]),
+        createSerializedKey(KeyCodes.X, [KeyCodes.META]),
+      ],
+      allowCollision: true,
+    };
+
+    ShortcutRegistry.registry.register(cutShortcut);
+  }
+
+  /**
+   * Register the cut block action as a context menu item on blocks.
+   * This function mixes together the keyboard and context menu preconditions
+   * but only calls the keyboard callback.
+   */
+  private registerCutContextMenuAction() {
+    const cutAction: ContextMenuRegistry.RegistryItem = {
+      displayText: (scope) => `Cut (${this.getPlatformPrefix()}X)`,
+      preconditionFn: (scope) => {
+        const ws = scope.block?.workspace;
+        if (!ws) return 'hidden';
+
+        return this.cutPrecondition(ws) ? 'enabled' : 'disabled';
+      },
+      callback: (scope) => {
+        const ws = scope.block?.workspace;
+        if (!ws) return;
+        return this.cutCallback(ws);
+      },
+      scopeType: ContextMenuRegistry.ScopeType.BLOCK,
+      id: 'blockCutFromContextMenu',
+      weight: BASE_WEIGHT,
+    };
+
+    ContextMenuRegistry.registry.register(cutAction);
+  }
+
+  /**
+   * Precondition function for cutting a block from keyboard
+   * navigation. This precondition is shared between keyboard shortcuts
+   * and context menu items.
+   *
+   * @param workspace The `WorkspaceSvg` where the shortcut was
+   *     invoked.
+   * @returns True iff `cutCallback` function should be called.
+   */
+  private cutPrecondition(workspace: WorkspaceSvg) {
+    if (this.canCurrentlyEdit(workspace)) {
+      const curNode = workspace.getCursor()?.getCurNode();
+      if (curNode && curNode.getSourceBlock()) {
+        const sourceBlock = curNode.getSourceBlock();
+        return !!(
+          !Gesture.inProgress() &&
+          sourceBlock &&
+          sourceBlock.isDeletable() &&
+          sourceBlock.isMovable() &&
+          !sourceBlock.workspace.isFlyout
+        );
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Callback function for cutting a block from keyboard
+   * navigation. This callback is shared between keyboard shortcuts
+   * and context menu items.
+   *
+   * @param workspace The `WorkspaceSvg` where the shortcut was
+   *     invoked.
+   * @returns True if this function successfully handled cutting.
+   */
+  private cutCallback(workspace: WorkspaceSvg) {
+    const sourceBlock = workspace
+      .getCursor()
+      ?.getCurNode()
+      .getSourceBlock() as BlockSvg;
+    this.copyData = sourceBlock.toCopyData();
+    this.copyWorkspace = sourceBlock.workspace;
+    this.navigation.moveCursorOnBlockDelete(workspace, sourceBlock);
+    sourceBlock.checkAndDelete();
+    return true;
+  }
+
   /**
    * Create and register the keyboard shortcut for the copy action.
    */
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
Notes from Matt for follow-up work: "Feels like Paste should be on the context menu in more places: next/previous connections + statement input (? e.g. in empty draw block), and e.g. the test input on "if".  I think all those work via shortcut."

microbit-matt-hillsdon:
By what I suspect is more luck than judgement, this PR ends up with the context menu more or less sequenced how I suggested on #224 in MakeCode (closer than the demo itself gets):

image

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  ContextMenuRegistry,
+  Gesture,
+  ShortcutRegistry,
+  utils as BlocklyUtils,
```
cpcallen: Since this is a new name I'd probably stick with styleguide advice re: case:

```suggestion
  utils as blocklyUtils,
```

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+   */
+  private canCurrentlyEditFn: (ws: WorkspaceSvg) => boolean;
+
+  /**
+   * Weight for the first of these three items in the context menu.
+   * Changing base weight will change where this group goes in the context
+   * menu; changing individual weights relative to base weight can change
+   * the order within the clipboard group.
+   */
+  private baseWeight = 11;
```
cpcallen: Since this is private and not set anywhere, consider making it a top level `const`:
```
const BASE_WEIGHT = 11;
```

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+   */
+  uninstall() {
+    ContextMenuRegistry.registry.unregister('blockCopyFromContextMenu');
+    ContextMenuRegistry.registry.unregister('blockPasteFromContextMenu');
+    ContextMenuRegistry.registry.unregister('blockCutFromContextMenu');
+
+    ShortcutRegistry.registry.unregister(Constants.SHORTCUT_NAMES.COPY);
+    ShortcutRegistry.registry.unregister(Constants.SHORTCUT_NAMES.CUT);
+    ShortcutRegistry.registry.unregister(Constants.SHORTCUT_NAMES.PASTE);
+  }
```
cpcallen: Nit: inconsistent ordering install vs. uninstall: AaBbCc vs ABCabc.

Also, in most situations (e.g. macOS Edit menu) the ordering is cut / copy / paste (not copy / paste / cut).  Why not use that order in this module?

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+   */
+  private copyPreconditionFn(workspace: WorkspaceSvg) {
+    if (!this.canCurrentlyEditFn(workspace)) return false;
+    switch (this.navigation.getState(workspace)) {
+      case Constants.STATE.WORKSPACE:
+        const curNode = workspace?.getCursor()?.getCurNode();
+        const source = curNode?.getSourceBlock();
+        return !!(
+          source?.isDeletable() &&
+          source?.isMovable() &&
```
cpcallen, line 239: These two are essential for cut, but it's not obvious that you should not be able to copy blocks just because they can't be moved or deleted.  Since this is a holdover from the previous code it's fine for this PR but I think we should open a bug to reconsider this in due course.

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * Register the copy block action as a context menu item on blocks.
+   * This function mixes together the keyboard and context menu preconditions
+   * but only calls the keyboard callback.
+   */
+  private registerCopyContextMenuAction() {
+    const copyAction: ContextMenuRegistry.RegistryItem = {
+      displayText: (scope) => {
+        return 'Copy (' + this.getPlatformPrefix() + ' + C)';
+      },
```
cpcallen: * No need for curlies if body is a single return.
* Prefer template literals.
* On macOS the notation is "⌘C", not "⌘ + C", so any such notation should be included in the return value from `getPlatformPrefix`.

```suggestion
      displayText: (scope) => `Copy (${this.getPlatformPrefix()}C)`,
```

(Here and various places below.)

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * Check the platform and return a prefix for the keyboard shortcut.
+   * TODO: https://github.com/google/blockly-keyboard-experimentation/issues/155
+   * This will eventually be the responsibility of the action code ib
+   * Blockly core.
+   *
+   * @returns A platform-appropriate string for the meta key.
+   */
+  private getPlatformPrefix() {
+    return navigator.platform.startsWith('Mac') ? 'Cmd' : 'Ctrl';
```
cpcallen:
```suggestion
    return navigator.platform.startsWith('Mac') ? '⌘' : 'Ctrl+';
```

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+  private copyData: ICopyData | null = null;
+
+  /** The workspace a copy or cut keyboard shortcut happened in. */
+  private copyWorkspace: WorkspaceSvg | null = null;
+
+  /**
+   * Function provided by the navigation controller to say whether editing
+   * is allowed.
+   */
+  private canCurrentlyEditFn: (ws: WorkspaceSvg) => boolean;
```
cpcallen: Hungarian notation is disrecommended.  A "Fn" suffix might occasionally be disambiguating (e.g. in a context when you have a `fooFn` and the resulting `foo` value), and is regrettably baked in to our API for certain property names, but should otherwise be avoided.

```suggestion
  private canCurrentlyEdit: (ws: WorkspaceSvg) => boolean;
```

(Here and multiple places below—especially for method names like `copyPreconditionFn`, since methods are _always_ functions.)

src/actions/clipboard.ts
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * Precondition function for cutting a block from keyboard
+   * navigation. This precondition is shared between keyboard shortcuts
+   * and context menu items.
+   *
+   * @param workspace The `WorkspaceSvg` where the shortcut was
+   *     invoked.
+   * @returns True iff `cutCallbackFn` function should be called.
+   */
+  private cutPreconditionFn(workspace: WorkspaceSvg) {
```
cpcallen: Probably better to consistently order precondition then callback.

cpcallen (approved):
Some small style and code organisation nits, but otherwise LGTM.

src/actions/clipboard.ts
rachel-fenichel: Done

src/actions/clipboard.ts
rachel-fenichel: Done

src/actions/clipboard.ts
rachel-fenichel: Changed to be in the same order in this function (cut, copy, paste)

src/actions/clipboard.ts
rachel-fenichel: Changed to the one-liner in all places.

Changed `getPlatformPrefix` to `return navigator.platform.startsWith('Mac') ? '⌘' : 'Ctrl + ';`

#155 includes the work for getting the prefix right for each platform.

@microbit-matt-hillsdon @kmcnaught Please confirm that those are the right prefixes for testing.

src/actions/clipboard.ts
rachel-fenichel: Done

src/actions/clipboard.ts
rachel-fenichel: Done, here and below and in `delete.ts`

src/actions/clipboard.ts
```diff
@@ -0,0 +1,368 @@
...
+  private copyPrecondition(workspace: WorkspaceSvg) {
+    if (!this.canCurrentlyEdit(workspace)) return false;
+    switch (this.navigation.getState(workspace)) {
+      case Constants.STATE.WORKSPACE:
+        const curNode = workspace?.getCursor()?.getCurNode();
+        const source = curNode?.getSourceBlock();
+        return !!(
+          source?.isDeletable() &&
+          source?.isMovable() &&
+          !Gesture.inProgress()
```
rachel-fenichel, line 240: This and similar lines cause https://github.com/google/blockly-keyboard-experimentation/issues/237

Filed as a separate bug so that the structural change can land.

src/actions/clipboard.ts
rachel-fenichel: Also reordered in the rest of the code.

src/actions/clipboard.ts
rachel-fenichel: Done.

src/actions/clipboard.ts
rachel-fenichel: I think this was taken from core and should match core. Also, if you copy you get the deletable and movable state, so you could tehn paste blocks that you would be unable to delete or move. Blocking copy is safer.

PR #236, opened 2025-02-22
feat: use CSS for passive focus on blocks and next connections

rachel-fenichel:
Demo code for how to do https://github.com/google/blockly-keyboard-experimentation/issues/212 with CSS for blocks and next connections.

Passive focus on a next connection:

Passive focus on a block:

These can be controlled with the CSS in `test/index.html`, or CSS in the page where you are doing testing.

The use of hot pink with dashes is a cry for help, not a suggestion. It should not be merged with that color--I just wanted to make it easy to see in the screenshots.

### DIFF
--- a/src/passive_focus.ts
+++ b/src/passive_focus.ts
@@ -96,7 +96,11 @@ export class PassiveFocus {
     const block = node.getLocation() as BlockSvg;
     // Note that this changes rendering but does not change Blockly's
     // internal selected state.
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@microbit-matt-hillsdon and @BenHenning please take a look and copy this code into your PRs as appropriate. I do not expect that I will have time to fix and merge this on Monday.

PR #239, opened 2025-02-24
feat: Scroll blocks into bounds when focusing them.

### LINKED ISSUE #215: Basic implementation of scrolling workspace to keep the selected block in view
In the short term, `LineCursor.setCurNode` should call `workspace.centerOnBlock(activeBlockId, true)`. The cursor must identify the active block and its ID.

See the parent issue (#214) for context and open questions. This issue tracks the single task needed for testing.

gonfunko:
This PR fixes #215 by providing a basic function for scrolling blocks into bounds when they receive focus.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -593,19 +593,15 @@ export class LineCursor extends Marker {
     let deltaY = 0;
 
     if (bounds.left < viewport.left) {
-      deltaX += viewport.left - bounds.left;
-    }
-
-    if (bounds.right > viewport.right) {
-      deltaX -= bounds.right - viewport.right;
+      deltaX = viewport.left - bounds.left;
+    } else if (bounds.right > viewport.right) {
+      deltaX = viewport.right - bounds.right;
     }
 
     if (bounds.top < viewport.top) {
-      deltaY += viewport.top - bounds.top;
-    }
-
-    if (bounds.bottom > viewport.bottom) {
-      deltaY -= bounds.bottom - viewport.bottom;
+      deltaY = viewport.top - bounds.top;
+    } else if (bounds.bottom > viewport.bottom) {
+      deltaY = viewport.bottom - bounds.bottom;
     }
 
     deltaX *= scale;
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -552,7 +552,65 @@ export class LineCursor extends Marker {
   override setCurNode(newNode: ASTNode) {
     const oldNode = this.getCurNode();
     super.setCurNode(newNode);
+    if (newNode?.getType() === ASTNode.types.BLOCK) {
+      this.scrollBlockIntoView(newNode.getLocation() as Blockly.BlockSvg);
+    }
     this.updateFocusIndication(oldNode, newNode);
+  }
+
+  /**
+   * Scrolls the provided block into view.
+   *
+   * This is a basic implementation that scrolls just enough to get the block
+   * into bounds. For very small workspaces/high zoom levels the entire block
+   * may not fit, but a decent portion of it should still be visible at least.
+   *
+   * @param block The block to scroll into bounds.
+   */
+  private scrollBlockIntoView(block: Blockly.BlockSvg) {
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
This is great from a user testing/micro:bit perspective, thank you.

Tweaks that we definitely wouldn't delay merging for:

Some small margin might be nice - particularly at the top the cursor can be clipped. E.g. this scenario where we've scrolled "create canvas" into view:
image

Should we consider scrolling more than just the block into view? E.g. this scenario where we've scrolled "draw" into view:
image
I suspect thinking through these cases will complicate things a lot so definitely up for waiting for user feedback. 

Is there scope to reuse what's here for #90? I think this PR has the core behaviour working well.

microbit-matt-hillsdon:
Please can you test the toolbox & block insertion with this PR? The scrolling is great but I'm getting errors when trying to insert blocks via the keyboard:

```
Uncaught TypeError: this.selectChild_ is not a function
    at eg.toolboxHandler (contentscript-extra.js:2:1920200)
    at Object.callback (contentscript-extra.js:2:1915079)
    at Zp.onKeyDown (contentscript-extra.js:2:1479147)
    at Li (contentscript-extra.js:2:1142491)
    at HTMLDivElement.s (contentscript-extra.js:2:1069186)
contentscript-extra.js:2 Uncaught TypeError: this.selectNext_ is not a function
    at eg.toolboxHandler (contentscript-extra.js:2:1920163)
    at Object.callback (contentscript-extra.js:2:1914685)
    at Zp.onKeyDown (contentscript-extra.js:2:1479147)
    at Li (contentscript-extra.js:2:1142491)
    at HTMLDivElement.s (contentscript-extra.js:2:1069186)

```

I've deployed your code here to allow a quick try and to make sure it wasn't a local build issue for me: https://scroll-into-view.blockly-keyboard-experimentation.pages.dev/

The first logged error is on right arrow into the flyout and the second is when I pressed down subsequently (assuming that right would have worked). I wonder if this is related to the Blockly version change.

gonfunko:
Fixed the exceptions, this PR needed to update the underlying Blockly to the latest 11.2.1 to use a block sizing function introduced recently, but that broke some monkeypatching of private APIs that got renamed. Agreed some padding is probably reasonable, I'll take a look at that later if I get a chance too.

microbit-matt-hillsdon:
Working great since the fixes. This is good to merge from my point of view. Padding tweak is nice to have but we'd take this for UK morning tomorrow over waiting longer for that.

src/line_cursor.ts
```diff
@@ -552,9 +552,67 @@ export class LineCursor extends Marker {
...
+      deltaX -= bounds.right - viewport.right;
+    }
+
+    if (bounds.top < viewport.top) {
+      deltaY += viewport.top - bounds.top;
+    }
+
+    if (bounds.bottom > viewport.bottom) {
+      deltaY -= bounds.bottom - viewport.bottom;
+    }
```
cpcallen: The proposed implementation seems like it will deal badly with blocks that are too big to fit in the viewport—possibly even causing scroll to oscillate every time this method is called.

Might it be better to ensure that the top left corner of the block is within the viewport, no matter what?  Something like:

```suggestion
    if (bounds.left < viewport.left) {
      deltaX = viewport.left - bounds.left;
    } else if (bounds.right > viewport.right) {
      deltaX = viewport.right - bounds.right;
    }

    if (bounds.top < viewport.top) {
      deltaY = viewport.top - bounds.top;
    } else if (bounds.bottom > viewport.bottom) {
      deltaY = viewport.bottom - bounds.bottom;
    }
```

(Please verify implementation.)

cpcallen (approved):

src/line_cursor.ts
kmcnaught: Agreed, top left is more natural than bottom right. With Matt's deployment at [scroll-into-view deployment link](https://scroll-into-view.blockly-keyboard-experimentation.pages.dev/) there is no weird oscillating behaviour, but it does end up favouring bottom right.

PR #240, opened 2025-02-24
fix: Widget disposal correctly now redirects focus

### LINKED ISSUE #136: Navigation after field edit broken by recent change

### LINKED ISSUE #222: Flyout auto-closes after creating a variable

BenHenning:
Fixes #136
Fixes #222

TODO: Fill in description.

### DIFF
--- a/src/field_input_monkey_patch.js
+++ b/src/field_input_monkey_patch.js
@@ -1,0 +1,37 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+function addTemporaryMarkFocusedOverrideToWorkspaceSvg() {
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
I realise this is in draft, but we took a look while trying to understand everything that might impact user testing. At the moment:
- this breaks clicking on a block then using keyboard navigation
- we're worried this approach can only work for known field types e.g. it's currently broken for colour and dropdown in the demo

BenHenning:
Replaced by #326.

PR #242, opened 2025-02-25
feat: Display action menu, containing insert action, for input connections

### LINKED ISSUE #184: Context menu at connection location

cpcallen:
Have the context menu shortcut display an action menu, containing for now only an insert action, on `INPUT` connection nodes (for both value and statement inputs).

* `fakeEventForStackNode` is renamed `fakeEventForConnectionNode`, and
* is modified to position the action menu relative to the connection's `.x` and `.y` location (converted from workspace to screen coordinates).

Fixes #184.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This plugin for Blockly enables keyboard navigation. It is intended to
 experiment with different actions that might help visually impaired and motor
 impaired people navigate a Blockly workspace.
 
-Keyboard navigation and screenreader support are closely coupled. The Blockly 
+Keyboard navigation and screenreader support are closely coupled. The Blockly
 team intends to add screenreader support incrementally in Q2 and Q3 of 2025,
 as we validate the general approach to navigation.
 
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
This is working well. 

I think for consistency with other menu items the text should be capitalised as "Insert Block Here" and similar.

Other than that I think we should get it merged for UT.

Broader thoughts / questions:
- [As noted recently by @kmcnaught here](https://github.com/google/blockly-keyboard-experimentation/issues/224#issuecomment-2674954199), we're not convinced that "Insert Block Above" is a reasonable way to build programs. Not trying to change that on this PR though.
- Which issue tracks this action getting a shortcut (and showing it in the action menu)?
- I think it would be great to get paste on this menu too (useful for all the same reasons as insert)

microbit-matt-hillsdon:
Checking my understanding: 

I think the keyboard shortcut for this should come by entirely replacing the current insert entry in navigation_controller, which is confusingly unrelated to the insert context menu entry and currently useless. Something along the lines of DeleteAction that registers the shortcut and the context menu entry (as currently in registerInsertAction).

cpcallen:
> I think for consistency with other menu items the text should be capitalised as "Insert Block Here" and similar.

Done.

> Other than that I think we should get it merged for UT.

Also done.

> Broader thoughts / questions:
> 
> * [As noted recently by @kmcnaught here](https://github.com/google/blockly-keyboard-experimentation/issues/224#issuecomment-2674954199), we're not convinced that "Insert Block Above" is a reasonable way to build programs. Not trying to change that on this PR though.

[I agree, but don't think that's relevant to this menu option](https://github.com/google/blockly-keyboard-experimentation/issues/224#issuecomment-2683058746).

> * Which issue tracks this action getting a shortcut (and showing it in the action menu)?

#246.

> * I think it would be great to get paste on this menu too (useful for all the same reasons as insert)

Ahh, putting paste in the action menu on connections?  Yes, that makes sense.  We could have it be clever and be disabled if the clipboard is empty of if it contains blocks that are not compatible with the current location.  In the mean time, 

> I think the keyboard shortcut for this should come by entirely replacing the current insert entry in navigation_controller…

[Agreed](#246).

PR #243, opened 2025-02-25
feat: Allow using space or enter to activate items.

### LINKED ISSUE #138: Space and Enter should be synonymous
In October's user testing, we found a lot of children used Space as a "confirm/enter" action by default. This works on native elements like the "Run code" button or scenario dropdown, but does _not_ work in any of the block coding UI. 

I believe we are in agreement that Space and Enter should remain synonymous throughout the application. I don't think this has been captured in any other issues.

gonfunko:
This PR fixes #138 by adding `space` as an alias for `enter` for activating/marking elements in the workspace.

### DIFF
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -424,7 +424,9 @@ export class NavigationController {
     enter: {
       name: Constants.SHORTCUT_NAMES.MARK, // FIXME
       preconditionFn: (workspace) => this.canCurrentlyEdit(workspace),
-      callback: (workspace) => {
+      callback: (workspace, event) => {
+        event.preventDefault();
+
         let flyoutCursor;
         let curNode;
         let nodeType;
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -455,7 +455,7 @@ export class NavigationController {
             return false;
         }
       },
-      keyCodes: [KeyCodes.ENTER],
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
Tried space over enter in the demo with this branch.
Some of these might need tracking separately:
- Space to edit canvas width activates and also overwrites with a space (needs a preventDefault fix?). This also applies to text shadow blocks in MakeCode (contrary to my initial version of this comment).
- Colour picker cannot pick an option via space (not relevant to user testing but worth tracking)
- Dropdowns work well

Would be great to see this merged for user testing.

microbit-matt-hillsdon:
Confirmed that this fixes the space overwrite issue.

```patch
diff --git a/src/navigation_controller.ts b/src/navigation_controller.ts
index a269b9b..ada867d 100644
--- a/src/navigation_controller.ts
+++ b/src/navigation_controller.ts
@@ -425,7 +425,9 @@ export class NavigationController {
     enter: {
       name: Constants.SHORTCUT_NAMES.MARK, // FIXME
       preconditionFn: (workspace) => this.canCurrentlyEdit(workspace),
-      callback: (workspace) => {
+      callback: (workspace, e) => {
+        e.preventDefault();
+
         let flyoutCursor;
         let curNode;
         let nodeType;
```

(lines might be off a bit, this is from my user testing integration branch)

gonfunko:
Thanks for the fix! Added that in.

RoboErikG (approved):

PR #245, opened 2025-02-25
feat: Focus the first contextual menu item by default.

### LINKED ISSUE #244: When opening the context menu using the keyboard, the first item in the menu should be selected by default
At the moment, when you open the context menu using the keyboard no item is selected, and you have to press the down arrow to select the first item.

Having the first item selected by default will help make it more clear that you can use the keyboard to choose, as well as making it easy to select the topmost item.

gonfunko:
This PR fixes #244 by firing a synthetic keydown event after a brief delay to select the first item in contextual menus. The contextualmenu module unfortunately doesn't provide a method to get at the backing menu in order to call its selection methods.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1387,6 +1387,20 @@ export class Navigation {
     if (!menuOptions?.length) return true;
     const fakeEvent = fakeEventForNode(node);
     Blockly.ContextMenu.show(fakeEvent, menuOptions, rtl, workspace);
+    setTimeout(() => {
+      Blockly.WidgetDiv.getDiv()
+        ?.querySelector('.blocklyMenu')
+        ?.dispatchEvent(
+          new KeyboardEvent('keydown', {
+            key: 'ArrowDown',
+            code: 'ArrowDown',
+            keyCode: 40,
+            which: 40,
+            bubbles: true,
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -1387,6 +1387,20 @@ export class Navigation {
...
     Blockly.ContextMenu.show(fakeEvent, menuOptions, rtl, workspace);
+    setTimeout(() => {
+      Blockly.WidgetDiv.getDiv()
+        ?.querySelector('.blocklyMenu')
+        ?.dispatchEvent(
+          new KeyboardEvent('keydown', {
+            key: 'ArrowDown',
+            code: 'ArrowDown',
+            keyCode: 40,
+            which: 40,
```
cpcallen: Can use the existing enum, no?:

```suggestion
            keyCode: Blockly.utils.KeyCodes.DOWN,
            which: Blockly.utils.KeyCodes.DOWN,
```

cpcallen (approved):
LGTM but I think this deserves a TODO tagging a bug in core to allow it to be fixed properly in due course.

gonfunko:
Filed [8786](https://github.com/google/blockly/issues/8786) in core to track.

PR #248, opened 2025-02-26
fix: Scroll flyout contents into bounds.

### LINKED ISSUE #90: Scroll toolbox flyout to follow cursor

### LINKED ISSUE #150: Error to console when handling enter on a label in a flyout
If you add content of kind=label to a toolbox category then the keyboard navigation will visit it and pressing enter on it will cause an error on the console.

```
bundle.js:2 Uncaught Error: No callback function found for flyout button.
    at Ht.triggerButtonCallback (bundle.js:2:1888927)
    at Object.callback (bundle.js:2:1899298)
    at Fp.onKeyDown (bundle.js:2:1476389)
    at ai (bundle.js:2:1142398)
    at HTMLDocument.s (bundle.js:2:1068151)
```

There's an example label [in the subcategory on this deployment](https://category-toolbox.blockly-keyboard-experimentation.pages.dev/).

Image

I think the typescript type might be missing the kind=label case.

MakeCode uses labels as headings and subheadings in its toolbox flyouts.

gonfunko:
This PR fixes #90 by updating the flyout to scroll the selected item into view. It refactors similar logic for the main workspace out into a shared function used by both. It also fixes #150.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -1454,7 +1454,7 @@ export class Navigation {
     );
     if (typeof buttonCallback === 'function') {
       buttonCallback(button);
-    } else {
+    } else if (!button.isLabel()) {
       throw new Error('No callback function found for flyout button.');
     }
   }
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -162,7 +162,9 @@ export class Navigation {
       cursorRegistrationName,
     );
     if (FlyoutCursorClass) {
-      flyoutWorkspace.getMarkerManager().setCursor(new FlyoutCursorClass());
+      flyoutWorkspace
+        .getMarkerManager()
+        .setCursor(new FlyoutCursorClass(flyout));
     }
   }
 
--- a/src/workspace_utilities.ts
+++ b/src/workspace_utilities.ts
@@ -1,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Scrolls the provided bounds into view.
+ *
+ * In the case of small workspaces/large bounds, this function prioritizes
+ * getting the top left corner of the bounds into view. It also adds some
+ * padding around the bounds to allow the element to be comfortably in view.
+ *
+ * @param bounds A rectangle to scroll into view, as best as possible.
+ * @param workspace The workspace to scroll the given bounds into view in.
+ */
+export function scrollBoundsIntoView(
+  bounds: Blockly.utils.Rect,
+  workspace: Blockly.WorkspaceSvg,
+) {
+  const scale = workspace.getScale();
+
+  // Add some padding to the bounds so the element is scrolled comfortably
+  // into view.
+  bounds.top -= 10;
+  bounds.bottom += 10;
+  bounds.left -= 10;
+  bounds.right += 10;
+
+  const rawViewport = workspace.getMetricsManager().getViewMetrics(true);
+  const viewport = new Blockly.utils.Rect(
+    rawViewport.top,
+    rawViewport.top + rawViewport.height,
+    rawViewport.left,
+    rawViewport.left + rawViewport.width,
+  );
+
+  if (
+    bounds.left >= viewport.left &&
+    bounds.top >= viewport.top &&
+    bounds.right <= viewport.right &&
+    bounds.bottom <= viewport.bottom
+  ) {
+    // Do nothing if the block is fully inside the viewport.
+    return;
+  }
+
+  let deltaX = 0;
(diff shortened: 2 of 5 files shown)


microbit-matt-hillsdon:
It would be great to support labels/buttons too as this can feel quite broken with MakeCode's label use for headings and flyout sections.

They don't seem to have `getBoundingRectangle` but perhaps it would be easy to add. If not it seems to work with this (I've added one for an example below, you'll have to make a very short browser window though).

[Code also here](https://github.com/google/blockly-keyboard-experimentation/compare/flyout-scroll...microbit-matt-hillsdon:blockly-keyboard-experimentation:label-flyout-scroll) and [deployed here](https://label-flyout-scroll.blockly-keyboard-experimentation.pages.dev/).

```patch
diff --git a/src/flyout_cursor.ts b/src/flyout_cursor.ts
index fcb134c..84eefb5 100644
--- a/src/flyout_cursor.ts
+++ b/src/flyout_cursor.ts
@@ -85,15 +85,21 @@ export class FlyoutCursor extends Blockly.Cursor {
     super.setCurNode(node);
 
     const location = node.getLocation();
+    let bounds: Blockly.utils.Rect | undefined;
     if (
-      !(
-        'getBoundingRectangle' in location &&
-        typeof location.getBoundingRectangle === 'function'
-      )
-    )
-      return;
-
-    const bounds = location.getBoundingRectangle();
+      'getBoundingRectangle' in location &&
+      typeof location.getBoundingRectangle === 'function'
+    ) {
+      bounds = location.getBoundingRectangle();
+    } else if (location instanceof Blockly.FlyoutButton) {
+      const {x, y} = location.getPosition();
+      bounds = new Blockly.utils.Rect(
+        y,
+        y + location.height,
+        x,
+        x + location.width,
+      );
+    }
     if (!(bounds instanceof Blockly.utils.Rect)) return;
 
     scrollBoundsIntoView(bounds, this.flyout.getWorkspace());
diff --git a/test/blocks/toolbox.js b/test/blocks/toolbox.js
index a402ae7..125b755 100644
--- a/test/blocks/toolbox.js
+++ b/test/blocks/toolbox.js
@@ -35,6 +35,10 @@ export const p5CategoryContents = [
       },
     },
   },
+  {
+    kind: 'label',
+    text: "Writing text"
+  },
   {
     kind: 'block',
     type: 'write_text_with_shadow',
```

microbit-matt-hillsdon:
[Screen recording of the MakeCode flyout so you can see my motivation for adding label support](https://github.com/user-attachments/assets/4bbedd22-5aea-4c43-87b1-5271b8a933da).

gonfunko:
Ah, they already have that method in Blockly v12 and I was referencing that branch and thought they'd be covered, but it's new in v12. There's a lot of flyout changes in that version (you can now add arbitrary types of elements to the flyout), but the API contract requires getBoundingRectangle(), so the good news is this should work well going forward. That said, I'll go ahead and patch in your changes momentarily to unblock things now; thank you!

src/navigation.ts
```diff
@@ -1452,7 +1454,7 @@ export class Navigation {
     );
     if (typeof buttonCallback === 'function') {
       buttonCallback(button);
-    } else {
+    } else if (!button.isLabel()) {
```
microbit-matt-hillsdon, line 1457: ❤️

I think this is a fix for #150

src/workspace_utilities.ts
```diff
@@ -0,0 +1,68 @@
...
+  workspace: Blockly.WorkspaceSvg,
+) {
+  const scale = workspace.getScale();
+
+  // Add some padding to the bounds so the element is scrolled comfortably
+  // into view.
+  bounds.top -= 10;
+  bounds.bottom += 10;
+  bounds.left -= 10;
+  bounds.right += 10;
```
BenHenning, line 32: Should we be making a copy of the `Rect` here? Unsure of the implications of changing the bounding rect returned from the block.

BenHenning (approved):
Thanks @gonfunko! Wondering if #150 should be marked as fixed here, too, per @microbit-matt-hillsdon's comment.

Otherwise had one comment, but approving to unblock.

src/workspace_utilities.ts
gonfunko: Yeah that's a good call, done.

Issue #65, opened 2024-10-07
Display active keyboard shortcuts to users


rachel-fenichel:
Currently the active shortcuts are displayed in the announcer region as a hard-to-parse table. This was mostly for engineering purposes.

Edward is working on a nicer overlay to show keyboard shortcuts. Here's a mock-up:
![A list of keyboard shortcuts in a modal over a Blockly workspace](https://github.com/user-attachments/assets/a4fc16ca-70e7-43f3-94af-eab8cf39e922)

This will replace the existing behaviour when the `?` key is pressed.

This modal checks which system is being used and only shows shortcuts that are appropriate to the active OS.

cpcallen:
@kmcnaught: Are you planning to use this dialog in testing, or just print out a list of shortcuts to give to participants?

kmcnaught:
Hi @cpcallen, yes we expect the dialog to be used/useable in testing, but will likely also print out a list of shortcuts to have available if appropriate to offer at some point (we don't expect to present it to everyone up front).

rachel-fenichel:
This exists and is usable. Thanks @microbit-matt-hillsdon for the cleaning and tidying. We can open targeted issues to track future work on this dialog.

PR #250, opened 2025-02-26
feat: tidy shortcut help listings

microbit-matt-hillsdon:
- Remove outdated, duplicate and work-in-progress/not-yet-useful actions.
- Tweak styling as we drop down to three columns.
- Drive order from the category listing and reorder.

I also think we should consider removing 'e' as Escape works consistently (e.g. to close colour picker and other editors) but 'e' does not. Unsure of its original motivation though so will raise separately for discussion.

I'm planning to use this for user testing but I think we should consider merging it here too. I think likely we'll try to replace this dialog with one that is specific to MakeCode in the longer term.

Related to #65


src/constants.ts
```diff
@@ -78,41 +78,37 @@ export enum MODIFIER_KEY {
  * Categories used to organised the shortcut dialog.
  * Shortcut name should match those obtained from the Blockly shortcut register.
  */
-export const SHORTCUT_CATEGORIES = {
+export const SHORTCUT_CATEGORIES: Record<
+  string,
+  // Also allow undo/redo. Document the non-keyboard-nav versions of others for better text
+  // because temporarily the name in the table is derived from these id-like names.
```
microbit-matt-hillsdon: This was a surprise! I looked for the "Toggle keyboard nav" text to improve it and was surprised it didn't exist as such. Hopefully #155 can address this.

gonfunko (approved):

src/constants.ts
cpcallen: This entire file would have gone away quite a long time ago except that that refactor has never been as important as making functional changes, but you are correct to suspect that we'll likely be changing how all of these shortcuts are defined at some point in the near future.

(In the mean time, though: may I encourage you to try to keep comment lines to <= 80 chars, please?)

microbit-matt-hillsdon: Sure, 88374cefdd8cb8b663901f1d742d74c88c22c5b8

microbit-matt-hillsdon:
Added documentation for Enter which we spotted was missing while preparing for user testing.

microbit-matt-hillsdon:
Pushed a final change to tidy order of "General" and drop from the docs the action to toggle keyboard navigation based on feedback ahead of user testing. I'll raise a separate issue shortly proposing the removal of the toggle action.

PR #252, opened 2025-02-27
fix: Put cursor on newly-created block, not its top connection

### LINKED ISSUE #221: Cursor position after insertion
After inserting a stack block, the cursor currently lands on that block's previous connection. 

After inserting a reporter block, the cursor lands on the reporter block's output connection.

Neither of these can be reached through normal keyboard navigation.

In both cases, the cursor should move to the inserted block instead.

### LINKED ISSUE #223: Extra cursor/selection after mouse block insertion
1. Load [the demo](https://google.github.io/blockly-keyboard-experimentation/)
2. With the mouse, drag a Logic > "if" block between "create canvas" and "Set background color"
    - This isn't block specific and is a general and easy to encounter issue
3. Note unexpected box / shading around the if and "Set background color" blocks

@kmcnaught I'm not sure of the impact on this on user testing. Are we expecting some mixed mouse/keyboard use?

### Actual state:

Image

### Expected state

As you'd get without the plugin, e.g. in the playground:

Image

cpcallen:
Set the cursor's `curNode` to a `BLOCK` node for the newly-creatred block, rather than its `PREVIOUS` or `OUTPUT` node.

This change revealed that there was an issue with the marker being shown for newly-inserted blocks, even when the cursor was on the block itself (and the block was therefore selected).  To fix this, override `Marker.prototype.setCurNode` and `.draw` to prevent those methods from calling `this.drawer.draw` directly; instead, route those calls via `.drawMarker` (a revised version of the previous `.updateFocusIndication` method).

This ensures that `this.drawer.draw()` is only called if `curNode` is not a `BLOCK` node, even when the block to which the marker is attached is being rendered.

Since `MARKER_MOVED` events are normally fired by `MarkerSvg.prototype.draw` calling `this.fireMarkerEvent`, `drawMarker` will arrange to call `this.drawer.fireMarkerEvent` directly when it does not call `this.drawer.draw`.

Fixes #221.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -523,39 +523,49 @@ export class LineCursor extends Marker {
       return curNode;
     }
     const newNode = new ASTNode(ASTNode.types.BLOCK, selected);
-    super.setCurNode(newNode);
-    this.updateFocusIndication(curNode, newNode);
+    this.setCurNode(newNode);
     return newNode;
   }
 
   /**
    * Set the location of the cursor and draw it.
    *
-   * Overrides drawing logic to call `setSelected` if the location is
-   * a block, or `addSelect` if it's a shadow block (since shadow
-   * blocks can't be selected).
-   *
-   * TODO(#142): The selection and fake-selection code was originally
-   * a hack added for testing on October 28 2024, because the default
-   * drawer behaviour was to draw a box around the block and all
-   * attached child blocks, which was confusing when navigating
-   * stacks.
-   *
-   * Since then we have decided that we probably _do_ in most cases
-   * want navigating to a block to select the block, but more
-   * particularly that we want navigation to move _focus_.  Replace
-   * this selection hack with non-hacky changing of focus once that's
-   * possible.
+   * Overrides normal Marker setCurNode logic to call
+   * this.drawMarker() instead of this.drawer.draw() directly.
    *
    * @param newNode The new location of the cursor.
    */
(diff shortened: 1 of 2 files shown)


microbit-robert:
Worth pointing out that this also appears to largely fix https://github.com/google/blockly-keyboard-experimentation/issues/223, though the cursor outline is partially covered by the "set background..." block below.

image

cpcallen:
Thanks @microbit-robert: I was hoping it might but hadn't actually checked yet—the hour was a bit late, even for me…

The issue with the block selection indicator being partially obscured is in core and has probably existed for a while but I have filed google/blockly#8787.  Given that it has probably been around for a while and is purely cosmetic it's probably not a high priority, but I've asked one of my colleagues to have an initial look at it anyway: he's been laying waste to our bug backlog and this one _might_ have an easy fix.

BenHenning:
I think @RoboErikG has been looking at this in depth, so assigning over for review.

Edit: I also took a glance at the changes and didn't have any specific concerns. It seems like a reasonable approach.

RoboErikG (approved):
Merging as it works and markers are a low priority for non-hacky fixes.

If you have time or you find yourself making more changes to line_cursor.ts you might consider making a local instance of the drawer and currNode variables and removing all the calls to super (eg, copy all the code from the super and just use it as an interface without calling the super methods). That would let you modify whatever you need without all the kludgey hacks.

PR #261, opened 2025-02-28
fix: Cursor position after delete

### LINKED ISSUE #254: Cursor position after delete
If I delete "random colour" in this setup:

Image

Then I get

Image

Which is a position that can't normally be navigated to.

Similar to #221 but with a different fix (unaffected by @cpcallen's #252). It would be worth reviewing the rest of `moveCursorOnBlockDelete` at the same time.

cpcallen:
Refactor to have `LineCursor` rather than `Navigation` decide where the cursor will move to when a block deletion occurs.

The cursor is moved to first of the following locations that is valid:

- The current location.
- The connection to which the deleted block is attached.
- The block connected to the next connection of the deleted block.
- The parent block of the deleted block.
- A location on the workspace beneath the deleted block.

Part of #254.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -487,6 +487,78 @@ export class LineCursor extends Marker {
   }
 
   /**
+   * Prepare for the deletion of a block by moving the cursor out of
+   * the subtree of blocks rooted at that block (if necessary).
+   *
+   * N.B.: When block is deleted, all of the blocks conneccted to that
+   * block's inputs are also deleted, but not blocks connected to its
+   * next connection.
+   *
+   * The cursor is moved to one of the following locations (in order
+   * of preference):
+   *
+   * - The connection to which the deleted block attached, if valid.
+   * - The block connected to the next connection of the deleted block.
+   * - The parent block of the deleted block.
+   * - A location on the workspace beneath the deleted block.
+   *
+   * @param deletedBlock The block that is being deleted.
+   */
+  prepareForDelete(deletedBlock: Blockly.Block) {
+    const curNode = this.getCurNode();
+
+    // See if we need to move the cursor by walking up tee from
+    // curNode to see if we get to deletedBlock.
+    const curBlock = curNode.getSourceBlock();
+    for (let block = curBlock; block !== deletedBlock; block = block.getParent()) {
+      // If we get to top block of stack (or were not on a stack at
+      // all), or discover we are next in stack below deleted block
+      // then no move needed.
+      if (!block || block.getPreviousBlock() === deletedBlock) return;
+    }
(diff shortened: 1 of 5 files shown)


cpcallen:
Please feel free to merge this if there are no substantial fixes needed.

microbit-matt-hillsdon:
In the random colour case from the original issue this now selects the draw circle block. I expected it to select the shadow block that is in the same position that the random colour block was in. This feels like the minimum selection change that can be made.

This PR is definitely an improvement though as it does not select a position the user cannot navigate to.

cpcallen:
> In the random colour case from the original issue this now selects the draw circle block. I expected it to select the shadow block that is in the same position that the random colour block was in. This feels like the minimum selection change that can be made.

Hmm.  Good point, and I agree that that behaviour is desirable.  Unfortunately I think it may entail a substantial rewrite, since AFAIK the shadow block does not actually exist before the delete occurs, so no `ASTNode` can be created for it by `preDelete`.  🤔

> This PR is definitely an improvement though as it does not select a position the user cannot navigate to.

That was certainly the main objective!

cpcallen:
I have realised that not only is there not a quick fix for the situation @microbit-matt-hillsdon raises above, but in fact he has opened a small can of worms that I had not properly considered when designing this PR.

Since further discussion will be required, I think it best to try to merge this PR in roughly its present form (pending review) ASAP and postpone further improvment to a subsequent PR.

rachel-fenichel:
I'm reviewing now.

src/line_cursor.ts
```diff
@@ -486,6 +504,82 @@ export class LineCursor extends Marker {
...
+    const curNode = this.getCurNode();
+
+    const nodes: Blockly.ASTNode[] = [curNode];
+    // The connection to which the deleted block is attached.
+    const parentConnection =
+      deletedBlock.previousConnection?.targetConnection ??
+      deletedBlock.outputConnection?.targetConnection;
+    if (parentConnection) {
+      const parentNode = Blockly.ASTNode.createConnectionNode(parentConnection);
+      if (parentNode) nodes.push(parentNode);
```
rachel-fenichel, line 539: (For team discussion, and not a blocker for this PR)

I strongly prefer to only use single-line `ifs` for returns. If there's anything at all happening in the if, such as here, I prefer to have the brackets. The style guide says they _may_ be elided; I believe in core we only elide them for returns.

rachel-fenichel (approved):
LGTM for the behaviour fix.

I can see the new can of worms; it will be interesting figuring out where this code lives in core, since it's similar to the healStack behaviour where we attempt to do something magical but intuitive after potentially large changes to the actual AST.

PR #268, opened 2025-03-04
fix: shortcut dialog ChromeOS detection, tweak Linux

microbit-matt-hillsdon:
navigator.platform is "Linux x86_64" so we need the user agent to distinguish.

I've also changed Meta -> Ctrl for Linux users too as I think it's really only
Emacs users who talk about a meta key these days and Ctrl is what will work for
current Linux environments.

Essentially we're only distinguishing ChromeOS for the dialog title.

Flagged by @kmcnaught as an issue for user testing but we'll merge this PR locally so no rush to review.

Demo: https://fix-chromeos.blockly-keyboard-experimentation.pages.dev/

I have tested on ChromeOS, Mac and Windows.
I'll review Linux this evening UK time.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -65,16 +65,6 @@ export enum LOGGING_MSG_TYPE {
 }
 
 /**
- * Platform specific modifier key used in shortcuts.
- */
-export enum MODIFIER_KEY {
-  Windows = 'Ctrl',
-  ChromeOS = 'Ctrl',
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):

microbit-matt-hillsdon:
Belatedly confirming that Linux is fine too (Fedora 41/GNOME/Chrome).

PR #270, opened 2025-03-04
feat: Implement proposed changes to `preDelete` / `postDelete`

### LINKED ISSUE #254: Cursor position after delete

cpcallen:
Move most of the responsibility for choosing where to move the cursor, upon deletion of a block, from `preDelete` to `postDelete`:

- `preDelete` now merely saves some information about the connection to which the deleted block is attached and the cursor's current location on the workspace.
- `postDelete` does all the work of deciding where to move the cursor to.

The list of places to try to move the cursor is revised to:

- The current location (if still valid)
- Any block connected in the same place the deleted block was.
- The connection to which the deleted block was attached.
- The parent block of the deleted block.
- A location on the workspace beneath the deleted block.

The upshot of this is that, when a deleted block is replaced by a shadow block, the cursor will normally be placed on that shadow block.

Fixes #254.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -46,8 +46,11 @@ export class LineCursor extends Marker {
   /** Old Cursor instance, saved during installation. */
   private oldCursor: Blockly.Cursor | null = null;
 
-  /** Locations to try moving the cursor to after a deletion. */
-  private potentialNodes: Blockly.ASTNode[] | null = null;
+  /** Info to be used to moving the cursor after a deletion. */
+  private deleteInfo: {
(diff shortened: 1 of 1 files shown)


cpcallen:
Leaving this as draft until there is slightly clearer consensus on the desired behaviour.

Issue #146, opened 2025-01-14
Keyboard navigation not functional after reloading blocks


microbit-matt-hillsdon:
If you [modify the test app like this](https://github.com/google/blockly-keyboard-experimentation/compare/main...microbit-matt-hillsdon:blockly-keyboard-experimentation:client-side-reload?expand=1) to handle the scenario selector client side, then you hit a problem.

1. Load the modified page ([deployed here](https://client-side-reload.blockly-keyboard-experimentation.pages.dev/))
2. Use keyboard navigation to move the cursor onto a block
3. Use the scenario selector to load blocks
4. Try to use keyboard navigation again and note it doesn't highlight anything

This corresponds to a real scenario in MakeCode. When MakeCode switches from JavaScript to Blocks mode it reloads (via XML) blocks into Blockly. They may or may not have changed at this point due to edits in JavaScript mode. From that point on the keyboard navigation is broken.

You can recover navigation by navigating to the toolbox with "T" and adding a new block.

Is this something the plugin can be modified to cope with?

rachel-fenichel:
Debugging notes:
The failure here is that the keyboard navigation code is using the same cursor it had before, and the cursor is pointing to a no-longer-extant-node. In my test case, the stack trace was
```
Cannot read properties of null (reading 'fieldRow')
TypeError: Cannot read properties of null (reading 'fieldRow')
    at ASTNode$$module$build$src$core$keyboard_nav$ast_node.findNextForField (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1026:163)
    at ASTNode$$module$build$src$core$keyboard_nav$ast_node.next (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1035:83)
    at LineCursor.getNextNode (webpack-internal:///./src/line_cursor.ts:297:43)
    at LineCursor.in (webpack-internal:///./src/line_cursor.ts:58:30)
    at Object.callback (webpack-internal:///./src/navigation_controller.ts:177:104)
    at ShortcutRegistry$$module$build$src$core$shortcut_registry.onKeyDown (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1280:193)
    at onKeyDown$$module$build$src$core$inject (webpack-internal:///./node_modules/blockly/blockly_compressed.js:423:269)
    at HTMLDocument.f (webpack-internal:///./node_modules/blockly/blockly_compressed.js:86:159)
```

[...]

PR #271, opened 2025-03-04
fix: don't keep cursor on disposed blocks

microbit-matt-hillsdon:
Perhaps this should be cleared earlier, but this scenario happens in MakeCode when switching back to Blocks mode from JavaScript.

One approach to https://github.com/google/blockly-keyboard-experimentation/issues/146

This feels like a last-ditch way to notice an invalid selection but I think it's common (as in MakeCode and the test app) to disable events during loading so the listeners can't do this for us.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -511,10 +511,10 @@ export class Navigation {
       return;
     }
 
-    // This can happen if blocks are reloaded.
-    const isDisposed = cursor.getCurNode()?.getSourceBlock()?.disposed;
-    if (cursor.getCurNode() && !isDisposed && keepPosition) {
-      // Retain the cursor's previous position since it's set.
+    const disposed = cursor.getCurNode()?.getSourceBlock()?.disposed;
+    if (cursor.getCurNode() && !disposed && keepPosition) {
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -511,7 +511,9 @@ export class Navigation {
       return;
     }
 
-    if (cursor.getCurNode() && keepPosition) {
+    // This can happen if blocks are reloaded.
+    const isDisposed = cursor.getCurNode()?.getSourceBlock()?.disposed;
+    if (cursor.getCurNode() && !isDisposed && keepPosition) {
       // Retain the cursor's previous position since it's set.
```
BenHenning:
```suggestion
      // Retain the cursor's previous position since it's set, but only if not
      // disposed (which can happen when blocks are reloaded).
```

Suggest making this change and removing the other comment since this if-check is doing more than just handling the reloaded case.

microbit-matt-hillsdon: Thanks - 8329211f8f96ca8a1426fe276e3c6bb5752b48d9

Also removed the inconsistent "is" prefix from the local.

rachel-fenichel (approved):

PR #273, opened 2025-03-04
fix: remove marked node field

### LINKED ISSUE #263: Paste over shadow block only works on the second attempt via menu
1. Load or reload [the demo](https://google.github.io/blockly-keyboard-experimentation/)
2. 'T' to open toolbox
3. p5 blocks
4. Copy random colour via e.g. Cmd+C
5. Navigate to the colour block
6. Paste via the menu
    - It works fine via the shortcut
7. Note it pasted on the workspace
8. Try again, note it works

https://github.com/user-attachments/assets/7963cb29-3af0-4015-a458-8ef758f12842

microbit-matt-hillsdon:
Instead hide/show cursor/passive focus indicator on the relevant focus changes.

Fixes #263

I totally might be missing a reason why we can't do this. But if we can it seems like a clear win.

Demo: https://simplify-marked-node.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -74,12 +74,6 @@ export class Navigation {
   protected passiveFocusIndicator: PassiveFocus = new PassiveFocus();
 
   /**
-   * The node that has passive focus when the cursor has moved to the flyout
-   * or toolbox; null if the cursor is moving around the main workspace.
-   */
-  protected markedNode: Blockly.ASTNode | null = null;
-
-  /**
    * Constructor for keyboard navigation.
    */
   constructor() {
@@ -418,9 +412,12 @@ export class Navigation {
     if (!toolbox) {
       return;
     }
+    const cursor = workspace.getCursor();
+    if (cursor) {
+      this.passiveFocusIndicator.show(cursor.getCurNode());
+      cursor.hide();
+    }
 
-    this.markAtCursor(workspace);
-    workspace.getCursor()?.hide();
     this.setState(workspace, Constants.STATE.TOOLBOX);
     this.resetFlyout(workspace, false /* shouldHide */);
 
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
Note to self: if we merge this then the same logic needs to be run in the external toolbox scenario.

src/navigation.ts
```diff
@@ -418,9 +412,12 @@ export class Navigation {
     if (!toolbox) {
       return;
     }
+    const cursor = workspace.getCursor();
+    if (cursor) {
+      this.passiveFocusIndicator.show(cursor.getCurNode());
+      cursor.hide();
```
cpcallen, line 418: It should be noted that (contrary to expectations) `Marker.prototype.hide` does not cause the cursor to be persistently hidden, but only hides the marker SVG, and only until the next time a render occurs.

Marker SVG is used for denoting position of cursor when it is on a connection or on the workspace (or on a whole-stack node, though these are only currently accessible via the context-out shortcut which will probably go away very soon).

So AFAICT this won't hide the cursor if it's on a block, and if it isn't on a block then it will be hidden but may reappear unexpectedly if something causes a render to occur.

Probably not a big deal, given that the existing code does exactly the same thing, though I do wonder if it might be better to hide the cursor before showing the passive focus indicator (here and in `focusFlyout`).

src/navigation.ts
```diff
@@ -443,8 +440,11 @@ export class Navigation {
    * @param workspace The workspace the flyout is on.
    */
   focusFlyout(workspace: Blockly.WorkspaceSvg) {
-    workspace.getCursor()?.hide();
-    this.markAtCursor(workspace);
+    const cursor = workspace.getCursor();
+    if (cursor) {
+      this.passiveFocusIndicator.show(cursor.getCurNode());
+      cursor.hide();
```
cpcallen, line 446: Does it still work if you call `cursor.hide()` before `passiveFocusIndicator.show`?  It would seem more tidy to do it in that order.

cpcallen (approved):
Code looks fine (though note comments below).  Would like to verify that this works in practice before merging, though, as @rachel-fenichel had some concerns.

cpcallen:
OK, I played with this for a while and couldn't find any situations in which the cursor was not in the expected place.

LGTM.

microbit-matt-hillsdon:
Closing in favour of discussion around #326

Issue #132, opened 2024-11-15
Add context menu items to facilitate editing


cpcallen:
Add the following items to the context menus for blocks, with the the given keyboard shortcuts:

* [x] **Edit contents** (`→`): Moves the cursor (focus) to the first focusable field or input.
  * This option exists mainly to facilitate discoverability of the right arrow as a means to navigate to the contents of a block.  It should probably be the first item in the context menu.
* [ ] **Move block** (`m`): "Picks up" the block to move it using the dragging workflow (see #106).
* [x] #173
  * See #102 for details of how this should work.
* [x] **Delete block** (`del`+`backspace`): delete the block, as current delete shortcut does.
* [x] [**Cut** (`⌘X` etc)](https://github.com/google/blockly-keyboard-experimentation/issues/217)
* [x] [**Copy** (`⌘X` etc)](https://github.com/google/blockly-keyboard-experimentation/issues/218)
* [x] [**Paste** (`⌘X` etc): ](https://github.com/google/blockly-keyboard-experimentation/issues/219)Paste (see #128).

Add the following item to the context menu for unconnected next and input connections, as well as when the cursor is visiting the workspace:

* [ ] **Place block** (same as replace block, above): place a new block at this location using the marker workflow.

The names given above are not necessarily intended to be the final English names for the given items, merely descriptive.

cpcallen:
@rachel-fenichel: as I [noted in the discussion with Kirsty](https://github.com/google/blockly-keyboard-experimentation/issues/130#issuecomment-2605515323), I think "Insert" should be the primary shortcut for adding blocks, and I'm not sure that if we have insert we need replace.

kmcnaught:
Is this a question about naming, shortcut key choice, or functionality? Or all three?

If you are at a prev/next connection then "insert" would add a new block in the current location. 

If you were at an in/out connection (_not_ a block) then an action called "insert" might replace the currently-connected block(s) with the newly selected one. Are there any geras-style platforms where it might be appropriate/meaningful to insert a block between in/out as an extra block, without replacing anything? I don't think this would ever be the case in MakeCode.

If you were at a [single cursor location that collapses a block and its out connection] then I was imagining an action called "replace" would replace the currently-connected block(s) with the newly selected one. 

[...]

PR #274, opened 2025-03-04
feat: Edit option in block context menu

cpcallen:
Introduce an "Edit block contents (→)" block context menu option.  This is shown any time the cursor is on a block and not in the rightmost position 'on the current line'; consequently the description may sometimes be slightly misleading, because it might not be the contents of the _current_ block that's being navigated to, but rather that of a sibling or parent block.  The menu option weights are adjusted to put edit between insert and delete.

Also:
- Rename the main directional shortcuts to match the names of the corresponding keys.  (The corresponding methods on `LineCursor`are not renamed, as that seems like a riskier change.)
- Sort inputs in `navigation_controller.ts`.

Part of #132.


cpcallen:
Marking as draft, as somehow I lost a big chunk of code in last commit!

cpcallen:
Turns out this PR is fine; it was my local branch that I had somehow got in to some werid state that was not building.

src/constants.ts
```diff
@@ -23,10 +23,10 @@ export enum STATE {
  * Default keyboard navigation shortcut names.
  */
 export enum SHORTCUT_NAMES {
-  PREVIOUS = 'previous',
-  NEXT = 'next',
-  IN = 'in',
-  OUT = 'out',
+  UP = 'up',
+  DOWN = 'down',
+  RIGHT = 'righ',
```
microbit-matt-hillsdon:
```suggestion
  RIGHT = 'right',
```

src/constants.ts
```diff
@@ -23,10 +23,10 @@ export enum STATE {
  * Default keyboard navigation shortcut names.
  */
 export enum SHORTCUT_NAMES {
-  PREVIOUS = 'previous',
-  NEXT = 'next',
-  IN = 'in',
-  OUT = 'out',
+  UP = 'up',
```
rachel-fenichel: You're gonna hate me for this, but--this breaks for RTL and horizontal flyouts.

cpcallen: This change breaks nothing, and is arguably not even in incorrect, since this really only pertains to the cursor keys, and AFAIK neither the glyphs moulded into the keycaps nor the keycodes change in RTL.

But yes: we do need to figure out how to handle RTL and horizontal flyouts correctly—but I don't think either of those are direclty relevant here.

cpcallen: I thought about this a bit more, and wrote up [a lengthy comment about it](https://github.com/google/blockly/issues/8788#issuecomment-2707693217) on google/blockly#8788.

I am currently inclined to think that it is probably best that we continue to have a per-workspace (in future per-FocusTree) cursor (in future, cursor-equivalent), make those cursors responsible for deciding how to map physical directions to logical directions, and have four simple actions (up/down/left/right) that delegate to the currently-active cursor.

The code you comment on moves us in that direction; nevertheless it is not actually relevant to this PR so I have removed it.

cpcallen:
@BenHenning: this remains ready for review.

src/actions/edit.ts
```diff
@@ -0,0 +1,81 @@
...
+ * discoverability.
+ *
+ * N.B.: This item is shown any time the cursor is on a block and not
+ * in the rightmost position 'on the current line'; that means that
+ * sometimes the label ("Edit block contents") is possibly misleading,
+ * because it might not be the contents of the _current_ block that's
+ * being edited, but rather that of a sibling or parent block.
+ *
+ * This action registers itself only as a context menu item, as there
+ * is already a corresponding "right" shortcut item.
```
BenHenning, line 39: Does it perhaps make sense to consolidate this with the shortcut for workspace contexts? It seems to be doing roughly the same thing, and it feels a bit odd to have an inconsistency here as compared to other action classes (where those manage both context menu and shortcut triggers).

It wouldn't seem odd, I think, for there to be a workspace-only shortcut handled here either since the context menu callback and precondition are already assuming workspace only. I suppose the only aspects I'm unsure of is having multiple actions bound for the same key (RIGHT) or whether splitting that into multiple locations is wise.

src/line_cursor.ts
```diff
@@ -175,6 +175,27 @@ export class LineCursor extends Marker {
...
+  /**
+   * Returns true iff the node to which we would navigate if in() were
+   * called, which will be a validInLineNode, is also a validLineNode
+   * - in effect, if the LineCursor is at the end of the 'current
+   * line' of the program.
+   *
+   * This is used to determine whether to display an "Edit Block
+   * contents (→︎)" entry in the block context menu, to help users
+   * discover that the right arrow key can be used to navigate to
+   * block contents.
```
BenHenning: Nit, but this might be an overly specific documentation item as it's concerning itself with how this function may be used (which also may not stay true forever if context menu logic changes or we find new uses for this functionality).

src/line_cursor.ts
```diff
@@ -175,6 +175,27 @@ export class LineCursor extends Marker {
...
+   */
+  public atEndOfLine(): boolean {
+    const curNode = this.getCurNode();
+    if (!curNode) return false;
+    const rightNode = this.getNextNode(
+      curNode,
+      this.validInLineNode.bind(this),
+    );
+    return this.validLineNode(rightNode);
```
BenHenning, line 191: Is this needed? `getNextNode` seems guaranteed to either return `null` or a 'valid' node (i.e. one for which `validInLineNode()` already returns `true`).

BenHenning (commented):
Thanks @cpcallen! No major concerns, just some clarification questions. PTAL.

src/actions/edit.ts
cpcallen: I think this code could be merged in to `actions/ws_movement.ts`, now that has been separated out in #281.

The precondition for showing the "edit" option is a bit different from the precondition for allowing the right arrow to work, though: the right arrow key is practically always enabled, wheras the edit option is only shown in some places.

(And the edit option will need to be updated once the cursor properly supports RTL, since in that case it will do the same thing as the left arrow key, rather than the right arrow key.)

Unfortunately I won't have time to do such a refactor before I leave today, and I won't be back for a couple of weeks.

src/line_cursor.ts
cpcallen: Hmm, perhaps this should be moved to the class.

src/line_cursor.ts
cpcallen: Yes, because `getNextNode` returns nodes that are `validInLineNode`s, but only some of those are `validLineNode`s.  In particular, a field is a valid inline node but not a valid line node.

(This is actually the key feature of the whole PR!)

BenHenning: It turns out that validLineNode and validInLineNode read the exact same to me without much more careful watching. That was the source of my confusion, actually. Thanks for clarifying.

src/actions/edit.ts
BenHenning: I think it's fine to merge this as-is for the benefits of the clean-up. Will approve, thanks!

BenHenning (approved):
PR LGTM for the clean-up. Any additional cleaning can be done later.

BenHenning:
@rachel-fenichel since Chris is out how do you want to handle merging this?

rachel-fenichel:
I fixed some merge conflicts/bad imports and am merging now.

PR #281, opened 2025-03-06
chore: move workspace movement into a separate file

rachel-fenichel:
We don't know exactly what free movement on the workspace will look like yet, but we know we'll want some form of it.

This PR consolidates the existing code into a single action file, including implementation code that used to be in `navigation.ts`. As a side benefit, you can see from the result that there's no dependency on the `Navigation` class at this point: all of workspace navigation can be done by interacting directly with the workspace and its cursor.

Tested by modifying `Navigation.setCursorOnWorkspaceFocus` to force the cursor to go to the workspace, then using the relevant keyboard shortcuts.

This is pure cleanup with no behaviour changes.

### DIFF
--- a/src/actions/ws_movement.ts
+++ b/src/actions/ws_movement.ts
@@ -1,0 +1,133 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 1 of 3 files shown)


BenHenning (approved):
Another nice cleanup, thanks!

PR #283, opened 2025-03-06
chore: move paste implementation into clipboard.ts

rachel-fenichel:
Sibling of https://github.com/google/blockly-keyboard-experimentation/pull/281

Moves the actual implementation of paste out of `navigation.ts` and into `clipboard.ts` along with the rest of the callback. It's only used in one place and there was not a good reason to have the callback split into two pieces like this.

The second commit closes the event group in the case where paste succeeds; this was previously omitted, which was a bug.

Tested by copying and pasting in various places and checking that behaviour is consistent with main.

### DIFF
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -374,6 +374,7 @@ export class Clipboard {
         );
       }
       this.navigation.removeMark(pasteWorkspace);
+      Events.setGroup(false);
       return true;
     }
     Events.setGroup(false);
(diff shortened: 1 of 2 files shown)


cpcallen (approved):

cpcallen (commented):
> The second commit closes the event group in the case where paste succeeds; this was previously omitted, which was a bug.

There is only one commit in this PR!

rachel-fenichel:
Oops! Added the second commit.

PR #284, opened 2025-03-06
chore: move disconnect action to its own file

rachel-fenichel:
Sibling of https://github.com/google/blockly-keyboard-experimentation/pull/283

Moves the disconnect shortcut definition and callback into a single file, `src/actions/disconnect.ts`.

Tested by running and disconnecting various things.

This is purely code cleanup and does not change behaviour. There is an existing bug with the disconnect behaviour (https://github.com/google/blockly-keyboard-experimentation/issues/311) that this does not address.

### DIFF
--- a/src/actions/disconnect.ts
+++ b/src/actions/disconnect.ts
@@ -1,0 +1,133 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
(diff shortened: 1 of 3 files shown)


BenHenning (approved):
Nice cleanup, thanks!

PR #285, opened 2025-03-06
chore: move action menu to its own file

rachel-fenichel:
Sibling of https://github.com/google/blockly-keyboard-experimentation/pull/284

Moves the action menu and all of its code from `navigation.ts` and `navigation_controller.ts` to a new file, `src/actions/action_menu.ts`.

This action is different from the others in that it needs a `canCurrentlyNavigate` function rather than a `canCurrentlyEdit` function for precondition checks.

This PR just moves code around. It doesn't change any behaviour. Testing by running it.


src/actions/action_menu.ts
```diff
@@ -0,0 +1,299 @@
...
+import * as Constants from '../constants';
+import type {BlockSvg, WorkspaceSvg} from 'blockly';
+import {Navigation} from '../navigation';
+
+const KeyCodes = BlocklyUtils.KeyCodes;
+const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
+  ShortcutRegistry.registry,
+);
+
+interface Scope {
```
cpcallen, line 26: Perhpas this should be exported so it can be imported by the menu items that now support connection contexts (which IIRC currenlty have their own copy of this type).

cpcallen (approved):

rachel-fenichel:
Superseded by #314

PR #291, opened 2025-03-06
feat: support external toolbox (e.g. MakeCode)

microbit-matt-hillsdon:
MakeCode's toolbox is a separate UI component (the toolbox Blockly sees is empty).  Add just enough glue to get keyboard navigation working in its flyout and be able to close the flyout when we leave the toolbox.

This is analogous to the monkey patching that was done for the original keyboard navigation plugin (which isn't viable anymore due to better encapsulation).

One to discuss next week, but it would be really helpful to merge something like this even if we aspire to do it differently in the long term.

I've tried to clarify this a little based on a past chat discussion and it's got simpler as the plugin has evolved.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,10 +7,16 @@
 import * as Blockly from 'blockly/core';
 import {NavigationController} from './navigation_controller';
 import {CursorOptions, LineCursor} from './line_cursor';
+import * as Constants from './constants';
+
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
Closing in favour of the approach on #358 which leaves all the messy bits on the MakeCode end.

PR #303, opened 2025-03-07
fix: change cursor rendering for empty value inputs in zelos

### LINKED ISSUE #288: Replace tiny circle cursor with outline for a value input with no shadow block (e.g. if)
In Zelos at least, we think the cursor should outline the space where the block would go.

So instead of this:

Image

...do this (but without the hand-drawn vibe):

Image

rachel-fenichel:
This fixes #288 but only for zelos. Falls back on old behaviour for zelos.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -606,25 +606,87 @@ export class LineCursor extends Marker {
       }
     }
 
-    // If curNode node is not block, just use the drawer.
-    if (curNode?.getType() !== ASTNode.types.BLOCK) {
+    if (this.isValueInputConnection(oldNode)) {
+      this.hideAtInput(oldNode);
+    }
+
+    const curNodeType = curNode?.getType();
+    const isValueInputConnection = this.isValueInputConnection(curNode);
+
+    // If drawing can't be handled locally, just use the drawer.
+    if (curNodeType !== ASTNode.types.BLOCK && !isValueInputConnection) {
       this.getDrawer()?.draw(oldNode, curNode);
       return;
     }
 
-    // curNode is a block.  Hide any visible marker SVG and instead
-    // select the block or make it look selected.
+    // Hide any visible marker SVG and instead do some manual rendering.
     super.hide(); // Calls this.drawer?.hide().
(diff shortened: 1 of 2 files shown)


src/line_cursor.ts
```diff
@@ -702,27 +702,89 @@ export class LineCursor extends Marker {
...
+   * Check whether the node represents a value input connection.
+   *
+   * @param node The node to check
+   * @returns True if the node represents a value input connection.
+   */
+  private isValueInputConnection(node: ASTNode) {
+    if (!node) return false;
+    if (node.getType() != ASTNode.types.INPUT) {
+      return false;
+    }
```
cpcallen: Even by your rules this can be simplified:  😉
```suggestion
    if (node?.getType() !== ASTNode.types.INPUT) return false;
```
(Though it may not be immediately obvious that `node?.getType()` evaluates to `undefined` if `node` is null or undefined, and `undefined !== INPUT` evaluates to true.)

src/line_cursor.ts
```diff
@@ -702,27 +702,89 @@ export class LineCursor extends Marker {
...
+    const connection = node.getLocation() as Blockly.Connection;
+    return connection.type == Blockly.ConnectionType.INPUT_VALUE;
```
cpcallen:
```suggestion
    return connection.type === Blockly.ConnectionType.INPUT_VALUE;
```

src/line_cursor.ts
```diff
@@ -702,27 +702,89 @@ export class LineCursor extends Marker {
...
-    if (!block.isShadow()) {
-      Blockly.common.setSelected(block);
-    } else {
-      block.addSelect();
+
+    if (this.isValueInputConnection(curNode)) {
+      this.showAtInput(curNode);
+    }
+
+    if (curNodeType == ASTNode.types.BLOCK) {
```
cpcallen:
```suggestion
    } else if (curNodeType === ASTNode.types.BLOCK) {
```

rachel-fenichel:
@gonfunko PTAL

gonfunko (approved):

PR #304, opened 2025-03-10
refactor: Move arrow key shortcuts into their own class.

### LINKED ISSUE #296: Move previous/next/in/out shortcuts to a new file
Part of a push to organize the action code and identify the APIs that we actually need in or from core.

Proposed changes:
- Create a new file, called `ArrowNavigation` for the purposes of this issue, in the `actions` directory.
- Move the shortcuts for `previous`, `next`, `in`, and `out` into `ArrowNavigation`.
- Move `NavigationController.fieldShortcutHandler` into `ArrowNavigation`.

Alternatives considered:
- Move all of the logic from shortcuts into cursor implementations
- Move all of the logic from shortcuts into workspace, toolbox, and flyout implementations
  - This is already sort of true for the toolbox, because the arrow key handlers just hand off to the toolbox's `onShortcut` function if defined.

gonfunko:
This PR fixes #296 by moving the left/right/up/down arrow key actions out of navigation_controller.ts and into their own class.


rachel-fenichel (commented):
Note: hold off on merging until Ben's current focus changes land.

cpcallen (changes requested):
Per [my remarks on #274](https://github.com/google/blockly-keyboard-experimentation/pull/274#discussion_r1985843752)—see also [this lengthy discussion about cursor actions](https://github.com/google/blockly/issues/8788#issuecomment-2707693217)—I think the actions should be named after the cursor keys they are bound to.

(And we should in due course refactor the cursors so that their APIs are also directional, with each cursor being responsible for mapping from arrow direction to AST directions depending on LTR, flyout layout, etc.)

gonfunko:
To be clear, you want separate `LeftAction`, `RightAction`, `UpAction` and `DownAction` classes/files?

cpcallen (commented):
> To be clear, you want separate `LeftAction`, `RightAction`, `UpAction` and `DownAction` classes/files?

No, not at all.  I just think we should rename the actions (id, name, description, etc.) to up/down/left/right, rather than prev/next/out/in.

(I think we should also rename the corresponding methods on the cursors, but that can be done as part of a subsequent PR that actually makes them RTL / horizontal flyout aware.)

gonfunko:
Got it, thanks for clarifying! Updated accordingly (and also removed other deleted actions from `SHORTCUT_NAMES`).

rachel-fenichel (approved):

PR #306, opened 2025-03-10
refactor: Move enter action into its own class.

### LINKED ISSUE #297: Move enter action and helpers to a new file
Create an action for the `enter` action, which is a keyboard shortcut that is triggered by either enter or space.

Several associated helpers are only used by this action and can also move into the new file:
- `navigation.handleEnterForWS`
- `navigation.insertFromFlyout`
- `navigation.triggerButtonCallback`
- `navigation.tryShowFullBlockFieldEditor`

gonfunko:
This PR fixes #297 by moving the Enter action into its own class, and moving several helper functions used only by that action into its class and out of navigation.ts. This also incorporates changes from #273 in `insertFromFlyout()`, which was moved into the enter action class.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -1,0 +1,231 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {ASTNode, ShortcutRegistry, utils as BlocklyUtils} from 'blockly/core';
+
+import type {
+  Block,
+  BlockSvg,
+  Field,
+  FlyoutButton,
+  WorkspaceSvg,
+} from 'blockly/core';
+
+import * as Constants from '../constants';
+import type {Navigation} from '../navigation';
+
+const KeyCodes = BlocklyUtils.KeyCodes;
+
+/**
+ * Class for registering a shortcut for the enter action.
+ */
+export class EnterAction {
+  constructor(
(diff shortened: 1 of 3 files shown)


cpcallen (commented):
I am about to add another meaning for the enter key: when moving a block, pressing enter will place the block at its current location.  I was going to just wire it in to `handleEnterForWS` but as this PR is now in-flight it occured to me that a better approach is to do what we should eventually do anyway, and make it a completely separate action.  Similarly, I think the several unrelated activities here should be split in to separate actions:

1. One which, when on a flyout button, presses the button.
0. One which, when on a flyout stack, creates new block(s) on the workspace.
0. One which, when on a field, opens the field editor.
0. One which, when on a connection or the workspace, runs the insert action (this should be in `insert.ts`)
0. One which, when none of the above apply, shows the help message about the context menu.
   * This needs to be registered first, since shortcuts are tried in the reverse of the order they are registered in.

It might be reasonable for 1 and 2 to be kept as part of a single action, but the other three are unrelated and ought to be separated out.  (No objection to doing this as part of  a separate follow-up PR, or even just filing a bug to do it later.)

rachel-fenichel:
Let's still merge this PR, and land your new meaning of Enter, and then assess. I am leery of splitting this into multiple actions because it spreads out the logic for deciding which action to do/which takes prioritiy.

rachel-fenichel (approved):

Issue #129, opened 2024-11-15
Change line cursor behaviour to be more intuitive


cpcallen:
In discussions following the recent UX testing, it was decided to make the following changes to the behaviour of the line cursor:

* [x] The cursor should not visit stack nodes.
* [x] The cursor should skip value connections (input/output) if there is a connected block.
* [x] The cursor should be able to visit all fields, including whole-block fields on simple reporter blocks.
* [x] When navigating between top-level stacks (or rather the constituent blocks thereof, as we will no longer visit the stack nodes), it should be possible to visit all top-level stack, and they should be visited in the same sorted order that is used for code generation.

The following change was discussed but the result of discussions seems to be in dispute (@cpcallen believes it was generally agreed, but @rachel-fenichel thinks it was not):

* [x] The cursor should skip stack connections (next/previous) if there is a connected block.

Additionally, I (@cpcallen) propose that the workspace cursor behaviour be modified to bring it in line with proposed changes to the block movement mechanism in #106, principally:

* [ ] The cursor should not visit the workspace (i.e. arbitrary xy coordinates), unless a modifier key (probably shift) was held down in conjunction with one of the arrow keys, in which case it will move to the nearest workspace gridpoint location in the specified direction (within a ±45° arc).
* [ ] When the cursor is on the workspace, and an arrow key is pressed, the cursor should move to the nearest block in the given direction (within a ±45° arc).

cpcallen:
Revised above description to allow the cursor to visit connections that do not have a connected block.  That's kind of necessary if we want to make it reasonably easy to place new blocks using the marker workflow (see #102).

kmcnaught:
>  The cursor should not visit the workspace (i.e. arbitrary xy coordinates), unless a modifier key (probably shift) was held down in conjunction with one of the arrow keys, in which case it will move to the nearest workspace gridpoint location in the specified direction (within a ±45° arc).

Agree with the principle, but not keen on Shift - we are hoping to have multiline selection (#139) available, which I'd like to reserve Shift for please :)

>  When the cursor is on the workspace, and an arrow key is pressed, the cursor should move to the nearest block in the given direction (within a ±45° arc).

Is this saying that when you are on the workspace (arbitrary xy) then:
- arrow keys jump you immediately back to a block, and
- {modifier} + arrow keys move you in (x,y) steps?

If so, I think it sound reasonable, but it wasn't immediately clear when I first read it.

[...]

PR #309, opened 2025-03-10
fix: Remove complex navigation actions.

### LINKED ISSUE #286: Remove unused actions for complex movement
- [ ] Next sibling
- [ ] Previous sibling
- [ ] Jump to root
- [ ] Context out
- [ ] Context in

These should either be deleted wholesale or extracted from `navigation_controller.ts` and moved into a separate action file until we can decide what to do with them. I vote for deleting, since we can always recreate.

Out current navigation system does not rely on these complex actions, although we may want to add them back if we see a need for them in user testing. These are also less useful than originally thought because we intend to give users a search function to jump long distances within their programs.

gonfunko:
This PR fixes #286 by removing the next/previous sibling, jump to root and context in/out actions, along with supporting functions used only by those actions.


cpcallen:
Can we hold off on submitting this until we have another way to move the cursor to the workspace?

rachel-fenichel:
What other way are you suggesting?
I'd prefer to submit this and add an explicit action to move the cursor to the workspace, probably in the same file as the rest of workspace movement, rather than keep all of these around just in case.

BenHenning (approved):
Thanks @gonfunko!

cpcallen:
> What other way are you suggesting? I'd prefer to submit this and add an explicit action to move the cursor to the workspace, probably in the same file as the rest of workspace movement, rather than keep all of these around just in case.

Well, ultimately unconstrained cursor movement, as described in #129, but for now any other way: we have left ourselves in a position where the only way to test the worksapce cursor,  workspace context menu, and any new actions that appear therein, is to place a new top-level block and then delete it.

Issue #106, opened 2024-11-08
Make it possible to move blocks on the workspace


cpcallen:
It should be possible to choose where to place newly-created blocks on the workspace, and to move existing blocks to new locations.

There should be a new mode that shall be in effect while a block is being moved (i.e., while it is 'picked up', as described below).

There should be two types of movement, which the user can easily and seamlessly (but not unintentionally) switch between:

1. Constrained movement: the block can be moved only between valid positions (for that block) within existing stacks.
2. Unconstrained movement: the block can be moved to any xy position on the workspace.

Specifically:

* [x] **Starting a move:** There should be a shortcut with corresponding block context menu item (at least when the context menu is opened via keyboard shortcut) to move the currently selected block.  When this option is selected, the block(s) will be 'picked up', and keyboard navigation will switch into 'move mode'.
  * [x] When a block that is in a stack is 'picked up', it should be removed the stack, replaced by an insertion marker (or equivalent affordance, e.g. [Rudolph](https://raw.githubusercontent.com/microsoft/pxt/refs/heads/master/docs/static/blog/microbit/2020-beta/rudolph.gif)), and moved slightly aside so as to not entirely occlude the insertion marker etc..
  * [x] When a new block is created by use of the keyboard (e.g. by navigating to a block in the flyout and pressing enter), it should initially be created in 'picked up' state:
    * If the workspace cursor was at a valid connection point for the new block, it should be as if the new block had just been picked up from that location (e.g. an insertion marker should be shown) except that if the move is aborted the newly created block should either be deleted.
    * Otherwise it should positioned nearby to the location of the workspace cursor, as if it had been moved there via unconstrained move.
  * [ ] When a block is 'picked up', it should be drawn with a suitable affordance to indicate that it it is being moved—e.g. a drop shadow.
* [x] **Constrained movement:** Absent any modifier keys, pressing the cursor keys should move the block to the next valid connection in the given direction:
  * If the moved block had not been denoted (by shadow etc.) as being connected to a valid connection point, then it should be moved to the next nearest valid connection point in the general direction of the cursor key if there is one, or to the nearest valid connection point in _any_ direction, if there are any.  (See below re: if there are no valid connection points.)
  * If the moved block had been denoted as connected, then it should be moved to the next valid connection point in navigation order, subject to the proviso that it be possible to reach _every_ valid attachment point by _some_ sequence of arrow key presses.
[...]

PR #317, opened 2025-03-13
feat: Introduce `Mover` actions + experimental `DragMover` implementation

cpcallen:
Introduces a `Mover` class which:

* Creates actions ~~(for now only keyboard shortcuts)~~ to start a move, move in cursor-key directions, and end or abort a move.
* Implements a mode for when a block move is taking place.  This is done using only `Mover`-internal state.  When in moving mode, the arrow keys are hooked to trigger block moving actions rather than their normal cursor moving actions.
* When a move is initiated, saves information about the original location of the block, so that the block can be restored to that location if the move is aborted.
* Includes a framework for—but does not yet implement—constrained and unconstrained movement.

Additionally, introduces a `DragMover` subclass that provides an experimental implementation of unconstrained block movement using `Dragger`.

**To try this out, move the cursor to a movable block, press `m`**, then use `alt`+arrows to move the block.  Press `enter` to finish the move, or `esc` to abort it.

Moving using the dragger sort of works, with a few known issues:

* Blocks being moved lurch around somewhat wildly, especially when they are near compatible connectors.
* Sometimes blocks not being moved also move around.
* ~~Moving top-level blocks works reasonably well, but attempting to move a non-top-level block will result in it being deleted unless the move is aborted.~~ FIXED!

The purpose of separating this experimental implementation into a separate subclass is that it is anticipated that it may well be thrown away, both because of the above noted issues and because this approach may be fundamentally incompatible with constrained movement.  Specifically: constrained movement should first concern itself with where the moving block(s) might be connected, and only then decide where to position them on screen (e.g. so that they are very near but not obscuring an insertion marker)—whereas dragging begins with the position of the block and from that decides where the block might connect.

Part of #106.

### DIFF
--- a/src/actions/drag_mover.ts
+++ b/src/actions/drag_mover.ts
@@ -129,7 +129,7 @@ export class DragMover extends Mover {
     // Not yet implemented.  Absorb keystroke to avoid moving cursor.
     alert(`Constrained movement not implemented.
 
-Use alt+arrow (option+arrow on macOS) for unconstrained move.
+Use ctrl+arrow or alt+arrow (option+arrow on macOS) for unconstrained move.
 Use enter to complete the move, or escape to abort.`);
     return true;
   }
--- a/src/actions/drag_mover.ts
+++ b/src/actions/drag_mover.ts
@@ -1,0 +1,179 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Constants from '../constants';
+import {
+  ASTNode,
+  Connection,
+  ShortcutRegistry,
+  common,
+  registry,
+  utils,
+} from 'blockly';
+import type {Block, BlockSvg, IDragger, WorkspaceSvg} from 'blockly';
+import {Mover, MoveInfo} from './mover';
+import {Navigation} from '../navigation';
+
+/**
+ * An experimental implementation of Mover that uses a dragger to
+ * perform unconstraind movment.
+ */
+export class DragMover extends Mover {
+  /**
+   * The distance to move an item, in workspace coordinates, when
+   * making an unconstrained move.
+   */
+  UNCONSTRAINED_MOVE_DISTANCE = 20;
+
+  /**
+   * Map of moves in progress.
+   *
+   * An entry for a given workspace in this map means that the this
+   * Mover is moving a block on that workspace, and will disable
+   * normal cursor movement until the move is complete.
+   */
+  protected declare moves: Map<WorkspaceSvg, DragMoveInfo>;
+
+  /**
+   * Start moving the currently-focused item on workspace, if
+   * possible.
+   *
+   * Should only be called if canMove has returned true.
+   *
+   * @param workspace The workspace we might be moving on.
+   * @returns True iff a move has successfully begun.
+   */
+  override startMove(workspace: WorkspaceSvg) {
+    const cursor = workspace?.getCursor();
+    const curNode = cursor?.getCurNode();
+    const block = curNode?.getSourceBlock() as BlockSvg | null;
+    if (!cursor || !block) throw new Error('precondition failure');
+
+    // Select and focus block.
+    common.setSelected(block);
+    cursor.setCurNode(ASTNode.createBlockNode(block)!);
+    // Begin dragging block.
+    const DraggerClass = registry.getClassFromOptions(
+      registry.Type.BLOCK_DRAGGER,
+      workspace.options,
+      true,
+    );
+    if (!DraggerClass) throw new Error('no Dragger registered');
+    const dragger = new DraggerClass(block, workspace);
+    // Record that a move is in progress and start dragging.
+    this.moves.set(workspace, new DragMoveInfo(block, dragger));
+    dragger.onDragStart(new PointerEvent('pointerdown'));
+    return true;
+  }
+
+  /**
+   * Finish moving the currently-focused item on workspace.
+   *
+   * Should only be called if isMoving has returned true.
+   *
+   * @param workspace The workspace on which we are moving.
+   * @returns True iff move successfully finished.
+   */
+  override finishMove(workspace: WorkspaceSvg) {
+    if (!workspace) return false;
+    const info = this.moves.get(workspace);
+    if (!info) throw new Error('no move info for workspace');
+
+    info.dragger.onDragEnd(
+      new PointerEvent('pointerup'),
+      new utils.Coordinate(0, 0),
+    );
+
+    this.moves.delete(workspace);
+    return true;
+  }
+
+  /**
+   * Abort moving the currently-focused item on workspace.
+   *
+   * Should only be called if isMoving has returned true.
+   *
+   * @param workspace The workspace on which we are moving.
+   * @returns True iff move successfully aborted.
+   */
+  override abortMove(workspace: WorkspaceSvg) {
+    if (!workspace) return false;
+    const info = this.moves.get(workspace);
+    if (!info) throw new Error('no move info for workspace');
+
+    // Monkey patch dragger to trigger call to draggable.revertDrag.
+    (info.dragger as any).shouldReturnToStart = () => true;
+    info.dragger.onDragEnd(
+      new PointerEvent('pointerup'),
+      new utils.Coordinate(0, 0),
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
[Deployed here](https://feat-106.blockly-keyboard-experimentation.pages.dev/) so I could play with it easily (won't automatically update for subsequent pushes)

kmcnaught:
Just to note that Alt+left arrow takes me back to the previous page in Chrome on Windows 11. (we get away with Alt+right arrow for testing as there isn't usually a subsequent page, though of course there could be)

cpcallen:
> Just to note that Alt+left arrow takes me back to the previous page in Chrome on Windows 11. (we get away with Alt+right arrow for testing as there isn't usually a subsequent page, though of course there could be)

I had originally proposed using `shift` for the unconstrained movement modifier, but you (IMHO quite correctly) objected to this on the grounds that it should be reserved for doing mutiselect.  If `alt` also doesn't work then we are quickly running out of modifiers.  For now I will make both `ctrl` and `alt` usable (`ctrl` won't work for macOS users who use [Spaces](https://support.apple.com/en-gb/guide/mac-help/mh14112/mac) with default shortcut keys), and add a note to #106 that we need to figure out what to do instead.

cpcallen:
Now provides a context menu item (on blocks) for the move action, and supports `ctrl`+arrow keys as an alternative to `alt`+arrow keys.

cpcallen:
Fixed issue with blocks being deleted when the move finished.  The fix was to provide better fake `PointerEvent`s to the dragger.  A nice side-effect of the fix is that moving a block to the trash will now delete it!  (You have to move it so that the upper-left corner of the block overlaps the trash can icon).

The blocks still lurch around a bit, and the implementation is kind of janky (don't look at what happens when you abort a move!), but it is absolutley good enough to be useful now, so I'm marking this PR as ready for review.

gonfunko:
I notice that if you move a block such that an insertion marker appears, and abort the move at that point, you get the following stack trace:

```
Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at BlockSvg$$module$build$src$core$block_svg.setParent (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1246:28)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnectInternal (webpack-internal:///./node_modules/blockly/blockly_compressed.js:983:430)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnectInternal (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1238:479)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnect (webpack-internal:///./node_modules/blockly/blockly_compressed.js:982:323)
    at BlockSvg$$module$build$src$core$block_svg.unplugFromStack (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1000:210)
    at BlockSvg$$module$build$src$core$block_svg.unplug (webpack-internal:///./node_modules/blockly/blockly_compressed.js:998:417)
    at InsertionMarkerPreviewer$$module$build$src$core$insertion_marker_previewer.hideInsertionMarker (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1410:264)
    at InsertionMarkerPreviewer$$module$build$src$core$insertion_marker_previewer.hidePreview (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1409:486)
    at BlockDragStrategy$$module$build$src$core$dragging$block_drag_strategy.revertDrag (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1223:140)
    at BlockSvg$$module$build$src$core$block_svg.revertDrag (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1272:444)
```

And wind up with an orphaned insertion marker in some cases.

src/navigation_controller.ts
```diff
@@ -99,8 +100,12 @@ export class NavigationController {
     this.canCurrentlyNavigate.bind(this),
   );
 
+  mover = new Mover(this.navigation, this.canCurrentlyEdit.bind(this));
+
   navigationFocus: NAVIGATION_FOCUS_MODE = NAVIGATION_FOCUS_MODE.NONE;
 
+  hasNavigationFocus: boolean = false;
+
```
gonfunko, line 108: I think this was inadvertently introduced resolving a merge conflict?

src/navigation_controller.ts
```diff
@@ -516,10 +522,7 @@ export class NavigationController {
    * Removes all the keyboard navigation shortcuts.
    */
   dispose() {
-    for (const shortcut of Object.values(this.shortcuts)) {
-      ShortcutRegistry.registry.unregister(shortcut.name);
-    }
-
+    this.mover.install();
```
gonfunko, line 525: `uninstall()`?

src/actions/mover.ts
```diff
@@ -0,0 +1,339 @@
...
+   *
+   * An entry for a given workspace in this map means that the this
+   * Mover is moving a block on that workspace, and will disable
+   * normal cursor movement until the move is complete.
+   */
+  protected moves: Map<WorkspaceSvg, MoveInfo> = new Map();
+
+  constructor(
+    protected navigation: Navigation,
+    canEdit: (ws: WorkspaceSvg) => boolean,
```
gonfunko, line 45: `canEdit` could also just be declared `protected` here for consistency

src/actions/mover.ts
```diff
@@ -0,0 +1,339 @@
...
+    }
+  }
+
+  /**
+   * Returns true iff we are able to begin moving the given block (or,
+   * if no block supplied, the block wich currently has focus) on the
+   * given workspace.
+   *
+   * @param workspace The workspace to move on.
+   * @returns True iff we can beign a move.
```
gonfunko, line 193: nit: begin

src/actions/mover.ts
```diff
@@ -0,0 +1,339 @@
...
+   */
+  canMove(workspace: WorkspaceSvg, block?: Block) {
+    if (!block) {
+      const cursor = workspace?.getCursor();
+      const curNode = cursor?.getCurNode();
+      block = curNode?.getSourceBlock() ?? undefined;
+    }
```
gonfunko, line 200: Might be worth factoring this out into a `getCurrentBlock()` since `startMove` also appears to need it

src/actions/mover.ts
```diff
@@ -0,0 +1,339 @@
...
+    if (!cursor || !block) throw new Error('precondition failure');
+
+    // Select and focus block.
+    common.setSelected(block as BlockSvg);
+    cursor.setCurNode(ASTNode.createBlockNode(block)!);
+
+    // Additional implementation goes here.
+    console.log('startMove');
```
gonfunko, line 243: Remove? Applies throughout, assuming this was just debugging logging.

src/actions/mover.ts
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Finish moving the currently-focused item on workspace.
+   *
+   * Should only be called if isMoving has returned true.
+   *
+   * @param workspace The workspace on which we are moving.
+   * @returns True iff move successfully finished.
+   */
+  finishMove(workspace: WorkspaceSvg) {
+    if (!workspace) return false;
```
gonfunko, line 258: Make the `workspace` arg to this function nullable or remove this null check, same for `abortMove`. Or perhaps unify them since they're currently identical?

src/actions/drag_mover.ts
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Finish moving the currently-focused item on workspace.
+   *
+   * Should only be called if isMoving has returned true.
+   *
+   * @param workspace The workspace on which we are moving.
+   * @returns True iff move successfully finished.
+   */
+  override finishMove(workspace: WorkspaceSvg) {
+    if (!workspace) return false;
```
gonfunko, line 84: As in the base class, either remove the null check or make the arg nullable here and in `abortMove`

src/actions/mover.ts
rachel-fenichel: beignet

rachel-fenichel:
Closed in favor of #352

Issue #229, opened 2025-02-21
Keyboard plugin breaks flow of mouse block drag and drop


microbit-matt-hillsdon:
I think this has been the case for a while but I finally observed it clearly.

1. [Open the demo](https://google.github.io/blockly-keyboard-experimentation/)
2. Drag but don't drop a block from the toolbox and don't significantly move the mouse 
3. Note the flyout has already closed, compare with Blockly playground where it would still be open
4. Note the block isn't even visible
5. Move the mouse and the block reappears and can be dropped as usual

Normally I do this much more quickly and it just feels a bit off vs the normal experience.

https://github.com/user-attachments/assets/f96b5c92-865a-4e1b-a7b6-11b2f464d551

kmcnaught:
Ah! I keep doing that and being momentarily flummoxed, good to know it wasn't me

kmcnaught:
Note also that you [can't use the scroll bars with the mouse](https://www.loom.com/share/932a0d12305541018318e804d83f5dec) as the flyout closes as soon as you click on the scrollbar.

microbit-matt-hillsdon:
#326 fixed this except for the scroll bar case. Perhaps a similar approach can be taken in `focusListener` in `index.ts`.

PR #326, opened 2025-03-19
feat: consistent focus handling; remove markedNode

### LINKED ISSUE #200: Plugin breaks Variables > "Create variable" mouse interaction
1. In [the demo](https://google.github.io/blockly-keyboard-experimentation/) click Variables in the toolbox
2. Click "Create variable"
3. Note that the flyout closes without prompting for a variable name

This is working fine in e.g. Blockly playground where the flyout stays open with the newly created variable visible.

If you use keyboard nav then the variable is successfully created but the flyout closes which differs from the (normal, working) mouse behaviour. That may have a separate cause in which case I can split into a new issue.

https://github.com/user-attachments/assets/c23216e4-2f6f-4dac-b60a-2143de62d106

### LINKED ISSUE #222: Flyout auto-closes after creating a variable

### LINKED ISSUE #227: Persistent marker after copying from toolbox
I've discovered a flow that can cause a persistent blue marker to appear at a next/previous connection.

Steps to reproduce:
- Navigate to a next/previous connection
- Hit "T" to open the toolbox
- Navigate to a block and use the keyboard shortcut to copy the block
- The flyout is closed and a blue marker remains at the next/previous position if you do not immediately paste
- It is persistent during further keyboard navigation until some other meaningful block interaction

Video:
https://github.com/user-attachments/assets/a66a7720-1d05-4621-9b80-2b4c2e2020e9

### LINKED ISSUE #287: Click and drag on initial load selects a block and drags it rather than panning
This breaks a common mouse interaction when viewing larger projects.

https://github.com/user-attachments/assets/74a5ee31-c1fb-40de-ac00-698cb5c4682a

microbit-matt-hillsdon:
Ensure we always focus the same element when keyboard navigation works in that element. Previously the toolbox sometimes was and sometimes wasn't focused when it navigated, the workspace group or the workspace SVG could be focused and the flyout never took the focus.

With this approach the relevant container always takes focus and we explicitly move browser focus more often and let the focus/blur handlers do their work.

Manage the passive focus indicator based on focus rather than markedNode state. But we take care not to make a selection when focus returns due to a click as that can interrupt gestures.

Integrate with Gesture to determine how mouse behaviours relate to focus. This prevents issues where a gesture would cause a focus change which would close the flyout, interrupting a normal drag. In time hopefully this code can move to core where it will be more natural and less fragile.

Fixes #227
Fixes #200
Part of #229 (scrollbars still an issue)
Fixes #222
Fixes #287

Compromises:
- Knows about the flyout DOM for focus purposes (similar to toolbox before). I hope that focus changes in core can give API for this need.
- Uses more gesture internals. I think clean code with the same effect can be done in core.

Future steps:
- Make the flyout cursor and movement actions behave the same way, so clicking to create a variable (or on the flyout itself) doesn't move the cursor the first thing in the flyout.
- Always clear the current node when clicking on a workspace click. This happens for selection (which will clear the cursor via the listener) but not if you're e.g. at next/previous. The inconsistency feels weird.

**Demo**: https://rejig-focus.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/gesture_monkey_patch.js
+++ b/src/gesture_monkey_patch.js
@@ -5,55 +5,27 @@
  */
 
 /**
- * @fileoverview Overrides methods on Blockly.Gesture in order to allow users
- * to move the cursor to blocks or the workspace using shift click.
- * TODO(google/blockly#4584): We do not have a way to do this currently without
- * monkey patching Blockly.
+ * @fileoverview Overrides methods on Blockly.Gesture to integrate focus mangagement
+ * with the gesture handling.
  * @author aschmiedt@google.com (Abby Schmiedt)
  */
 
 import * as Blockly from 'blockly/core';
 
-const oldDoWorkspaceClick = Blockly.Gesture.prototype.doWorkspaceClick_;
+const oldDispose = Blockly.Gesture.prototype.dispose;
 
 /**
- * Execute a workspace click. When in accessibility mode shift clicking will
- * move the cursor.
- * @param {!Event} e A mouse up or touch end event.
+ * Intercept the end of a gesture and ensure the workspace is focused.
+ * See also the listener is index.ts that subverts the markFocused call
+ * in `doStart`.
  * @this {Blockly.Gesture}
  * @override
  */
-Blockly.Gesture.prototype.doWorkspaceClick_ = function (e) {
-  oldDoWorkspaceClick.call(this, e);
-  const ws = this.creatorWorkspace_;
-  if (e.shiftKey && ws.keyboardAccessibilityMode) {
-    const screenCoord = new Blockly.utils.Coordinate(e.clientX, e.clientY);
-    const wsCoord = Blockly.utils.svgMath.screenToWsCoordinates(
-      ws,
-      screenCoord,
-    );
-    const wsNode = Blockly.ASTNode.createWorkspaceNode(ws, wsCoord);
-    ws.getCursor().setCurNode(wsNode);
+Blockly.Gesture.prototype.dispose = function () {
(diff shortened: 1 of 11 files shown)


microbit-matt-hillsdon:
So this might conflict badly with ongoing work to rejig focus with changes in core, but I think it's worth some thought. Tagging @BenHenning where I expect some overlap.

src/gesture_monkey_patch.js
```diff
@@ -5,55 +5,27 @@
...
- * Execute a workspace click. When in accessibility mode shift clicking will
- * move the cursor.
- * @param {!Event} e A mouse up or touch end event.
+ * Intercept the end of a gesture and ensure the workspace is focused.
+ * See also the listener is index.ts that subverts the markFocused call
+ * in `doStart`.
  * @this {Blockly.Gesture}
  * @override
  */
-Blockly.Gesture.prototype.doWorkspaceClick_ = function (e) {
```
microbit-matt-hillsdon, line 26: This removed code has been broken for a while (underscore change) and I don't think we want this behaviour now selection is better integrated with focus.

microbit-matt-hillsdon:
Demo: https://rejig-focus.blockly-keyboard-experimentation.pages.dev/

microbit-matt-hillsdon:
Latest change introduced an issue returning after using field editors. Reconsidering and moving to draft.

rachel-fenichel (approved):

PR #331, opened 2025-03-24
fix: right click menu issues

microbit-matt-hillsdon:
PR #326 introduced an issue with stale current node when showing the context menu by mouse. It's only possible to fix this by reinstating the updating of curNode from the selection in `getCurNode`.

We also attempt to avoid showing passive focus when using the context menu and widgets (via mouse and keyboard) for consistency with the normal Blockly experience.

You can now no longer get this scenario:

image

Fixes an issue clicking on shadow blocks which could previously show an outline around the shadow and parent block. You can now no longer get these scenarios:

image

image

Demo: https://fix-right-click-2.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -475,20 +475,56 @@ export class LineCursor extends Marker {
   }
 
   /**
+   * Get the current location of the cursor.
+   *
+   * Overrides normal Marker getCurNode to update the current node from the selected
+   * block. We typically update the node from the selection via a listener but
+   * that is not called immediately when `Gesture` calls `Blockly.common.setSelected`.
+   * In particular the listener runs after showing the context menu.
+   *
+   * @returns The current field, connection, or block the cursor is on.
+   */
+  override getCurNode(): ASTNode {
+    const curNode = super.getCurNode();
+    let selected = Blockly.common.getSelected();
+    if (
+      selected === null &&
+      curNode?.getType() === Blockly.ASTNode.types.BLOCK
+    ) {
+      // Selection says our curNode is not selected anymore.
+      // this.setCurNode(null as never, true);
+      return super.getCurNode();
+    }
+    if (selected?.workspace !== this.workspace) return curNode;
+    // Selection has a block in our workspace.
+    if (selected instanceof Blockly.BlockSvg) {
+      if (selected.isShadow()) {
+        // Although the shadow block is selected it's the parent that has the
+        // visual selection.
+        selected = selected.getParent();
+      }
+      if (selected) {
+        this.setCurNode(new ASTNode(ASTNode.types.BLOCK, selected), true);
+      }
+    }
+
+    return super.getCurNode();
+  }
+
+  /**
    * Set the location of the cursor and draw it.
    *
    * Overrides normal Marker setCurNode logic to call
    * this.drawMarker() instead of this.drawer.draw() directly.
    *
    * @param newNode The new location of the cursor.
+   * @param selectionUpToDate If false (the default) we'll update the selection too.
    */
-  override setCurNode(newNode: ASTNode, selectionInSync = false) {
-    if (newNode?.getLocation() === this.getCurNode()?.getLocation()) {
-      return;
-    }
-    if (!selectionInSync) {
+  override setCurNode(newNode: ASTNode, selectionUpToDate = false) {
+    if (!selectionUpToDate) {
       if (
         newNode?.getType() === ASTNode.types.BLOCK &&
+        // For shadow blocks we need to clear the selection that's drawn on their parent.
         !(newNode.getLocation() as Blockly.BlockSvg).isShadow()
       ) {
         if (Blockly.common.getSelected() !== newNode.getLocation()) {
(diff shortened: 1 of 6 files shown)


src/navigation.ts
```diff
@@ -432,6 +440,34 @@ export class Navigation {
...
+  ) {
+    if (relatedTarget === null) {
+      // Workaround:
+      // Skip document.body/null case until these blur bugs are fixed to avoid
+      // flipping to passive focus as the user moves their mouse over the
+      // dropdown/widget at the cost of clicks on body showing the wrong focus
+      // style.
+      // https://github.com/google/blockly-samples/issues/2498
+      // https://github.com/google/blockly/issues/8819
+      return;
```
microbit-matt-hillsdon: Update now Blockly is upgraded fixing one of the bugs.
We could also just drop this but it's very weird to experience in the demo with the colour picker.

```suggestion
      // Workaround:
      // Skip document.body/null case until this blur bugs is fixed to avoid
      // flipping to passive focus as the user moves their mouse over the
      // colour picker.
      // https://github.com/google/blockly-samples/issues/2498
      return;
```

src/navigation.ts
```diff
@@ -432,6 +440,32 @@ export class Navigation {
...
+   * @param workspace The workspace.
+   * @param relatedTarget The related target (newly focused element if any).
+   */
+  handleFocusOutWidgetDropdownDiv(
+    workspace: Blockly.WorkspaceSvg,
+    relatedTarget: EventTarget | null,
+  ) {
+    if (relatedTarget === null) {
+      // Workaround:
+      // Skip document.body/null case until this blur bugs is fixed to avoid
```
BenHenning:
```suggestion
      // Skip document.body/null case until this blur bug is fixed to avoid
```

Nit on extra 's'.

src/line_cursor.ts
```diff
@@ -475,6 +475,43 @@ export class LineCursor extends Marker {
...
+    if (selected?.workspace !== this.workspace) return curNode;
+    // Selection has a block in our workspace.
+    if (selected instanceof Blockly.BlockSvg) {
+      if (selected.isShadow()) {
+        // Although the shadow block is selected it's the parent that has the
+        // visual selection.
+        selected = selected.getParent();
+      }
+      if (selected) {
+        this.setCurNode(new ASTNode(ASTNode.types.BLOCK, selected), true);
```
BenHenning: Could you perhaps use `Blockly.ASTNode.createBlockNode` here, instead? That's conventional with other locations in the plugin where an `ASTNode` is being created.

src/line_cursor.ts
```diff
@@ -673,22 +709,16 @@ export class LineCursor extends Marker {
...
-        const node = ASTNode.createBlockNode(block);
-        if (node) {
-          this.setCurNode(node, true);
-        }
-      }
-    } else {
-      this.setCurNode(null as never, true);
-    }
+    // This runs too late so the logic to update the selection is in
+    // `getCurNode`.
```
BenHenning: The context here is missing. It makes sense when reading the PR, but this comment by itself post-merge will be hard to understand without blame. Can more context perhaps be added?

Separately, might it be possible to factor out the bits of `getCurNode` which actually update the state into a common helper called both there and here? That might be the cleanest way to indicate what needs to happen in both locations.

BenHenning (approved):
Thanks @microbit-matt-hillsdon! I think the fixes make sense. I'm happy to approve the PR, though I did have a few comments on clarity. Feel free to self-resolve and merge, or assign me back if you'd like me to take another review pass.

src/line_cursor.ts
```diff
@@ -475,6 +475,43 @@ export class LineCursor extends Marker {
...
+   */
+  override getCurNode(): ASTNode | null {
+    const curNode = super.getCurNode();
+    let selected = Blockly.common.getSelected();
+    if (
+      selected === null &&
+      curNode?.getType() === Blockly.ASTNode.types.BLOCK
+    ) {
+      // Selection says our curNode is not selected anymore.
+      // this.setCurNode(null as never, true);
```
microbit-matt-hillsdon: Oops, this is definitely a problem. Re-enabling it threw up a shadow block issue addressed in 1233f578fd94a0941a8e028a0417d7e83a3eda3f

src/line_cursor.ts
microbit-matt-hillsdon: Makes sense. Done as part of 1233f578fd94a0941a8e028a0417d7e83a3eda3f.

src/line_cursor.ts
microbit-matt-hillsdon: That really helped. 1233f578fd94a0941a8e028a0417d7e83a3eda3f has separate methods for the node -> selection and selection -> node updates and uses the selection -> node one in the listener and `getCurNode`.

microbit-matt-hillsdon: Ultimately I hope https://github.com/google/blockly/issues/8820 might make this less messy

src/navigation.ts
microbit-matt-hillsdon: Thanks, fixed in 586a6f0734847fc76e0871619aba6a615f8ae17d

microbit-matt-hillsdon:
Thanks for the review @BenHenning. Assigning back as it would be good to take another look (and I can't merge it in any case).

src/line_cursor.ts
BenHenning: I find the latest changes much cleaner--thanks!

src/line_cursor.ts
BenHenning: I actually missed this commented out bit the first time around--great catch and thanks for fixing it in the follow-up!

BenHenning (approved):
Thanks @microbit-matt-hillsdon. Erm, good point on the merging bit. This looks good to me!

Issue #336, opened 2025-03-25
Create a build step for webdriverio tests


rachel-fenichel:
Followup to #335 

- Add a build step to build for webdriverio testing and put the generated code as a bundle in a place that is accessible to the tests
  - This may be similar to the existing `ghpages` script, which has an associated webpack config
- Add scripts to clean and rebuild for testing
- Expose those scripts in `package.json`

PR #342, opened 2025-03-25
feat: test page for webdriverio

### LINKED ISSUE #335: Create a test page for webdriverio tests
#333 runs tests on `test/index.html`. Mixing manual and automated testing in this way is fragile. Instead:
- Create a directory for webdriverio tests (`test/webdriverio`)
- Create a test page (`index.html` and `index.ts`) that is similar to the existing test page but removes
  - The execution area
  - Code for generating and running the code
  - Left-side checkboxes
- Retain the ability to set the renderer through the URL

rachel-fenichel:
Fixes #335 and #336

- Adds a test page for webdriverio testing (`index.html` and `index.ts`) that looks similar to the standard test page but removes the ability to run code or set options through a form on the page. Setting options through the URL is still allowed.
- Adds a webpack config to build a static HTML page with all associated imports. This means we won't need to run a server to do these tests.
  - The output is in `test/webdriverio/build`
- Add scripts for `wdio:build` and `wdio:clean` to `package.json`

### DIFF
--- a/test/webdriverio/index.html
+++ b/test/webdriverio/index.html
@@ -3,13 +3,6 @@
   <head>
     <meta charset="utf-8" />
     <title>Keyboard-experimentation webdriverio page</title>
-    <link rel="preconnect" href="https://fonts.googleapis.com" />
-    <link
-      rel="stylesheet"
-      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap" />
-    <link
-      rel="stylesheet"
-      href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200" />
     <style>
       body {
         font-family: Roboto, sans-serif;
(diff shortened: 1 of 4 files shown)


test/webdriverio/index.html
```diff
@@ -0,0 +1,92 @@
...
+    <title>Keyboard-experimentation webdriverio page</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap" />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200" />
+    <style>
+      body {
```
RoboErikG, line 7: The font styles don't seem to be used for anything. Do these need to be included or am I missing where they're being applied?

RoboErikG (approved):
Overall looks good. I think there's some extra css, though. If I'm mistaken feel free to merge.

test/webdriverio/index.html
rachel-fenichel: Good point. Removed.

PR #348, opened 2025-03-28
chore: fix lint errors and run lint in CI

### LINKED ISSUE #345: Fix lint errors and enable lint in CI
`npm run lint` fails - I'm not convinced it has passed any time recently.

Please can we run it in CI and either fix/ignore the errors or adjust the rules to suit the team?

If you're happy with the existing rules I can take a first pass through fixing and flag controversial bits here. It seems likely to mostly be uncontroversial fixes except for a few places where some Blockly naming conventions appear to differ from the rules. I think it would make sense to ignore those rules on a file or line basis in a first pass.

I tend to have eslint diagnostics enabled in my editor and find all the lint errors distract me from the new errors I'm introducing 😄  

## ASTNode createXXX return type

There's an annoyingly prevalent pattern of a non-null assertion (not allowed by the rules) in code like `ASTNode.createBlockNode(newBlock)!` that could do with a separate fix. Why can it (and similar, but mainly this one in practice) return null? The implementation seems to be undecided about its types:

```
  static createBlockNode(block: Block): ASTNode | null {
    if (!block) {
      return null;
    }
    return new ASTNode(ASTNode.types.BLOCK, block);
  }
```

microbit-matt-hillsdon:
Fixes #345 

This adds ignores for the non-null checks that relate to createXXXNode. It wasn't as bad as I expected as `setCurNode` permits null now. Raised https://github.com/google/blockly/issues/8827 in case there's interest in revisiting the signature or behaviour here.

It might also be nice to explore converting the tests to TypeScript before we make more of them. It didn't seem too hard to fix the current lint issues there so I've left lint enabled for them.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -129,7 +129,7 @@ module.exports = [
     },
   },
   {
-    files: ['**/*.mocha.js'],
+    files: ['**/*.mocha.js', 'test/webdriverio/test/*_test.mjs'],
     languageOptions: {
       globals: {
         ...globals.mocha,
--- a/src/actions/action_menu.ts
+++ b/src/actions/action_menu.ts
@@ -10,7 +10,6 @@ import {
   ContextMenu,
   ContextMenuRegistry,
   ShortcutRegistry,
-  comments,
   utils as BlocklyUtils,
   WidgetDiv,
 } from 'blockly';
@@ -23,10 +22,7 @@ const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
   ShortcutRegistry.registry,
 );
 
-export interface Scope {
-  block?: BlockSvg;
-  workspace?: WorkspaceSvg;
-  comment?: comments.RenderedWorkspaceComment;
+export interface ScopeWithConnection extends ContextMenuRegistry.Scope {
   connection?: Connection;
 }
 
@@ -100,6 +96,8 @@ export class ActionMenu {
    * Returns true if it is possible to open the action menu in the
    * current location, even if the menu was not opened due there being
    * no applicable menu items.
+   *
+   * @param workspace The workspace.
    */
   private openActionMenu(workspace: WorkspaceSvg): boolean {
     let menuOptions: Array<
@@ -114,7 +112,7 @@ export class ActionMenu {
     if (!node) return false;
     const nodeType = node.getType();
     switch (nodeType) {
-      case ASTNode.types.BLOCK:
+      case ASTNode.types.BLOCK: {
         const block = node.getLocation() as BlockSvg;
         rtl = block.RTL;
         // Reimplement BlockSvg.prototype.generateContextMenu as that
@@ -130,11 +128,12 @@ export class ActionMenu {
         }
         // End reimplement.
         break;
+      }
 
       // case Blockly.ASTNode.types.INPUT:
       case ASTNode.types.NEXT:
       case ASTNode.types.PREVIOUS:
-      case ASTNode.types.INPUT:
+      case ASTNode.types.INPUT: {
         const connection = node.getLocation() as Connection;
         rtl = connection.getSourceBlock().RTL;
 
@@ -143,6 +142,7 @@ export class ActionMenu {
         // a possible kind of scope.
         this.addConnectionItems(connection, menuOptions);
         break;
+      }
 
       default:
         console.info(`No action menu for ASTNode of type ${nodeType}`);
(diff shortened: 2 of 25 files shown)


eslint.config.js
```diff
@@ -129,7 +129,7 @@ module.exports = [
     },
   },
   {
-    files: ['**/*.mocha.js'],
+    files: ['**/*.mocha.js', 'test/webdriverio/test/*_test.mjs'],
```
microbit-matt-hillsdon, line 132: Alternatively we could rename `basic_test.mjs` to `basic.mocha.mjs` and just tweak the pattern to add the `mjs` extension.

src/keynames.ts
```diff
@@ -154,29 +166,23 @@ function keyCodeToString(keycode, index) {
     piece = pieces[i];
     strrep = keyNames[piece] ?? piece;
     const className = getKeyClassName(strrep);
-
-    if (i.length === 1) {
```
microbit-matt-hillsdon, line 158: This line is clearly mad. In practice it has always been undefined so the else branch was taken. The title case version of a single character is the same as uppercasing it anyway.

src/actions/ws_movement.ts
```diff
@@ -13,6 +13,11 @@ const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
   ShortcutRegistry.registry,
 );
 
+/**
+ * The distance to move the cursor when the cursor is on the workspace.
+ */
+const WS_MOVE_DISTANCE = 40;
```
microbit-matt-hillsdon, line 19: Moved from the class, but not sure if this was intended to be mutable as a sort of configuration.

src/navigation.ts
```diff
@@ -24,6 +24,22 @@ import {
...
+ */
+const DEFAULT_WS_COORDINATE: Blockly.utils.Coordinate =
+  new Blockly.utils.Coordinate(100, 100);
+
+/**
+ * The default coordinate to use when moving the cursor to the workspace
+ * after a block has been deleted. In pixel coordinates, but will be
+ * converted to workspace coordinates when used to position the cursor.
+ */
+const WS_COORDINATE_ON_DELETE: Blockly.utils.Coordinate =
```
microbit-matt-hillsdon, line 40: Moved from the class, but not sure if this was intended to be mutable as a sort of configuration.

eslint.config.js
RoboErikG: Let's leave it as _test.mjs to match what's in core Blockly.

src/actions/action_menu.ts
```diff
@@ -143,6 +142,7 @@ export class ActionMenu {
         // a possible kind of scope.
         this.addConnectionItems(connection, menuOptions);
         break;
+      }
 
       default:
```
RoboErikG, line 147: Shouldn't this also trigger the switch-case-braces eslint check?

src/actions/action_menu.ts
```diff
@@ -205,14 +202,33 @@ export class ActionMenu {
...
+  /**
+   * Find a context menu action, throwing an `Error` if it is not present or
+   * not an action. This usefully narrows the type to `ActionRegistryItem`
+   * which is not exported from Blockly.
+   *
+   * @param id The id of the action.
+   * @returns the action.
+   */
+  private getContextMenuAction(id: string) {
+    const item = ContextMenuRegistry.registry.getItem('insert');
```
RoboErikG: Should be getting the id

src/keynames.ts
```diff
@@ -19,7 +19,8 @@
  *
  * Copied from goog.events.keynames
  */
-const keyNames = {
+const keyNames: Record<string, string> = {
```
RoboErikG, line 22: Should this be <number, string> or is there a reason to interpret the key code as a string?

src/keynames.ts
RoboErikG: It looks like the purpose was to avoid an exception if strrep is a single character. I'm not sure if that ever happens in practice, but might be worth adding a check into toTitleCase just in case?

microbit-matt-hillsdon: I don't think there's any need `toTitleCase("a")` will result in `"A"` via `'a'.toUpperCase() + ''.toLowerCase()`.

src/actions/action_menu.ts
microbit-matt-hillsdon: I don't think that rule is enabled. The rule the other case failed was https://eslint.org/docs/latest/rules/no-case-declarations which is a scope-confusion-related rule.

src/actions/action_menu.ts
microbit-matt-hillsdon: Good catch, thanks.

a264d3533a4b90b3421049aad974a7dedaec302e

src/keynames.ts
microbit-matt-hillsdon: I did that initially but it's only used to look up strings so that makes the only usage hard.

microbit-matt-hillsdon: Usage is in `keyCodeToString` using parts of a shortcut like `Control+88` split on "+".

(Edited to use a correct example input)

src/actions/action_menu.ts
RoboErikG: Ah, thanks for pointing me at the right eslint rule. 😀

src/keynames.ts
RoboErikG: Makes sense.

src/keynames.ts
RoboErikG: Ah, fair enough. I'd missed that substring returns an empty string in that case.

src/navigation.ts
RoboErikG: I suspect this will need to be configurable in the final version, but I think for now this makes sense as a const here.

RoboErikG (approved):

PR #352, opened 2025-03-29
feat: introduce `Mover` actions + experimental `DragMover` implementation

rachel-fenichel:
This is a rebuild/rebase of Christopher's https://github.com/google/blockly-keyboard-experimentation/pull/317.

I rebased it, tested that it worked, and then addressed Aaron's feedback in his review of #317.

I left alt in place because of Christopher's note about Mac users ("For now I will make both ctrl and alt usable (ctrl won't work for macOS users who use [Spaces](https://support.apple.com/en-gb/guide/mac-help/mh14112/mac) with default shortcut keys), and add a note to https://github.com/google/blockly-keyboard-experimentation/issues/106 that we need to figure out what to do instead.") but long-term I think we should actually move it to use shift+arrow for constrained and arrow for unconstrained. That matches what other dragging-related software does, generally.

I then fixed lint errors and formatted.

~I was not (yet) able to resolve the issue with aborting a drag when the insertion marker is shown.~

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -71,7 +71,7 @@ export class Mover {
       allowCollision: true,
     },
     {
-      name: 'Move right unconstraind',
+      name: 'Move right unconstrained',
       preconditionFn: (workspace) => this.isMoving(workspace),
       callback: (workspace) => this.moveConstrained(workspace /* , ... */),
       keyCodes: [KeyCodes.RIGHT],
@@ -103,7 +103,7 @@ export class Mover {
       ],
     },
     {
-      name: 'Move right, unconstraind',
+      name: 'Move right, unconstrained',
       preconditionFn: (workspace) => this.isMoving(workspace),
       callback: (workspace) => this.moveUnconstrained(workspace, 1, 0),
       keyCodes: [
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -5,11 +5,11 @@
  */
 
 import * as Constants from '../constants';
-import {ASTNode, ShortcutRegistry, utils as BlocklyUtils} from 'blockly';
-import type {Block, WorkspaceSvg} from 'blockly';
+import {ASTNode, Connection, ShortcutRegistry, common, utils} from 'blockly';
+import type {Block, BlockSvg, WorkspaceSvg} from 'blockly';
 import {Navigation} from '../navigation';
 
-const KeyCodes = BlocklyUtils.KeyCodes;
+const KeyCodes = utils.KeyCodes;
 const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
   ShortcutRegistry.registry,
 );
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
@cpcallen @maribethb @microbit-matt-hillsdon fyi

src/actions/mover.ts
```diff
@@ -0,0 +1,334 @@
...
+    // Constrained moves.
+    {
+      name: 'Move left, constrained',
+      preconditionFn: (workspace) => this.isMoving(workspace),
+      callback: (workspace) => this.moveConstrained(workspace /* , ...*/),
+      keyCodes: [KeyCodes.LEFT],
+      allowCollision: true,
+    },
+    {
+      name: 'Move right unconstraind',
```
gonfunko: Nit: constrained

src/actions/mover.ts
```diff
@@ -0,0 +1,334 @@
...
+      name: 'Move left, unconstrained',
+      preconditionFn: (workspace) => this.isMoving(workspace),
+      callback: (workspace) => this.moveUnconstrained(workspace, -1, 0),
+      keyCodes: [
+        createSerializedKey(KeyCodes.LEFT, [KeyCodes.ALT]),
+        createSerializedKey(KeyCodes.LEFT, [KeyCodes.CTRL]),
+      ],
+    },
+    {
+      name: 'Move right, unconstraind',
```
gonfunko: Nit: unconstrain*e*d

src/actions/mover.ts
```diff
@@ -0,0 +1,334 @@
...
+    const cursor = workspace?.getCursor();
+    const block = this.getCurrentBlock(workspace);
+    if (!cursor || !block) throw new Error('precondition failure');
+
+    // Select and focus block.
+    common.setSelected(block);
+    cursor.setCurNode(ASTNode.createBlockNode(block));
+
+    // Additional implementation goes here.
+    console.log('startMove');
```
gonfunko, line 226: Remove? Applies throughout these next few methods.

src/actions/mover.ts
```diff
@@ -0,0 +1,334 @@
...
+  abortMove(workspace: WorkspaceSvg) {
+    const info = this.moves.get(workspace);
+    if (!info) throw new Error('no move info for workspace');
+
+    // Additional implementation goes here.
+    console.log('abortMove');
+
+    this.moves.delete(workspace);
+    return true;
+  }
```
gonfunko, line 268: This appears identical to `finishMove()`;  perhaps rename to `finishOrAbort()`, or have this just call `finishMove()`?

gonfunko (approved):

src/actions/mover.ts
rachel-fenichel: I filed https://github.com/google/blockly-keyboard-experimentation/issues/357 to track removing all of these if we don't need them. I  think the point was to support multiple types of Movers in case we want to compare two different types of behaviour, but I'm not sure if we'll need to do that.

@cpcallen to confirm.

The same applies for `finishMove` and `abortMove`, which ar ethe same here but distinct in the drag mover.

PR #353, opened 2025-03-31
chore: convert webdriverio tests to TypeScript

microbit-matt-hillsdon:
There was some debate as to whether this was worthwhile, particularly whether it might impede pulling tests into Blockly core in future so marking draft for the moment.

As I understand it this is with @gonfunko to consider impact on core of taking this route.

### DIFF
--- a/test/webdriverio/test/test_setup.ts
+++ b/test/webdriverio/test/test_setup.ts
@@ -89,7 +89,7 @@ export async function driverTeardown() {
  *
  * @param playgroundUrl The URL to open for the test, which should be
  *     a Blockly playground with a workspace.
- * @returns A Promsie that resolves to a webdriverIO browser that tests can manipulate.
+ * @returns A Promise that resolves to a webdriverIO browser that tests can manipulate.
  */
 export async function testSetup(
   playgroundUrl: string,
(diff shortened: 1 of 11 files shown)


microbit-matt-hillsdon:
I can't change the reviewer otherwise I'd move this to @gonfunko

test/webdriverio/test/test_setup.ts
```diff
@@ -78,23 +75,27 @@ export async function driverSetup() {
...
 }
 
 /**
  * Navigate to the correct URL for the test, using the shared driver.
- * @param {string} playgroundUrl The URL to open for the test, which should be
+ *
+ * @param playgroundUrl The URL to open for the test, which should be
  *     a Blockly playground with a workspace.
- * @returns {Promise<webdriverio.Browser>} A Promsie that resolves to a webdriverIO browser that tests can manipulate.
+ * @returns A Promsie that resolves to a webdriverIO browser that tests can manipulate.
```
gonfunko: Nit: promise

gonfunko (approved):
Thanks! This looks quite reasonable to me, and doesn't seem like it would be overly painful to merge back to core.

test/webdriverio/test/test_setup.ts
microbit-matt-hillsdon: Thanks, 1ba7ae46d3719d1e2a2a9dd77a26b3cbfb81a490

(I think this one originates in the same file in core)

PR #377, opened 2025-04-02
fix: consolidate mover class and monkeypatch isDragging

### LINKED ISSUE #357: Consolidate Mover classes
Christopher created a Mover class and a DragMover subclass, with room to experiment on how moving works. It looks like we probably just need one Mover for both constrained and unconstrained movement. If we decide to just iterate within DragMover, delete Mover and consolidate wrapper code into DragMover.

### LINKED ISSUE #359: Monkeypatch `WorkspaceSvg.isDragging` to return true if a keyboard "drag" is in progress
Current implementation:

```
  /**
   * Is the user currently dragging a block or scrolling the flyout/workspace?
   *
   * @returns True if currently dragging or scrolling.
   */
  isDragging(): boolean {
    return this.currentGesture_ !== null && this.currentGesture_.isDragging();
  }
```

This is causing issues with testing the constrained and unconstrained dragging because an interaction with the `disableOrphans` listener makes the moving block appear disabled.

### Proposed solution

In `DragMover.startMove`, monkeypatch `isDragging` to return true.
In `DragMover.finishMove` and `DragMover.abortMove`, undo the monkeypatch and restore the normal function.

### Additional information

Longer term, we'll need the workspace to know whether a keyboard drag is in progress, because the keyboard drags won't be registered as gestures.

rachel-fenichel:
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/357
- Consolidates DragMover into Mover. I left the DragMover/Mover distinction in place in https://github.com/google/blockly-keyboard-experimentation/pull/352 to keep as close as possible to the original PR. Now that I'm working directly in movement code, I really just want one class instead of making coordinated changes in both.
- This follows feedback from @gonfunko in that review.

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/359
- Patches `isDragging` in `mover.ts`

Tested by:
- Doing keyboard drags that end in an Esc
- Doing keyboard drags that end in an Enter
- Doing mouse drags after the fact to confirm that mouse dragging still works


gonfunko (approved):

src/actions/mover.ts
```diff
@@ -34,6 +42,12 @@ export class Mover {
    */
   protected moves: Map<WorkspaceSvg, MoveInfo> = new Map();
 
+  /**
+   * The stashed isDragging function, which is replaced at the beginning
+   * of a keyboard drag and reset at the end of a keyboard drag.
+   */
+  oldIsDragging: (() => boolean) | null = null;
```
cpcallen, line 49: This should have been per-workspace (and patching/unpatching handled carefully) because `Mover` can handle multiple simultaneous moves on one or more workspaces.

Issue #364, opened 2025-04-01
Create a new `KeyboardDragStrategy` class


rachel-fenichel:
The strategy should
- Extend `BlockDragStrategy` from core
- Have a `getCandidateConnectionConstrained` that takes in a direction and returns the nearest candidate in that direction.
- Use `getCandidateConnectionConstrained` or `super.getCandidateConnection` based on the arrow keys and whether shift was pressed.

rachel-fenichel:
Issues encountered while working on this (list to be continued):
- `BlockDragStrategy` is not exported (but can be accessed by importing `dragging` and accessing `dragging.BlockDragStrategy`.
- Every instance of `ts-expect-error` in my code.
- `ConnectionCandidate` is not exported from core, so I had to redeclare it.

Issue #363, opened 2025-04-01
Implement constrained dragging


rachel-fenichel:
Both unconstrained and constrained dragging use a lot of the drag infrastructure from core, because it already handles adding and removing previews, reverting drags, and so on. 

Based on investigations in #323, the least impactful way to change code to add constrained dragging is to override the dragging block's `DragStrategy`.

Steps:

rachel-fenichel:
Closing as constrained dragging (and unconstrained dragging) now exist, and remaining issues are tracked individually.

PR #379, opened 2025-04-02
feat: monkeypatch the block's drag strategy during keyboard drags

### LINKED ISSUE #373: Modify block's drag strategy for constrained dragging
- In `DragMover.startMove`, replace the dragging block's drag strategy with an instance of the new `KeyboardDragStrategy` class. 
- Cache the old value.
- In `DragMover.finishMove` and `DragMover.abortMove`, reapply the cached `BlockDragStrategy`.

rachel-fenichel:
Fixes #373.

Adds a drag strategy class and the appropriate patching and unpatching at beginning and end of drag.

The new class extends core's `BlockDragStrategy`. Populating it with useful functions is tracked in https://github.com/google/blockly-keyboard-experimentation/issues/364.

Part of #363

Tested by:

Doing keyboard drags that end in an Esc
Doing keyboard drags that end in an Enter
Doing mouse drags after the fact to confirm that mouse dragging still works

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -15,7 +15,7 @@ import {
   registry,
   utils,
 } from 'blockly';
-import type {Block, BlockSvg, IDragger, IDragStrategy} from 'blockly';
+import type {BlockSvg, IDragger, IDragStrategy} from 'blockly';
 import {Navigation} from '../navigation';
 import {KeyboardDragStrategy} from '../keyboard_drag_strategy';
 
@@ -279,7 +279,7 @@ export class Mover {
     );
 
     this.unpatchIsDragging(workspace);
-    this.unpatchDragStrategy(info.block as BlockSvg);
+    this.unpatchDragStrategy(info.block);
     this.moves.delete(workspace);
     return true;
   }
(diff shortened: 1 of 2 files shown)


src/actions/mover.ts
```diff
@@ -271,6 +279,7 @@ export class Mover {
     );
 
     this.unpatchIsDragging(workspace);
+    this.unpatchDragStrategy(info.block as BlockSvg);
```
RoboErikG: out of curiosity, is there a reason MoveInfo.block is a Block instead of a BlockSvg given you assume it'll always be the latter?

RoboErikG (approved):
Looks fine, just a question on MoveInfo types which isn't directly part of this change.

src/actions/mover.ts
rachel-fenichel: Good point--and it becomes very obvious with these casts. Updated.

src/actions/mover.ts
```diff
@@ -48,6 +49,12 @@ export class Mover {
    */
   oldIsDragging: (() => boolean) | null = null;
 
+  /**
+   * The block's base drag strategy, which will be overridden during
+   * keyboard drags and reset at the end of the drag.
+   */
+  private oldDragStrategy: IDragStrategy | null = null;
```
cpcallen, line 56: This should be on `MoveInfo` because `Mover` can handle multiple simultaneous moves on different workspaces.

Issue #371, opened 2025-04-01
Add a helper to center the new candidate position during a keyboard drag


rachel-fenichel:
Otherwise it will be easy to move a constrained or unconstrained drag to somewhere far off the workspace and not know what happened.

Alternative to centering: scroll into view.

Issue #370, opened 2025-04-01
Add a helper to position the moving block near the appropriate connection during keyboard drag


rachel-fenichel:
The location of the moving block during keyboard drag is all over the place.

This is mostly because existing dragging code assumes that the source of truth is a pointer event with location information, and no such source of truth exists in the keyboard case.

Issue #369, opened 2025-04-01
Add helpers to loop around from the last block of the workspace to the first (and vice versa) when doing constrained movement.


rachel-fenichel:
We should match the behaviour of most search bars, where going past the last result goes back to the beginning of the content, rather than stopping.

PR #384, opened 2025-04-03
feat: add constrained movement

### LINKED ISSUE #364: Create a new `KeyboardDragStrategy` class

rachel-fenichel:
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/364

This PR adds constrained movement by selectively extending and overriding parts of the drag strategy. In particular:
- Adds a `searchNode` property, which is where a constrained search should start traversing the tree.
- Overrides `currCandidateIsBetter` to always return false for a constrained drag. That is, the new candidate is always preferred.
- Overrides `getConnectionCandidate` to call `getConstrainedConnectionCandidate` if the movement was constrained.

`getConstrainedConnectionCandidate` implements the following algorithm:
- Start a traversal at the saved `searchNode`.
- Choose traversal direction (next or previous) based on input direction.
- In a loop, get the next (or previous) connection node in the tree.
- If the connection is compatible with any of the connections on the dragging block/stack:
  - Create an appropriate candidate connection.
  - Update the search node.
  - Break out of the loop.
- Otherwise, continue until no new nodes are found and return null.

### Additional information

This does _not_ position the dragging block near the connection (tracked in https://github.com/google/blockly-keyboard-experimentation/issues/370). I'm still figuring out how to do that without causing very weird jumps--I have to push information backwards through the dragging pipeline.

This does _not_ loop from the last block in the workspace to the first, or vice versa (tracked in #369).

When you switch from unconstrained to constrained dragging, the traversal will always resume at the last valid connection that was tried. This may lead to unexpected jumps.

If you use the arrow keys to constrained move off the edge of the workspace, it will not scroll to keep it in view (tracked in #371). The drag will continue and end appropriately on an escape or enter.

I'm accessing a lot of private properties. Every `@ts-expect-error` is a sign of an API problem we need to sort out.

### DIFF
--- a/src/keyboard_drag_strategy.ts
+++ b/src/keyboard_drag_strategy.ts
@@ -4,11 +4,26 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {dragging, utils} from 'blockly';
+import {ASTNode, BlockSvg, RenderedConnection, dragging, utils} from 'blockly';
 import {Direction, getDirectionFromXY} from './drag_direction';
+import {LineCursor} from './line_cursor';
 
+// Copied in from core because it is not exported.
+interface ConnectionCandidate {
+  /** A connection on the dragging stack that is compatible with neighbour. */
+  local: RenderedConnection;
+
+  /** A nearby connection that is compatible with local. */
+  neighbour: RenderedConnection;
+
+  /** The distance between the local connection and the neighbour connection. */
+  distance: number;
+}
+
+// @ts-expect-error overrides a private function.
 export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
   private currentDragDirection: Direction | null = null;
+  private searchNode: ASTNode | null = null;
 
   override startDrag(e?: PointerEvent) {
     super.startDrag(e);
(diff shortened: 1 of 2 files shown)


src/keyboard_drag_strategy.ts
```diff
@@ -4,24 +4,137 @@
...
+        // @ts-expect-error connectionCandidate is private.
+        (this.connectionCandidate as ConnectionCandidate).neighbour,
+      );
+    }
+  }
+
+  /**
+   * Returns the next compatible connection in keyboard navigation order,
+   * based on the input direction.
+   * Always resumes the search at the last
```
maribethb: nit: this comment is cut off

src/keyboard_drag_strategy.ts
```diff
@@ -4,24 +4,137 @@
...
+   *
+   * @param draggingBlock The block where the drag started.
+   * @returns A valid connection candidate, or null if none was found.
+   */
+  private getConstrainedConnectionCandidate(
+    draggingBlock: BlockSvg,
+  ): ConnectionCandidate | null {
+    const cursor = draggingBlock.workspace.getCursor() as LineCursor;
```
maribethb, line 70: even when line cursor is the default in core, we'll still let people have other types right? can you leave a comment or something to remind us that we'll need to adjust this logic as well

src/keyboard_drag_strategy.ts
```diff
@@ -4,24 +4,137 @@
...
+        potential = cursor.getNextNode(potential, (node) => {
+          // @ts-expect-error isConnectionType is private.
+          return node && ASTNode.isConnectionType(node.getType());
+        });
+      }
+
+      localConns.forEach((conn: RenderedConnection) => {
+        const potentialLocation =
+          potential?.getLocation() as RenderedConnection;
+        if (connectionChecker.canConnect(conn, potentialLocation, true, 5000)) {
```
maribethb: is the 5000 here just a really big number? what about Infinity?

maribethb (approved):

src/keyboard_drag_strategy.ts
rachel-fenichel: Good point. I forgot Infinity exists. Changed.

src/keyboard_drag_strategy.ts
rachel-fenichel: Added a todo and filed an issue.

src/keyboard_drag_strategy.ts
rachel-fenichel: Fixed.

PR #390, opened 2025-04-07
feat: insert in move mode

microbit-matt-hillsdon:
This feels a bit rough while move mode disconnects from the connection you just inserted at.

@rachel-fenichel not ready to be merged but sharing in case it's useful for current work. Feel free to close.

An interesting issue is what to do with undo. At the moment the insert and the undo are separate but that feels wrong as the insert feels transient now until confirmed in position. Relatedly, undo during a move causes tragic errors (in common with most other non-move actions).

Demo: https://insert-move-mode.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -22,6 +22,7 @@ import type {
 
 import * as Constants from '../constants';
 import type {Navigation} from '../navigation';
+import {Mover} from './mover';
 
 const KeyCodes = BlocklyUtils.KeyCodes;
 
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):

rachel-fenichel:
If you'll mark it ready for review, I can merge this one.

microbit-matt-hillsdon:
@rachel-fenichel I merged main in and despite #392 you don't get the drag preview.

I think this breaks insertion as unless you move away and back everything lands on the workspace, so not ready to merge.

rachel-fenichel:
Superseded by #406

PR #399, opened 2025-04-07
fix: drag parent of shadow block

### LINKED ISSUE #395: Move starting with the cursor on a shadow block doesn't work
1. Move the cursor to a shadow block
2. Press M
3. Move with arrow keys, note jump to top left and weird/broken subsequently

https://github.com/user-attachments/assets/1369ee59-8a41-4538-ba9e-883a6ae0ec73

rachel-fenichel:
Fixes #395 

When starting a drag, walk up the tree until the first non-shadow block and drag it. If there's no non-shadow block, throw because that's not supposed to happen.

This matches [related code](https://github.com/google/blockly/blob/develop/core/dragging/block_drag_strategy.ts#L94) in core.

The core code can't handle this case because the drag strategy of the moving block needs to be patched. We're doing that in `mover.ts`, and getting the correct block in `mover.ts` means that we don't have to track patching/unpatching the drag strategy on multiple blocks.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -218,15 +218,27 @@ export class Mover {
   /**
    * Get the source block for the cursor location, or undefined if no
    * source block can be found.
+   * If the cursor is on a shadow block, walks up the tree until it finds
+   * a non-shadow block to drag.
    *
    * @param workspace The workspace to inspect for a cursor.
    * @returns The source block, or undefined if no appropriate block
    *     could be found.
    */
   protected getCurrentBlock(workspace: WorkspaceSvg): BlockSvg | undefined {
-    const cursor = workspace?.getCursor();
-    const curNode = cursor?.getCurNode();
-    return (curNode?.getSourceBlock() as BlockSvg) ?? undefined;
+    const curNode = workspace?.getCursor()?.getCurNode();
+    let block = curNode?.getSourceBlock();
+    if (!block) return undefined;
+    while (block && block.isShadow()) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@microbit-matt-hillsdon please review.

src/actions/mover.ts
```diff
@@ -218,15 +218,27 @@ export class Mover {
...
    *     could be found.
    */
   protected getCurrentBlock(workspace: WorkspaceSvg): BlockSvg | undefined {
-    const cursor = workspace?.getCursor();
-    const curNode = cursor?.getCurNode();
-    return (curNode?.getSourceBlock() as BlockSvg) ?? undefined;
+    const curNode = workspace?.getCursor()?.getCurNode();
+    let block = curNode?.getSourceBlock();
+    if (!block) return undefined;
+    while (block && block.isShadow()) {
```
microbit-matt-hillsdon, line 232: Nit: Because we check `!block` before and in the loop, this test can just be `block.isShadow()`. Makes it a bit clearer that block is always defined after the loop.

microbit-matt-hillsdon (commented):
Approved.

Deployment as tested: https://drag-shadow.blockly-keyboard-experimentation.pages.dev/

src/actions/mover.ts
rachel-fenichel: It can return `null` from `getParent`. It'll be an error (see below), but still possible in the type system.

PR #400, opened 2025-04-07
fix: patch WorkspaceSvg.getGesture to handle keyboard drags

### LINKED ISSUE #393: Mouse interactions cause odd behaviour during an in-progress keyboard move
1. Start a move with 'M'
2. Click on the moving block with the mouse
3. Note unexpected passive focus indicator and move no longer works

One option is to cancel a move if there's a mouse click.

The screenshot below shows what happens if after (2) I drag a block I didn't move.

Image

rachel-fenichel:
Fixes #393 

Adds an additional monkeypatch to the workspace. The current `Gesture` is always null during a keyboard drag, and trying to get a gesture also stops propagation and prevents default on the instigating event.

Without this patch, `WorkspaceSvg` automatically creates a new `Gesture` if none was found:

```
  getGesture(e: PointerEvent): Gesture | null {
    const isStart = e.type === 'pointerdown';

    const gesture = this.currentGesture_;
    if (gesture) {
      if (isStart && gesture.hasStarted()) {
        console.warn('Tried to start the same gesture twice.');
        // That's funny.  We must have missed a mouse up.
        // Cancel it, rather than try to retrieve all of the state we need.
        gesture.cancel();
        return null;
      }
      return gesture;
    }

    // No gesture existed on this workspace, but this looks like the start of a
    // new gesture.
    if (isStart) {
      this.currentGesture_ = new Gesture(e, this);
      return this.currentGesture_;
    }
    // No gesture existed and this event couldn't be the start of a new gesture.
    return null;
  }
```

The workspace then calls `handleWsStart` when the event propagates to the workspace, and `handleWsStart` calls `stopPropagation` and `preventDefault` on the event. 

Since `handlWsStart is not called in the new approach, directly cancel propagation in the getter.

### Additional information
As with anything marked as a monkeypatch, this will need formalization into core.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -141,7 +141,7 @@ export class Mover {
       new utils.Coordinate(0, 0),
     );
 
-    this.upatchWorkspace(workspace);
+    this.unpatchWorkspace(workspace);
     this.unpatchDragStrategy(info.block);
     this.moves.delete(workspace);
     return true;
@@ -172,7 +172,7 @@ export class Mover {
       new utils.Coordinate(0, 0),
     );
 
-    this.upatchWorkspace(workspace);
+    this.unpatchWorkspace(workspace);
     this.unpatchDragStrategy(info.block);
     this.moves.delete(workspace);
     return true;
@@ -263,7 +263,7 @@ export class Mover {
    *
    * @param workspace The workspace to unpatch.
    */
-  private upatchWorkspace(workspace: WorkspaceSvg) {
+  private unpatchWorkspace(workspace: WorkspaceSvg) {
     if (this.oldIsDragging) {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       (workspace as any).isDragging = this.oldIsDragging;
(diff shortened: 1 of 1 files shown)


src/actions/mover.ts
```diff
@@ -135,7 +141,7 @@ export class Mover {
       new utils.Coordinate(0, 0),
     );
 
-    this.unpatchIsDragging(workspace);
+    this.upatchWorkspace(workspace);
```
maribethb: typo: should be `unpatch` not `upatch`

maribethb (approved):

src/actions/mover.ts
rachel-fenichel: d'oh. fixed.

src/actions/mover.ts
```diff
@@ -43,6 +43,12 @@ export class Mover {
    */
   oldIsDragging: (() => boolean) | null = null;
 
+  /**
+   * The stashed getGesture function, which is replaced at the beginning
+   * of a keyboard drag and reset at the end of a keyboard drag.
+   */
+  oldGetGesture: ((e: PointerEvent) => Gesture | null) | null = null;
```
cpcallen, line 50: This should have been per-workspace (and patching/unpatching handled carefully) because `Mover` can handle multiple simultaneous moves on one or more workspaces.

PR #402, opened 2025-04-08
feat: loop line cursor

### LINKED ISSUE #401: shortcuts to jump between stacks of blocks
for the latest decision on implementation, see https://github.com/google/blockly-keyboard-experimentation/issues/401#issuecomment-2901905316

rest is provided as historical context

(This is largely @kmcnaught's text as an outcome from user testing with comments from @microbit-matt-hillsdon)

Users were not already familiar with tab navigation, but they learned quickly when shown. Using arrows to navigate in the toolbox was generally intuitive, with some caveats mentioned below. 

**However**, mode-switching between tab navigation and arrow-key navigation was a common source of confusion. When users were in the code workspace, they knew to switch to arrows, but as soon as arrow nav failed to do what they wanted, they switched to Tab instead. 

It was very common to end up in a situation where there were 2 top level blocks side by side, and this caused huge confusion when they wanted to move to the ‘other’ block. In the following screenshot, focus is on the leftmost stack, and you have to press **Left** to move to the right-most stack (!!) since it is positioned higher (Matt: or, _slightly_ more logically, up).

on start stack on left, with a longer forever stack on right, which extends above and below the on start block.

Possible mitigations:
1. Given current implementation, the navigation order should loop, when you get to the end of the AST, down/right should take you to the start. This would mitigate the blocker when people get to the end and think the arrow keys have stopped working.
    - Matt: There seems to be no downside to this so we'll propose a PR for this soon. 
2. Given current implementation, consider if there's any scope to change the ordering - should it be based on centre of mass rather than top left corner? I presume this has unacceptable consequences on order of operations (and it might be strange to have order of nav different to order of operation) I'm just including it for completeness. 
    - Matt: I assume this is not acceptable.  Adding blocks to a stack could change its centre of pass and therefore sequence in a weird way.
3. Add an extra shortcut key to jump between stacks. This would make it easier to navigate in larger projects, but only helps in the above simple case if you look up / learn the extra shortcut. 
    - Matt: I think we used to have this with m/n (m now taken for move).
4. Switch to using Tab to navigate _between_ stacks and arrows _within_ stacks. This would avoid the mental model mismatch of using 2d (arrow) navigation to move in a linear environment (list of stacks). It would also introduce many more tab stops in larger projects, and require skip links at the start and end of the workspace. It would mean you can't shift+tab to the toolbox to insert blocks (you'd lose your cursor location - and have further to go) so you would need to learn a shortcut to jump to toolbox.

microbit-grace:
Fixes #401 
Possible mitigation 1 in the issue

@rachel-fenichel Please feel free to not take this change and implement it differently. Not sure if there is a better way of implementing this.

### DIFF
--- a/src/line_cursor.ts
+++ b/src/line_cursor.ts
@@ -312,11 +312,13 @@ export class LineCursor extends Marker {
    * @param node The current position in the AST.
    * @param isValid A function true/false depending on whether the given node
    *     should be traversed.
+   * @param loop A boolean for whether to try to loop back to the top block.
    * @returns The next node in the traversal.
    */
   getNextNode(
     node: ASTNode | null,
     isValid: (p1: ASTNode | null) => boolean,
+    loop = true,
   ): ASTNode | null {
     if (!node) {
       return null;
@@ -332,6 +334,13 @@ export class LineCursor extends Marker {
       return siblingOrParent;
     } else if (siblingOrParent) {
       return this.getNextNode(siblingOrParent, isValid);
+    }
+    if (loop) {
+      // Loop back to first block if it exists.
+      const topBlocks = this.workspace.getTopBlocks(true);
+      return topBlocks.length > 0
+        ? Blockly.ASTNode.createTopNode(topBlocks[0])
+        : null;
     }
     return null;
   }
@@ -366,7 +375,34 @@ export class LineCursor extends Marker {
     } else if (newNode) {
       return this.getPreviousNode(newNode, isValid);
     }
-    return null;
+    // Loop back to last block if it exists.
+    return this.getLastNode();
+  }
+
+  /**
+   * Get the very last valid AST node.
+   *
+   * @returns The last AST node or null.
+   */
+  private getLastNode(): ASTNode | null {
+    const topBlocks = this.workspace.getTopBlocks(true);
+    if (topBlocks.length === 0) {
+      return null;
+    }
+    const lastTopBlockNode = Blockly.ASTNode.createTopNode(
+      topBlocks[topBlocks.length - 1],
+    );
+    let prevNode = lastTopBlockNode;
+    let nextNode: ASTNode | null = lastTopBlockNode;
+    while (nextNode) {
+      prevNode = nextNode;
+      nextNode = this.getNextNode(
+        prevNode,
+        this.validLineNode.bind(this),
+        false,
+      );
+    }
+    return prevNode;
   }
 
   /**


src/line_cursor.ts
```diff
@@ -312,11 +312,13 @@ export class LineCursor extends Marker {
    * @param node The current position in the AST.
    * @param isValid A function true/false depending on whether the given node
    *     should be traversed.
+   * @param loop A boolean for whether to try to loop back to the top block.
    * @returns The next node in the traversal.
    */
   getNextNode(
     node: ASTNode | null,
     isValid: (p1: ASTNode | null) => boolean,
+    loop = true,
```
rachel-fenichel, line 321: Please also add a loop param to `getPrevNode`

microbit-matt-hillsdon: FYI Grace won't be around until tomorrow so please feel free to take this and run with it.

We may need should pass loop to the recursive calls too.

microbit-matt-hillsdon:
This has given us looping during constrained move up (e.g. of draw circle in draw in the default demo). But if I move down then it breaks:

```
b.getSourceBlock is not a function
TypeError: b.getSourceBlock is not a function
    at ConnectionChecker$$module$build$src$core$connection_checker.doSafetyChecks (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1233:272)
    at ConnectionChecker$$module$build$src$core$connection_checker.canConnectWithReason (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1230:658)
    at ConnectionChecker$$module$build$src$core$connection_checker.canConnect (webpack-internal:///./node_modules/blockly/blockly_compressed.js:1230:526)
    at eval (webpack-internal:///./src/keyboard_drag_strategy.ts:88:39)
    at Array.forEach (<anonymous>)
    at KeyboardDragStrategy.getConstrainedConnectionCandidate (webpack-internal:///./src/keyboard_drag_strategy.ts:86:24)
    at KeyboardDragStrategy.getConnectionCandidate (webpack-internal:///./src/keyboard_drag_strategy.ts:111:25)
    at KeyboardDragStrategy.updateConnectionPreview (webpack-internal:///./node_modules/blockly/blockly_compressed.js:914:495)
    at KeyboardDragStrategy.drag (webpack-internal:///./node_modules/blockly/blockly_compressed.js:914:307)
    at KeyboardDragStrategy.drag (webpack-internal:///./src/keyboard_drag_strategy.ts:36:15)
```

microbit-matt-hillsdon:
This code has now moved into core so I think someone else will need to take #401 and implement looping in core's version. Suggest closing.

rachel-fenichel:
Agreed. I'm implementing it in core.

Issue #346, opened 2025-03-27
Delete `LineCursor` in favor of core implementation


rachel-fenichel:
Depends on https://github.com/google/blockly/issues/8822

Once that's merged, delete `LineCursor` in keyboard-experimentation.

PR #405, opened 2025-04-08
refactor: Update to v12.0.0-beta.3 and use cursor from core.

gonfunko:
This PR updates the Blockly dependency to v12.0.0-beta.3, and removes the `LineCursor` class in this plugin in favor of the backported version now in core as of beta 3. It also includes minor compatibility fixes for incidental changes in beta 3.


gonfunko:
This fixes #346.

maribethb (approved):
Sometimes the code uses `LineCursor` and sometimes it uses `Blockly.LineCursor`. I think this problem is prevalent outside this PR as well though, and the former is more common even though I'm 99% sure the latter is more correct. I don't think you necessarily need to fix it in this PR since we're already mixing the two styles, but I do wonder if they actually refer to the same object (a la 2 blockly problem) or if that matters

BenHenning (approved):
Thanks @gonfunko.

For future searching that comes to this PR: the backport PR was https://github.com/google/blockly/pull/8834.

Issue #398, opened 2025-04-07
Choose which blocks to drag based on modifier key


rachel-fenichel:
Once https://github.com/google/blockly/issues/8870 is fixed in core (and released in a beta for use), override `shouldHealStack` to return `false` if shift was pressed and `true` otherwise.

This corresponds to a single-block drag with `m` (default) and a multi-block drag with `M`.

As a first pass, this will require passing additional information into the drag strategy through side channels in the event, as I did with the direction of movement.

Later, the information should be passed in directly. I think that we may change the dragging code to accept a keyboard event | pointer event instead of just a pointer event.

microbit-matt-hillsdon:
We should make a corresponding change to the disconnect operation at the same time for consistency (it was updated to be consistent with move on #542).

gonfunko:
This was addressed in https://github.com/RaspberryPiFoundation/blockly/pull/9630.

PR #406, opened 2025-04-08
feat: start a move when inserting a block from the flyout

rachel-fenichel:
Works on https://github.com/google/blockly-keyboard-experimentation/issues/398 by switching to _always_ healing the stack (only dragging the selected block).

This is a better default while we wire up the actual behaviour.

Replaces https://github.com/google/blockly-keyboard-experimentation/pull/390

### Additional information

[Lines 225-248](https://github.com/google/blockly-keyboard-experimentation/compare/main...rachel-fenichel:blockly-keyboard-experimentation:insert_move?expand=1#diff-83090943936e3fd0aeacc7b73d1d244a2d87f974f8f6b649b9a3226033f04b0fR225-R248) in `keyboard_drag_strategy.ts` are essentially a copy of [code from `block_drag_strategy.ts`](https://github.com/google/blockly/blob/develop/core/dragging/block_drag_strategy.ts#L245-L267), but I had to rearrange the code to avoid non-null assertions.

Proposed followup: 
- Extract the logic out of `updateConnectionPreviewer` to a helper function in core.
- Restructure the code in core to avoid non-null assertions.
- Call the new helper in the subclass (instead of reimplementing).

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -22,6 +22,7 @@ import type {
 
 import * as Constants from '../constants';
 import type {Navigation} from '../navigation';
+import {Mover} from './mover';
 
 const KeyCodes = BlocklyUtils.KeyCodes;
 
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
@microbit-matt-hillsdon fyi

gonfunko (approved):

PR #408, opened 2025-04-09
Share shortcut formatting between help/menu/messages

microbit-matt-hillsdon:
Update "toast" message on enter to reference help which we think is a better option based on user testing.

Partly motivated by wanting to rebind list_shortcuts in MakeCode and have the correct shortcut be shown by the toast/dialog.

This also consistently avoids using + between shortcuts on Mac.

To test on Windows on Mac you can do:

```
<script>
      Object.defineProperty(navigator, 'platform', {
        value: 'Windows',
      });
</script>
```

near the top of the HTML file (needs to be early). Any value that doesn't start Mac is good.

### DIFF
--- a/src/shortcut_formatting.ts
+++ b/src/shortcut_formatting.ts
@@ -7,18 +7,31 @@ const isMacPlatform = navigator.platform.startsWith('Mac');
  * Format the primary shortcut for this platform in a user facing format.
  *
  * @param action The action name, e.g. "cut".
+ * @param format The key format.
  * @returns The formatted shortcut.
  */
-export function formatActionShortcut(action: string): string {
-  const parts = actionShortcutsForPlatform(action)[0];
-  return parts.join(' + ');
+export function formatActionShortcut(
+  action: string,
+  format: ShortcutFormat,
+): string {
+  const parts = actionShortcutsForPlatform(action, format)[0];
+  return parts.join(isMacPlatform ? ' ' : ' + ');
 }
 
-const modifierNames: Record<string, string> = {
-  'Control': 'Ctrl',
-  'Meta': '⌘ Command',
-  'Alt': isMacPlatform ? '⌥ Option' : 'Alt',
+const modifierNamesByFormat: Record<ShortcutFormat, Record<string, string>> = {
+  long: {
+    'Control': 'Ctrl',
+    'Meta': '⌘ Command',
+    'Alt': isMacPlatform ? '⌥ Option' : 'Alt',
+  },
+  short: {
+    'Control': 'Ctrl',
+    'Meta': '⌘',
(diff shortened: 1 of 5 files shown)


microbit-matt-hillsdon:
Back to draft for a mo to keep concise form on the context menu.

src/shortcut_formatting.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Keys are returned in a user facing format.
+ *
+ * This could be considerably simpler if we only bound shortcuts relevant to the
+ * current platform or tagged them with a platform.
+ *
+ * @param action The action name, e.g. "cut".
+ * @param format The key format.
+ * @returns The formatted shortcuts.
+ */
+export function actionShortcutsForPlatform(
```
maribethb: generally I'd prefer having a `getShortActionShortcut` and `getLongActionShortcut` instead of making the user pass a magic string, if there's only two options.

maribethb (approved):
Something to think about (no action needed for this PR) is how localizable this is. Getting the actual key out of the registry is a good first step, but I'm curious if anything will need to change e.g. for "ctrl" in other keyboard layouts? I don't know how non-US keyboards handle this kind of thing.

For the basic label I've filed #412 because all of the bare strings will need to be moved into Blockly.Msg. As long as all languages are cool with the order of "name of shortcut" followed by "keypresses" then it should be fine to just concatenate the two values we got out of Blockly.Msg. But if some languages would have them in different order then we'll need to research and design around that in the future.

src/shortcut_formatting.ts
maribethb: or potentially having one method with a boolean to switch to short mode or something like that

microbit-matt-hillsdon: Slightly complicated by there being short vs long but also individual keys vs joined string. Gone with API that looks nice for the callers in 6f692d428ee3400dd56c498a6db0ae876b5ceeb6 as in fact we only use short + joined string (for context menu) and long + individual keys (for help dialog).

maribethb (approved):

microbit-matt-hillsdon:
> Something to think about (no action needed for this PR) is how localizable this is. Getting the actual key out of the registry is a good first step, but I'm curious if anything will need to change e.g. for "ctrl" in other keyboard layouts? I don't know how non-US keyboards handle this kind of thing.

Coincidentally I just read a translation-related issue in the micro:bit Python Editor where someone requests localization for "Ctrl+Shift+Space", so this is a thing some folks want in a related context:

> Hello, in English the shortcut is Ctrl+Shift+Space which can't be kept in German as the German keyboard has a localized version of Ctrl (Strg), Shift (Umschalt) and Space ("Leertaste"). Can this shortcut be translated somehow?

PR #409, opened 2025-04-09
feat: add loopback when finding a valid node during a keyboard drag

### LINKED ISSUE #369: Add helpers to loop around from the last block of the workspace to the first (and vice versa) when doing constrained movement.

### LINKED ISSUE #396: Remove or improve workspace positions in constrained move
When doing a constrained move there seem to be two workspace positions at the beginning and end of the sequence. I didn't look at the code to write this, so this is my impression as a user.

They are not really very clearly differentiated from the equivalent constrained positions.

I think the constrained navigation should just loop around the constrained positions and you can use option/ctrl+arrow at any point to move away. Alternatively, have one workspace position that's in free space, similar to initial block placement. I lean towards the former to discourage creating disabled blocks, but that might be a MakeCode-specific view.

### LINKED ISSUE #410: Add tests for cursor loopback during drag
Test cases:
- multiple c-shape stacks
- multiple stacks of normal stack blocks
- an if-else block inside a c-shape block
- no blocks on workspace
- only value blocks on workspace
- one stack on workspace

Each case should be tested while dragging:
- a stack block (e.g. print)
- a c-shaped block (e.g. repeat)
- an e-shaped block (e.g. if/else)
- a value block (e.g. number block)

Test sequence:
- load initial blocks
- navigate to a location on a block (if applicable)
- i to insert a block
- arrow keys + enter to select the appropriate block from the toolbox and begin a drag
- arrow keys + assertions to check that the cursor visits every valid location

rachel-fenichel:
- Fixes #369 
- Fixes #396 

### Proposed followup
- [ ] Add tests for loopback search (#410)
- [ ] Move loopback code to core
- [ ] Use loopback code for navigation as well as for dragging

### Additional information
The first commit handles loopback; the second commit cleans up some lint.

### DIFF
--- a/src/keyboard_drag_strategy.ts
+++ b/src/keyboard_drag_strategy.ts
@@ -144,15 +144,20 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
     const cursor = draggingBlock.workspace.getCursor() as LineCursor;
     if (!cursor) return null;
 
+    // Helper function for traversal.
+    function isConnection(node: ASTNode | null): boolean {
+      return !!node && node.isConnection();
+    }
+
     const connectionChecker = draggingBlock.workspace.connectionChecker;
     let candidateConnection: ConnectionCandidate | null = null;
     let potential: ASTNode | null = this.searchNode;
     const dir = this.currentDragDirection;
     while (potential && !candidateConnection) {
       if (dir === Direction.Up || dir === Direction.Left) {
-        potential = this.getPreviousNode(potential, cursor);
+        potential = cursor.getPreviousNode(potential, isConnection, true);
       } else if (dir === Direction.Down || dir === Direction.Right) {
-        potential = this.getNextNode(potential, cursor);
+        potential = cursor.getNextNode(potential, isConnection, true);
       }
 
       localConns.forEach((conn: RenderedConnection) => {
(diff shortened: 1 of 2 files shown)


src/keyboard_drag_strategy.ts
```diff
@@ -289,4 +283,77 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
...
+  /**
+   * Get the previous node in the tree, with loopback to the last
+   * stack on the workspace if needed.
+   *
+   * @param start Where to start traversal.
+   * @param cursor The workspace's cursor
+   * @returns The previous node, or null if there were no valid nodes
+   *     on the workspace.
+   */
+  private getPreviousNode(start: ASTNode, cursor: LineCursor) {
```
maribethb: I guess it's not required by either ts or our current linting rules, but I think it's good to add return types here so it's obvious it might return null.

src/keyboard_drag_strategy.ts
```diff
@@ -289,4 +283,77 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
...
+      // Find the last stack.
+      const lastTopBlockNode = ASTNode.createStackNode(
+        topBlocks[topBlocks.length - 1],
+      );
+      let prevNode = lastTopBlockNode;
+      let nextNode: ASTNode | null = lastTopBlockNode;
+      // Iterate until you fall off the end of the stack.
+      while (nextNode) {
+        prevNode = nextNode;
+        nextNode = cursor.getNextNode(prevNode, (node) => {
```
maribethb: I could be wrong here, but if the very last thing in this stack is a compatible connection, would this code back up too far? On this line prevNode would be set to the connection node, then when you resume searching, you'd back up an extra step?

It seems like you should have an extra step to check if the prevNode is a compatible connection already, if it is return, else continue searching. But I could be misunderstanding something. otherwise, same applies to the `getNextNode` logic.

src/keyboard_drag_strategy.ts
```diff
@@ -289,4 +283,77 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
...
+  private getNextNode(start: ASTNode, cursor: LineCursor) {
+    let potential: ASTNode | null = start;
+    potential = cursor.getNextNode(potential, (node) => {
+      // @ts-expect-error isConnectionType is private.
+      return node && ASTNode.isConnectionType(node.getType());
+    });
+    if (!potential) {
+      // Loop back to first block if it exists.
+      // @ts-expect-error workspace is private
+      const topBlocks = this.workspace.getTopBlocks(true);
```
maribethb: Last we talked, I thought you decided to have a `getFirstNode` method on the cursor. did you change your mind and if so why?

I think this would be slightly clearer if getting the first node was a separate function, then the logic about "if the first node is compatible return it" would be more obvious imo.

rachel-fenichel:
Build fails on this because it relies on a paired change in core (https://github.com/google/blockly/pull/8883) and was developed by linking against core. I have checked that the tests pass and done some basic manual testing.

maribethb (approved):

Issue #362, opened 2025-04-01
Update action menu items to use new context menu api features


maribethb:
Once the work for the context menu changes is done in core, they can be used by the plugin. This involves removing hacks to get the location data for the context menu, and being able to hide items if the menu is opened via mouse vs keyboard.

PR #423, opened 2025-04-11
feat: show context menu on workspace

maribethb:
Works on #362 

Updates the context menu keyboard action to use new bits of the API in core. notably, don't have to calculate a location for blocks because core does that now.

Also, show the context menu on the workspace since core can support doing that with a keyboard event now.

This PR is paired with https://github.com/google/blockly/pull/8877 and you'll need to link blockly to a version with that PR submitted in order for the exp to load once both of these are submitted.

### DIFF
--- a/src/actions/action_menu.ts
+++ b/src/actions/action_menu.ts
@@ -92,7 +92,7 @@ export class ActionMenu {
   private openActionMenu(workspace: WorkspaceSvg): boolean {
     let rtl: boolean;
 
-    // TODO: Pass this through the precondition and callback instead of making it up.
+    // TODO(#362): Pass this through the precondition and callback instead of making it up.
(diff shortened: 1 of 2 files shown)


src/actions/action_menu.ts
```diff
@@ -90,12 +90,11 @@ export class ActionMenu {
    * @param workspace The workspace.
    */
   private openActionMenu(workspace: WorkspaceSvg): boolean {
-    let menuOptions: Array<
-      | ContextMenuRegistry.ContextMenuOption
-      | ContextMenuRegistry.LegacyContextMenuOption
-    > = [];
     let rtl: boolean;
 
+    // TODO: Pass this through the precondition and callback instead of making it up.
```
rachel-fenichel: File an issue and link

rachel-fenichel (approved):

PR #425, opened 2025-04-11
fix: Disallow toolbox/insert during a keyboard-driven move

### LINKED ISSUE #417: Opening the toolbox during keyboard move
This leaves the flyout open, but the active cursor remains on the block in move mode. The arrow keys continue to move the block and the flyout cannot be navigated. 

Image

microbit-robert:
This PR fixes https://github.com/google/blockly-keyboard-experimentation/issues/417

Demo: https://toolbox-dragging.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/actions/insert.ts
+++ b/src/actions/insert.ts
@@ -53,8 +53,7 @@ export class InsertAction {
   private registerShortcut() {
     const insertShortcut: ShortcutRegistry.KeyboardShortcut = {
       name: this.insertShortcutName,
-      preconditionFn: (workspace: WorkspaceSvg) =>
-        !workspace.isDragging() && this.insertPrecondition(workspace),
+      preconditionFn: this.insertPrecondition.bind(this),
       callback: this.insertCallback.bind(this),
       keyCodes: [KeyCodes.I],
     };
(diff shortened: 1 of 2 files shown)


src/actions/insert.ts
```diff
@@ -53,7 +53,8 @@ export class InsertAction {
   private registerShortcut() {
     const insertShortcut: ShortcutRegistry.KeyboardShortcut = {
       name: this.insertShortcutName,
-      preconditionFn: this.insertPrecondition.bind(this),
+      preconditionFn: (workspace: WorkspaceSvg) =>
+        !workspace.isDragging() && this.insertPrecondition(workspace),
```
maribethb: I think this should go inside the `insertPrecondition` function instead of splitting up the logic here and in that named function. otherwise lgtm

maribethb (approved):

src/actions/insert.ts
microbit-robert: Yep, that makes sense. Updated here [833e67b](https://github.com/google/blockly-keyboard-experimentation/pull/425/commits/833e67b3fad580fa6ec4e0ab5689a7ba26c23635)

maribethb (approved):

Issue #418, opened 2025-04-10
Unexpected insert placement when value block selected


microbit-robert:
See screen grab below. In this case, it would make more sense for the `if` block to be initially positioned below the `draw` block rather than elsewhere on the workspace.

![Image](https://github.com/user-attachments/assets/06005b75-337b-48bc-bae7-67441323b51c)

PR #426, opened 2025-04-11
fix: heuristic insert for statment block when value block is selected

microbit-robert:
Potential fix for https://github.com/google/blockly-keyboard-experimentation/issues/418.  I'm not at all convinced that this is the right approach as the first warning message we get is `This block can not be inserted at the marked location` from inside the `insertBlock` method. However, the behaviour demonstrated here is what I think we're after.

Please feel free to close this in favour of a better approach or more reasoned approach.

Demo: https://insert-at-value.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -715,7 +715,26 @@ export class Navigation {
 
       // 3. Output connection. This will wrap around or displace.
       if (stationaryBlock.outputConnection) {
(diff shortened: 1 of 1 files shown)


microbit-robert:
We're currently working to improve this PR, so converting to draft for now.

microbit-robert:
Closing in favour of https://github.com/google/blockly-keyboard-experimentation/pull/432

PR #431, opened 2025-04-12
fix: Disallow delete during a keyboard-driven move

### LINKED ISSUE #413: Delete during keyboard move
You can delete blocks during keyboard move and it doesn't end well.

An example of the "other shortcuts" part of #394 which has been closed.

1. Select statement block
2. Press M then Delete

Image

microbit-robert:
This PR fixes https://github.com/google/blockly-keyboard-experimentation/issues/413. The existing delete shortcut was never unregistered during the install of this action, so even with the new dragging check in place, the existing delete shortcut was still succeeding. This has been ammended by unregistering the existing delete shortcut in the same way as the existing context menu item.

Demo: https://delete-dragging.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/actions/delete.ts
+++ b/src/actions/delete.ts
@@ -29,6 +29,12 @@ export class DeleteAction {
   private oldContextMenuItem: ContextMenuRegistry.RegistryItem | null = null;
 
   /**
+   * Saved delete shortcut, which is re-registered when this action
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
Nice catch and an easy solution!

PR #432, opened 2025-04-14
feat: improve heuristic insert for statement blocks

### LINKED ISSUE #418: Unexpected insert placement when value block selected

microbit-matt-hillsdon:
Move up to the block if we try to insert a statement block in a value input, value connection, field or value block. This has a similar feel to the heuristic that uses the value input on a block if you insert a value block.

Fixes #418 

Demo: https://statement-vs-value-heuristic.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -675,11 +675,33 @@ export class Navigation {
     const stationaryType = stationaryNode.getType();
     const stationaryLoc = stationaryNode.getLocation();
 
-    if (stationaryNode.isConnection()) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
LGTM. Worth adding tests for these specific cases so we don't break them going forward.

PR #437, opened 2025-04-15
feat: auto confirm a move on subsequent action

### LINKED ISSUE #434: Can trigger context menu in keyboard move mode
You can trigger the context menu via e.g. Cmd+Enter in move mode (but not right-click).

This seems inadvisable in and of itself and likely any mutations you can trigger from the menu are buggy.

microbit-matt-hillsdon:
This is a hack to get a behaviour we like like. It's not clear what the right API for this would be.

For discussion later today.

Fixes #434 context menu during move as well as similar bugs such as clean up durning move, cut during move, likely others.

Nice API could be a workspace-wide precondition that applies to all actions that was invoked with the name of the intended shortcut.

Demo: https://auto-confirm-move.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/actions/move.ts
+++ b/src/actions/move.ts
@@ -22,6 +22,13 @@ const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
  * Actions for moving blocks with keyboard shortcuts.
  */
 export class MoveActions {
+  /**
+   * Stored to enable us to restore monkey patch.
+   */
+  private oldShortcutRegistryOnKeyDown:
+    | typeof ShortcutRegistry.registry.onKeyDown
+    | null = null;
+
   constructor(private mover: Mover) {}
 
   private shortcuts: ShortcutRegistry.KeyboardShortcut[] = [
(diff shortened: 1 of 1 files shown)


microbit-robert:
We've now merged a version of this into our user testing branch in our fork of the project that ignores all non-move-related actions while a move is in progress instead of finishing the move and running the non-move-related action following discussion yesterday.

It's still not clear the API for this should look like in core. Perhaps both the workspace and the actions have a 'normal' mode and 'move' mode where only the actions with a mode that matches that of the workspace are allowed through?

microbit-matt-hillsdon:
See also the discussion from [this comment](https://github.com/google/blockly-keyboard-experimentation/issues/444#issuecomment-2821932123) re clicking outside of the workspace.

microbit-matt-hillsdon:
What do folks think re this behaviour? @rachel-fenichel @maribethb

microbit-matt-hillsdon:
Closing in favour of #560

PR #438, opened 2025-04-15
feat: avoid insert then move; delete on abort

### LINKED ISSUE #421: Unexpected 'pop out' of conditional / logic block
If you have an `if-else-if` block:

- the first insert of a condition goes into the `if` statement by default
- "Enter" to accept this position / end move mode
- Select the if-else-if` block again
- insert a second condition
- this second condition also moves to the `if` statement (maybe there's an argument that it should move to the next free slot, but that's for a separate discussion later)
- move the second condition down into the free slot in the `else-if` statement
- "Enter" to accept
- the first condition in the `if` statement is unexpectedly popped out on to the workspace

https://github.com/user-attachments/assets/8bcfdc99-14da-498b-ae1e-79eb011900dd

microbit-matt-hillsdon:
Fixes #421 

Demo: https://insert-connection.blockly-keyboard-experimentation.pages.dev/

Problems with this PR:
- [x] Select create canvas block, insert statement block, escape to abort. We delete more than we intend.
- [ ] Not doesn't wrap around true in test of `if` by default if on `if`

https://github.com/user-attachments/assets/812c993d-bc04-4fbc-b2e9-2a81d1f078ae

To raise as separate issues:
- Post-abort cursor position isn't great (like delete in general).
- Undo has too many steps.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -132,8 +132,8 @@ export class EnterAction {
     const newBlock = this.createNewBlock(workspace);
     if (!newBlock) return;
 
-    const startConnection = stationaryNode
-      ? this.navigation.findBestInsertionConnection(stationaryNode, newBlock)
+    const insertStartPoint = stationaryNode
+      ? this.navigation.findInsertStartPoint(stationaryNode, newBlock)
       : null;
     if (workspace.getTopBlocks().includes(newBlock)) {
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
Closing in favour of #455, thanks @rachel-fenichel

PR #451, opened 2025-04-17
feat: scroll moving block into view

microbit-grace:
Suggested fix for #371 

Would welcome an alternative approach to using  `setTimeout`. 
Not sure about the expected behaviour, but is perhaps a step forward.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -145,7 +145,7 @@ export class Mover {
     this.unpatchDragStrategy(info.block);
     this.moves.delete(workspace);
     // Delay scroll until after block has finished moving.
-    setTimeout(() => this.scrollCurrentBlockIntoView(workspace), 0)
+    setTimeout(() => this.scrollCurrentBlockIntoView(workspace), 0);
     return true;
   }
 
@@ -178,7 +178,7 @@ export class Mover {
     this.unpatchDragStrategy(info.block);
     this.moves.delete(workspace);
     // Delay scroll until after block has finished moving.
-    setTimeout(() => this.scrollCurrentBlockIntoView(workspace), 0)
+    setTimeout(() => this.scrollCurrentBlockIntoView(workspace), 0);
     return true;
   }
 
@@ -315,7 +315,7 @@ export class Mover {
   }
 
   /**
-   * Scrolls the current block into view if exists one.
+   * Scrolls the current block into view if one exists.
    *
    * @param workspace The workspace to get current block from.
    */
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -55,12 +55,6 @@ export class Mover {
    */
   private oldDragStrategy: IDragStrategy | null = null;
 
-  /**
-   * The current block being moved, which is set at the start of a move and
-   * reset at the end of a move.
-   */
-  private currentBlock: BlockSvg | null = null;
-
   constructor(protected navigation: Navigation) {}
 
   /**
@@ -107,7 +101,6 @@ export class Mover {
     const cursor = workspace?.getCursor();
     const block = this.getCurrentBlock(workspace);
     if (!cursor || !block) throw new Error('precondition failure');
-    this.currentBlock = block;
 
     // Select and focus block.
     common.setSelected(block);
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
For info, we've already merged this to our integration branch for for user testing so no urgency to review.

src/actions/mover.ts
```diff
@@ -307,6 +313,20 @@ export class Mover {
       this.oldDragStrategy = null;
     }
   }
+
+  /**
+   * Scrolls the current block into view if exists one.
```
RoboErikG: Nit: "if exists one" should be "if one exists"

src/actions/mover.ts
```diff
@@ -307,6 +313,20 @@ export class Mover {
...
+   *
+   * @param workspace The workspace to get current block from.
+   */
+  private scrollCurrentBlockIntoView(workspace: WorkspaceSvg) {
```
RoboErikG: What are your thoughts on including an extra padding parameter in this method that defaults to 0?

I noticed when using constrained movement I wanted the scroll to make a bit more of the surrounding context visible to understand where I'm moving the blocks to. One way to do that would be to give the bounds extra padding when calling this from the constrained move method.

RoboErikG (commented):
Overall looks really good. Couple comments.

src/actions/mover.ts
microbit-grace: Makes sense, have updated the doc in [2bc7d7c](https://github.com/google/blockly-keyboard-experimentation/pull/451/commits/2bc7d7c3be965aca0f6a5f2285d5f854b89ffde1)

src/actions/mover.ts
microbit-grace: I don't see any downsides to adding an extra padding parameter. I have added it here: [6c8b345](https://github.com/google/blockly-keyboard-experimentation/pull/451/commits/6c8b3459f2169190ad0fa1716eead715695158fe). 

Passing a larger padding parameter does not seem to make the surrounding content more visible when I try it locally, will investigate why the padding does not take effect before requesting re-review

RoboErikG: Sounds good. I also just double checked the workspace.scrollBoundsIntoView method and realized the default there is 10, so you may also want to set the default here to 10 as well.

microbit-grace: Good point, I have corrected the default padding to 10 here: [140be72](https://github.com/google/blockly-keyboard-experimentation/pull/451/commits/140be725c15fd87ddc4655d30848eeff85c4e7f8)

Re why padding was not taking effect, turns out `scrollBoundsIntoView` was unexpectedly being called twice in `scrollCurrentBlockIntoView`. This was because `getCurrentBlock` eventually calls `scrollBoundsIntoView` in the call stack. I have suggested a fix for it here: [9657acd](https://github.com/google/blockly-keyboard-experimentation/pull/451/commits/9657acd0b8f61bbcb953fb2639712d65504cad26)

Re adding extra padding for constrained move, I have set the padding as 100 to ensure the block above or below are within view here: [420c224](https://github.com/google/blockly-keyboard-experimentation/pull/451/commits/420c2248fc124666a694c254682f026cc24b7d0d)

src/actions/mover.ts
```diff
@@ -315,15 +332,17 @@ export class Mover {
...
-   * @param workspace The workspace to get current block from.
+   * @param workspace The workspace to scroll the given bounds into view in.
+   * @param padding Amount of spacing to put between the bounds and the edge of
+   *     the workspace's viewport.
    */
-  private scrollCurrentBlockIntoView(workspace: WorkspaceSvg) {
-    const blockToView = this.getCurrentBlock(workspace);
-    if (blockToView) {
+  private scrollCurrentBlockIntoView(workspace: WorkspaceSvg, padding = 10) {
+    if (this.currentBlock) {
```
RoboErikG: I think this will always be null when called from setTimeout since you set this.currentBlock to null immediately after setting a timeout to call this method, so the assignment to null happens before the timeout function is called.

Your previous version where you got the current block in this method should have worked. What was the reason for storing it in a property?

RoboErikG: Ah, I'm seeing it now. Since it won't scroll if the block is already visible regardless of the extra padding the second call with a larger padding gets ignored, but this is equivalent to just removing all of the calls with setTimeout.

For now, can you revert the change to avoid multiple calls and we can merge it without the extra padding working? I think this will need a change to scrollBoundsIntoView to optionally require padding to be included.

RoboErikG: I've created https://github.com/google/blockly/pull/8925 to fix the issue with not scrolling when you increase the padding.

microbit-grace: Sounds good, I have reverted the recent changes which adds a padding for constrained move and avoid multiple calls

RoboErikG (approved):
Thanks for working through it with me!

Issue #439, opened 2025-04-15
Disconnecting through unconstrained movement takes multiple steps.


rachel-fenichel:
### Repro steps
- Load the basic test blocks
- Navigate to a stack block
- Press `m` to enter movement mode
- Notice that the block offsets slightly down and to the right
![Image](https://github.com/user-attachments/assets/14329379-8019-47d1-8f77-7b001aa87f0d)
- Press `ctrl + Right Arrow` to do an unconstrained move to the right.
- Notice that the block position has not changed
- Press `ctrl + Right Arrow` again
- Notice that the block has now moved to be disconnected
![Image](https://github.com/user-attachments/assets/3824da24-bdcf-47b2-8c87-a8d82c87ad05)

This issue happens inconsistently: when moving the block from its initial location, it always happens. When moving the block from a different preview location, the issue occurs when moving to the left but not the right.

### Analysis

There are several issues here. One is that `moveUnconstrained` does not call `info.updateTotalDelta` after calling `onDrag`, so the block's position and its record of total drag movement can get out of sync.

But if we fixed that by just calling `info.updateTotalDelta`, the block would be unable to move, because after every keypress it would snap back to its location slightly down and to the right of the candidate connection.

Another fix is to increase `UNCONSTRAINED_MOVE_DISTANCE`. This would mask but not resolve the problem.

To actually fix this, the keyboard dragger and/or mover code needs to check if it's an unconstrained move away from a candidate connection, move an appropriate distance to avoid the connection, and update the total delta appropriately.

microbit-matt-hillsdon:
Still an issue after https://github.com/google/blockly-keyboard-experimentation/pull/454

microbit-matt-hillsdon:
I think we can close this one based on the linked PR.

PR #454, opened 2025-04-18
fix: show correct preview during unconstrained movement near a connection

rachel-fenichel:
Proposed fix for https://github.com/google/blockly-keyboard-experimentation/issues/439

@microbit-robert @microbit-matt-hillsdon I'd appreciate your opinion on this.

Two parts to the change:

### Initial offset
In `mover.ts`, call `info.updateTotalDelta` immediately after `onDragStart`. 
If the drag immediately shows a preview, this will update the block's location accordingly. 
If a preview is shown, the block is moved down and to the right of the target connection. 
image

#### Rationale
Without this change, the block was wrong about its position until the first constrained movement occurred. I think this change is necessary no matter what.

### Offset during drag
In `keyboard_drag_strategy.ts`, only force the block's position offset if the most recent movement was constrained.
image

If the movement was unconstrained, the insertion marker still shows or hides as appropriate, but the dragging block can translate relative to the insertion marker. All of these become legal:
image
image
image
image

Moving away more causes the insertion marker to disappear: 
image

And here's an example as a gif. All movements in this gif are unconstrained.
582f55d7-5595-49db-a068-621fd105c8c7

#### Rationale
In discussion with Robert and Matt I suggested making the drag step larger during keyboard drags, or making it larger only when showing a preview and trying to do an uncosntrained movement. But as the drag step size increases, so does the chance of missing a connection entirely. This is especially acute when working with value inputs, where the connections may be very close to each other in a row.

While investigating, I realized that part of the issue is that the user has no feedback: there is no visible change on the first unconstrained movement away from a connection (or, in MakeCode, the first four). In mouse mode the feedback is that the block moves along with the mouse pointer. The insertion marker sticks at the new position, but the dragging block moves around. In fact, this is half of the reason for using a separate insertion marker as a preview.

This change mimics mouse dragging by moving the dragging block around while still showing the preview. It may take multiple keypresses to move away from a connection, but users get feedback that a movement is happening.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -121,6 +121,7 @@ export class Mover {
     this.moves.set(workspace, info);
     // Begin drag.
     dragger.onDragStart(info.fakePointerEvent('pointerdown'));
+    info.updateTotalDelta();
     return true;
   }
 
--- a/src/keyboard_drag_strategy.ts
+++ b/src/keyboard_drag_strategy.ts
@@ -60,12 +60,14 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
       // The next constrained move will resume the search from the current
       // candidate location.
       this.searchNode = ASTNode.createConnectionNode(neighbour);
-      // The moving block will be positioned slightly down and to the
-      // right of the connection it found.
-      // @ts-expect-error block is private.
-      this.block.moveDuringDrag(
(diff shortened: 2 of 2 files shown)


gonfunko (approved):

microbit-matt-hillsdon:
I think this works really well, thank you.

PR #455, opened 2025-04-19
feat: improved insertion flow

### LINKED ISSUE #421: Unexpected 'pop out' of conditional / logic block

rachel-fenichel:
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/421

Based on https://github.com/google/blockly-keyboard-experimentation/pull/438

This is intended to be reviewed one commit at a time; each commit is small.

### Behaviour after change
- Connecting a block to the target location happens after the move starts.
  - This avoids the problem identified in #421, where blocks were unexpectedly getting popped out onto the workspace at the start of drags.
  - Aborting a drag with `Escape` deletes the newly inserted block.
  - Aborting a drag with `Escape` puts the cursor at the last target connection, (if any).
  - Undo after insertion deletes the inserted block
  - Inserting around a value block happens correctly

### Additional information

This is largely the change that @microbit-matt-hillsdon made, rebased onto main.

I had a grand plan to do all the work to find the `insertStartPoint` in `KeyboardDragStrategy` instead of having to pass it into the drag strategy. That's the reason that I started by passing the block and `isNewBlock` into the drag strategy. This idea failed because `getStationaryNode` must be called _before_ creating the new block, to avoid updating the cursor. I could pass the stationary node into the drag strategy instead of the start point, but in both cases I'm doing layer bypass to move information around. So that was a dud.
 
A few changes:
- Pass the moving block into `startMove`, instead of expecting the mover to find it by looking at the cursor.
- Make the keyboard drag strategy responsible for selecting the moving block, rather than `startMove`.
- Make paired helpers to patch/unpatch the dragger. This is mostly so that we can see what we need to clean up later.
  - (Unfortunately, as I typed this I realized that the dragger is created new for every drag, so there is no point in unpatching it at the end).
- Set the cursor correctly after aborting an insertion https://github.com/google/blockly-keyboard-experimentation/commit/4da9e6df5400737c9821eb5d4e7d0e0d95288d03
- Fix event grouping so that undo deletes the newly inserted block https://github.com/google/blockly-keyboard-experimentation/commit/565031a1e6f61cf30832097758c559bc3a5e8a34
- Handle the issue when a value block is selected https://github.com/google/blockly-keyboard-experimentation/commit/f648b464e6828e6f93beb168a3fd9f0963bde56e

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -133,14 +133,9 @@ export class EnterAction {
     const stationaryNode = this.navigation.getStationaryNode(workspace);
     const newBlock = this.createNewBlock(workspace);
     if (!newBlock) return;
-    if (stationaryNode) {
-      if (!this.navigation.tryToConnectBlock(stationaryNode, newBlock)) {
-        console.warn(
-          'Something went wrong while inserting a block from the flyout.',
-        );
-      }
-    }
-
+    const insertStartPoint = stationaryNode
+      ? this.navigation.findInsertStartPoint(stationaryNode, newBlock)
+      : null;
     if (workspace.getTopBlocks().includes(newBlock)) {
       this.positionNewTopLevelBlock(workspace, newBlock);
     }
@@ -151,7 +146,7 @@ export class EnterAction {
     this.navigation.focusWorkspace(workspace);
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     workspace.getCursor()?.setCurNode(ASTNode.createBlockNode(newBlock)!);
-    this.mover.startMove(workspace, newBlock, true);
+    this.mover.startMove(workspace, newBlock, insertStartPoint);
 
     const isStartBlock =
       !newBlock.outputConnection &&
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -151,7 +151,7 @@ export class EnterAction {
     this.navigation.focusWorkspace(workspace);
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     workspace.getCursor()?.setCurNode(ASTNode.createBlockNode(newBlock)!);
-    this.mover.startMove(workspace, newBlock);
+    this.mover.startMove(workspace, newBlock, true);
 
     const isStartBlock =
       !newBlock.outputConnection &&
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -128,6 +128,7 @@ export class EnterAction {
    */
   private insertFromFlyout(workspace: WorkspaceSvg) {
     workspace.setResizesEnabled(false);
+    // Create a new event group or append to the existing group.
     const existingGroup = Events.getGroup();
     if (!existingGroup) {
       Events.setGroup(true);
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -128,7 +128,10 @@ export class EnterAction {
    */
   private insertFromFlyout(workspace: WorkspaceSvg) {
     workspace.setResizesEnabled(false);
-    Events.setGroup(true);
+    const existingGroup = Events.getGroup();
+    if (!existingGroup) {
+      Events.setGroup(true);
+    }
 
     const stationaryNode = this.navigation.getStationaryNode(workspace);
     const newBlock = this.createNewBlock(workspace);
@@ -140,7 +143,6 @@ export class EnterAction {
       this.positionNewTopLevelBlock(workspace, newBlock);
     }
 
-    Events.setGroup(false);
     workspace.setResizesEnabled(true);
 
     this.navigation.focusWorkspace(workspace);
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -151,7 +151,7 @@ export class EnterAction {
     this.navigation.focusWorkspace(workspace);
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     workspace.getCursor()?.setCurNode(ASTNode.createBlockNode(newBlock)!);
-    this.mover.startMove(workspace);
+    this.mover.startMove(workspace, newBlock);
 
     const isStartBlock =
       !newBlock.outputConnection &&
--- a/src/actions/move.ts
+++ b/src/actions/move.ts
@@ -5,6 +5,7 @@
  */
 
 import {
+  BlockSvg,
   ContextMenuRegistry,
   ShortcutRegistry,
   utils,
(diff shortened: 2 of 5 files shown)


microbit-matt-hillsdon:
'M' then 'Escape' on an existing statement block now moves the cursor between that block and the previous block which is a regression. Even weirder if you move it before cancelling as the connection location is based on where the preview was shown not where it moves back to on cancel.

For a value block (e.g. test in `if`) selection seems to be lost entirely in the same scenario (or at least is not visible).

BenHenning:
Sorry, will need to take a look at this tomorrow. There's a lot of context that I don't know here that I'll need to familiarize myself with in order to do an effective review.

Feel free to redirect to someone who has more context if that's preferable, but otherwise I will take a pass tomorrow (likely in the early afternoon PT).

src/actions/mover.ts
```diff
@@ -160,21 +176,37 @@ export class Mover {
...
+    // @ts-expect-error Access to private property connectionCandidate.
+    dragStrategy.connectionCandidate = null;
+
     info.dragger.onDragEnd(
       info.fakePointerEvent('pointerup'),
-      new utils.Coordinate(0, 0),
+      info.startLocation,
     );
 
+    if (target) {
```
microbit-matt-hillsdon: This seems a straightforward fix for my feedback:

```suggestion
    if (dragStrategy.isNewBlock && target) {
```

If we go ahead with removing connection positions then the code in the block will likely need a tweak too, but let's get this in ahead of thinking about that.

microbit-matt-hillsdon (commented):
Keen to get this one merged. It looks good from my perspective except for the one note I added previously re cancelling a move of a preexisting block. I've added a suggested fix.

@rachel-fenichel there's also your note in the PR description about the lack of need to unpatch the dragger as it's recreated. So that's potentially some easy simplification.

src/actions/enter.ts
```diff
@@ -126,30 +126,27 @@ export class EnterAction {
    */
   private insertFromFlyout(workspace: WorkspaceSvg) {
     workspace.setResizesEnabled(false);
-    Events.setGroup(true);
+    const existingGroup = Events.getGroup();
+    if (!existingGroup) {
```
BenHenning, line 133: Maybe add a line comment here explaining why this is important (otherwise the context feels maybe a bit hidden).

src/navigation.ts
```diff
@@ -726,38 +728,62 @@ export class Navigation {
...
+    return null;
   }
 
+  /**
+   * Tries to intelligently connect the blocks or connections
+   * represented by the given nodes, based on node types and locations.
+   *
+   * @param stationaryNode The first node to connect.
+   * @param movingBlock The block we're moving.
+   * @returns True if the key was handled; false if something went
```
BenHenning: Is this correct? Seems like it should be tied to whether the connection was successful (per `insertBlock`'s docs), not anything about keys.

src/actions/move.ts
```diff
@@ -28,8 +29,16 @@ export class MoveActions {
     // Begin and end move.
     {
       name: 'Start move',
-      preconditionFn: (workspace) => this.mover.canMove(workspace),
-      callback: (workspace) => this.mover.startMove(workspace),
+      preconditionFn: (workspace) => {
+        const startBlock = this.getCurrentBlock(workspace);
+        return !!startBlock && this.mover.canMove(workspace, startBlock);
```
BenHenning, line 34: Why not just `startBlock && ...`? '!!' seems unnecessary here. Ditto below in a few places.

src/actions/move.ts
```diff
@@ -168,4 +184,30 @@ export class MoveActions {
...
+   *
+   * @param workspace The workspace to inspect for a cursor.
+   * @returns The source block, or undefined if no appropriate block
+   *     could be found.
+   */
+  getCurrentBlock(workspace: WorkspaceSvg): BlockSvg | undefined {
+    const curNode = workspace?.getCursor()?.getCurNode();
+    let block = curNode?.getSourceBlock();
+    if (!block) return undefined;
+    while (block && block.isShadow()) {
```
BenHenning: I'm assuming the truthy check for `block` here is because TypeScript doesn't realize that it's impossible for it to ever be null or undefined?

If it's null inference is actually good enough to detect that, you should be able to drop that part of the condition.

src/keyboard_drag_strategy.ts
```diff
@@ -35,7 +36,24 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
...
+  ) {
+    super(block);
+    this.isNewBlock = !!this.insertStartPoint;
+    this.cursor = block.workspace.getCursor();
+  }
+
   override startDrag(e?: PointerEvent) {
+    if (!this.cursor) throw new Error('precondition failure');
+    // Select and focus block.
+    common.setSelected(this.block);
```
BenHenning: Won't `setCurNode` guarantee selection? It'd be nice to consolidate since otherwise every setSelected() needs to be reconsidered for focus.

src/keyboard_drag_strategy.ts
```diff
@@ -142,8 +159,9 @@ export class KeyboardDragStrategy extends dragging.BlockDragStrategy {
     draggingBlock: BlockSvg,
     localConns: RenderedConnection[],
   ): ConnectionCandidate | null {
-    const cursor = draggingBlock.workspace.getCursor();
-    if (!cursor) return null;
+    // TODO: Handle the case where the cursor is null, or never return null
```
BenHenning: This seems worthwhile to maybe file an issue for and point to it.

BenHenning (approved):
Apologies for the delayed review. I had a few comments, but they're largely nits so going ahead and approving.

The solutions make sense to me, though admittedly I don't grok the complete dragging context. :)

src/actions/move.ts
rachel-fenichel: Needs to return a boolean, not null, and the compiler yelled.

src/actions/enter.ts
rachel-fenichel: Added a short comment. This is also a pattern in all of Blockly.

src/actions/move.ts
rachel-fenichel: Done.

src/actions/mover.ts
rachel-fenichel: Done.

src/keyboard_drag_strategy.ts
rachel-fenichel: Removed `setSelected`

src/keyboard_drag_strategy.ts
rachel-fenichel: Filed https://github.com/google/blockly-keyboard-experimentation/issues/471 and removed the comment here because it's a general problem as described by the issue, this is just an instance of it.

src/navigation.ts
rachel-fenichel: Fixed.

rachel-fenichel:
@gonfunko please take a look--I fixed the issues Ben identified + accepted Matt's suggested patch. I think the tests failing are the ones that Erik is currently working on/investigating.

gonfunko (approved):

rachel-fenichel:
Retested manually and with automated tests after https://github.com/google/blockly/pull/8933 and it still works, so I'm merging now.

PR #457, opened 2025-04-21
feat: enable blocks on drag

### LINKED ISSUE #397: Not clear that a disabled block is in a move state
Disabled and selected:

Image

Disabled and selected and in move state after pressing "M":

Image

I can't tell and I think keyboard move mode needs to be clearly indicated somehow. Worth discussing more widely whether it's clear enough for regular blocks too before making a change for this.

maribethb:
Fixes #397 

Enables the blocks on drag, disables them at end of drag

See issue for demo

### DIFF
--- a/src/disabled_blocks.ts
+++ b/src/disabled_blocks.ts
@@ -1,0 +1,84 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+const lastBlockDisabledReasons: Map<string, Set<string>> = new Map();
+
+/**
+ * A change listener that enables disabled blocks when they
+ * are dragged, and re-disables them at the end of the drag.
+ *
+ * @param event Blockly event
+ */
+export function enableBlocksOnDrag(event: Blockly.Events.Abstract) {
+  // This listener only runs on Drag events that have a valid
+  // workspace and block id.
+  if (!isBlockDrag(event)) return;
+  const eventWorkspace = Blockly.common.getWorkspaceById(
+    event.workspaceId,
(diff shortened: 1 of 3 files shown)


test/index.ts
```diff
@@ -23,6 +23,8 @@ import {javascriptGenerator} from 'blockly/javascript';
 import {load} from './loadTestBlocks';
 import {runCode, registerRunCodeShortcut} from './runCode';
 
+(window as any).Blockly = Blockly;
```
gonfunko, line 26: Are you intending to check this in? I kinda think we should because I always wind up adding it too, but not sure if it was just left over or not.

src/disabled_blocks.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param event Blockly event
+ */
+export function enableBlocksOnDrag(event: Blockly.Events.Abstract) {
+  // This listener only runs on Drag events that have a valid
+  // workspace and block id.
+  if (!isBlockDrag(event)) return;
+  const eventWorkspace = Blockly.common.getWorkspaceById(
+    event.workspaceId,
+  ) as Blockly.WorkspaceSvg;
+  if (!event.blockId) return;
```
gonfunko: Maybe move this check before retrieving the workspace?

gonfunko (approved):

test/index.ts
maribethb: Oh yeah it was just left over, but we should check it in. it's in the test page so doesn't affect the published plugin, i don't see a downside

src/disabled_blocks.ts
maribethb: Done

test/index.ts
gonfunko: SGTM!

Issue #412, opened 2025-04-09
All user-visible strings need to be in Blockly.Msg in core


maribethb:
Currently there are a lot of user-visible strings in the bare code (e.g. returned as strings from `displayText` functions). These are not translateable for two reasons:

1) They're not in Blockly.Msg
2) This repo is not hooked up to translatewiki (neither is blockly-samples fwiw)

In order for these messages to be translated, they need to be included in the messages in blockly core, even if they are only referenced from the plugin/experiment.

It would be nice if we could make them available for translatewiki to start translating before the release happens, but that would be simplest if translatewiki were hooked up to develop. it's probably hooked up to master though, so this might just be a "nice to have"

rachel-fenichel:
@riknoll how does MakeCode do translation at this point? Do you rely on Blockly's translations for basic messages, or your own?

riknoll:
we use our own translations for everything; for the blocks from blockly core we override whatever is in Blockly.Msg like so: https://github.com/microsoft/pxt/blob/master/pxtblocks/builtins/logic.ts#L14

PR #458, opened 2025-04-21
refactor: Use localizable messages.

gonfunko:
This PR fixes #[412](https://github.com/google/blockly-keyboard-experimentation/issues/412) by updating the keyboard experimentation repo to reference localizable strings from core Blockly instead of hard coding them.

### DIFF
--- a/src/actions/delete.ts
+++ b/src/actions/delete.ts
@@ -7,10 +7,12 @@
 import {
   ContextMenuRegistry,
   Gesture,
+  Msg,
   ShortcutRegistry,
   utils as BlocklyUtils,
   LineCursor,
 } from 'blockly';
+import {getShortActionShortcut} from '../shortcut_formatting';
 import * as Constants from '../constants';
 import type {WorkspaceSvg} from 'blockly';
 import {Navigation} from '../navigation';
@@ -102,8 +104,9 @@ export class DeleteAction {
 
     const deleteItem: ContextMenuRegistry.RegistryItem = {
       displayText: (scope) => {
+        const shortcut = getShortActionShortcut(this.deleteShortcutName);
         if (!this.oldContextMenuItem) {
-          return 'Delete block (Del)';
+          return Msg['DELETE_BLOCK'].replace('%1', shortcut);
         }
 
         type DisplayTextFn = (p1: ContextMenuRegistry.Scope) => string;
@@ -111,7 +114,7 @@ export class DeleteAction {
         // of blocks that will be deleted.
         const oldDisplayText = this.oldContextMenuItem
           .displayText as DisplayTextFn;
-        return oldDisplayText(scope) + ' (Del)';
+        return oldDisplayText(scope) + ` (${shortcut})`;
       },
       preconditionFn: (scope, menuOpenEvent: Event) => {
         const ws = scope.block?.workspace;
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -6,6 +6,7 @@
 
 import {
   Events,
+  Msg,
   ShortcutRegistry,
   utils as BlocklyUtils,
   getFocusManager,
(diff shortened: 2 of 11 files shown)


src/actions/delete.ts
```diff
@@ -102,16 +104,17 @@ export class DeleteAction {
...
+          return Msg['DELETE_BLOCK'].replace('%1', shortcut);
         }
 
         type DisplayTextFn = (p1: ContextMenuRegistry.Scope) => string;
         // Use the original item's text, which is dynamic based on the number
         // of blocks that will be deleted.
         const oldDisplayText = this.oldContextMenuItem
           .displayText as DisplayTextFn;
-        return oldDisplayText(scope) + ' (Del)';
+        return oldDisplayText(scope) + ` (${shortcut})`;
```
BenHenning, line 117: Will this localize for RTL correctly?

src/actions/enter.ts
```diff
@@ -105,7 +106,7 @@ export class EnterAction {
       const block = curNode.getLocation() as Block;
       if (!this.tryShowFullBlockFieldEditor(block)) {
         const shortcut = getShortActionShortcut('list_shortcuts');
-        const message = `Press ${shortcut} for help on keyboard controls`;
+        const message = Msg['HELP_PROMPT'].replace('%1', shortcut);
```
BenHenning: Here & elsewhere: how descriptive do the names need to be? If they are global context, I'd expect something like: `KEYBOARD_SHORTCUTS_HELP_PROMPT` to be much clearer and avoid potential duplications in the future. If this level of specificity is preferred, can all of the other ones be updated accordingly?

BenHenning (approved):
Thanks @gonfunko! LGTM, though I did have some thoughts on the messages naming (though that would presumably require changes here & in v12, so I defer to your judgment on how you want to proceed).

src/actions/delete.ts
gonfunko: No, but this is kind of a hack in the first place so I think that's acceptable for now.

src/actions/enter.ts
gonfunko: ...I realize we just release core, so we're kind of committed now. Sorry I missed this until I went looking for things to pick up now that that was out the door :/

rachel-fenichel:
I think this needs one more rebase.

src/actions/move.ts
```diff
@@ -25,163 +27,179 @@ const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
...
-      preconditionFn: (workspace) => {
-        const startBlock = this.getCurrentBlock(workspace);
-        return !!startBlock && this.mover.canMove(workspace, startBlock);
+  private shortcutNames: string[] = [];
+  private menuItemNames: string[] = [];
+
+  private registerShortcuts() {
+    const shortcuts: ShortcutRegistry.KeyboardShortcut[] = [
+      // Begin and end move.
+      {
```
cpcallen, line 36: @gonfunko:

Moving the list of shortcuts from the class definition into a method makes it impossible to reference them from elsewhere, obliging you to add a separate private member just the shortcuts can later be unregistered.  This is inelegant and complicates certain other refactors, such as making [this horrid bit of code](https://github.com/google/blockly-keyboard-experimentation/blob/29e4b031c89dde08c3e962a299fcefdd6f7c36e4/src/actions/mover.ts#L154) a bit less opaque by having it actually reference the keybindings used by these shortcuts.

Since you are not an idiot, I presume you must have had a good reason to do this.

One obvious good reason would be to delay evaluation of the `Msg[…]` lookups until after messages are loaded.  But I cannot see any point in time between [when the `MoveActions` constructor is called](https://github.com/google/blockly-keyboard-experimentation/blob/29e4b031c89dde08c3e962a299fcefdd6f7c36e4/src/navigation_controller.ts#L76) and when [moveActions.init()](https://github.com/google/blockly-keyboard-experimentation/blob/29e4b031c89dde08c3e962a299fcefdd6f7c36e4/src/navigation_controller.ts#L250) is called when it is possible to load messages, since both things happen as a result of the construction of the `NavigationController` instance.

What am I missing here?

PR #460, opened 2025-04-21
fix: Make navigation direction align with arrow keys in RTL.

### LINKED ISSUE #249: Handle RTL for directional shortcuts
previous/next/in/out actions should be mapped appropriately to the left/right/up/down keys. This is dynamic (at installation time?) because workspaces may work left-to-right or right-to-left, and may have horizontal toolboxes and block renderers.

Currently, to navigate from the toolbox to the flyout, one uses the right arrow key, and then to navigate between items in the flyout, the up and down arrow keys are used. This is awkward for horizontal flyouts, where it would be more natural to use the up or down keys to navigate between the toolbox/flyout (depending on toolbox placement) and left and right to navigate between flyout items.

2d navigation on the workspace will not be flipped for RTL.

gonfunko:
This PR fixes #249 by inverting whether left/right arrow triggers the in or out action based on whether the workspace is RTL or LTR. This gives the intuitive behavior in both cases; pressing the left arrow moves you to the thing visually to the left, and vice versa for the right arrow. In particular, navigating between the toolbox and flyout and traversing through the fields in a block are much more comfortable in RTL mode now.

### DIFF
--- a/src/actions/arrow_navigation.ts
+++ b/src/actions/arrow_navigation.ts
@@ -47,6 +47,70 @@ export class ArrowNavigation {
    * Adds all arrow key navigation shortcuts to the registry.
    */
   install() {
+    const navigateIn = (
+      workspace: WorkspaceSvg,
+      e: Event,
+      shortcut: ShortcutRegistry.KeyboardShortcut,
+    ): boolean => {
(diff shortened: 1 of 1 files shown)


BenHenning (approved):
Changes makes sense, thanks @gonfunko!

PR #463, opened 2025-04-23
chore: Implement initial block navigation tests.

johnnesky:
Implements several unit tests described at https://docs.google.com/document/d/1pIIX5sKEG9f_NcWDIQJJEOMM7w7pQUf8eJoT3lrAElQ involving navigating with the arrow keys when starting from a block:

* Pressing down from a statement block selects its next connection.
* Pressing up from a statement block selects the previous statement's next connection.
* Pressing down from a parent block selects its input connection.
* Pressing up from a child block selects its parent's input connection.
* Pressing right from a block selects its first field.

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -129,4 +129,9 @@ document.addEventListener('DOMContentLoaded', () => {
   addP5();
   createWorkspace();
   document.getElementById('run')?.addEventListener('click', runCode);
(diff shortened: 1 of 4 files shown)


rachel-fenichel (approved):
Please fix lint, then LGTM.

PR #469, opened 2025-04-25
chore: Improve safety of flyout change listener.

gonfunko:
This PR fixes an issue where the flyout change listener could throw if an event's workspace wasn't found. This happens when attempting to integrate with Scratch, which does a fair bit of initialization work that might result in that scenario.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -239,9 +239,9 @@ export class Navigation {
     if (!e.workspaceId) {
       return;
     }
-    const flyoutWorkspace = Blockly.Workspace.getById(e.workspaceId) as
-      | Blockly.WorkspaceSvg
-      | undefined;
+    const flyoutWorkspace = Blockly.Workspace.getById(
+      e.workspaceId,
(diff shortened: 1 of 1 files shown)


src/navigation.ts
```diff
@@ -239,10 +239,10 @@ export class Navigation {
     if (!e.workspaceId) {
       return;
     }
-    const flyoutWorkspace = Blockly.Workspace.getById(
-      e.workspaceId,
-    ) as Blockly.WorkspaceSvg;
-    const mainWorkspace = flyoutWorkspace.targetWorkspace;
+    const flyoutWorkspace = Blockly.Workspace.getById(e.workspaceId) as
```
rachel-fenichel: If you don't cast it, `getById` returns `WorkspaceSvg | null` and that would work just as well.

gonfunko: It appears to be `Workspace | null`? But updated to `null` at least.

rachel-fenichel (approved):
Ah I see, so it still needs to cast to convince it to be a `WorkspaceSvg`.

PR #470, opened 2025-04-25
Fix: Fix tests broken by changes to block positions

### LINKED ISSUE #468: Tests failing due to changes in core
### Check for duplicates

### Description

Three of the tests are failing due to something that changed in core. Likely due to cursor changes

Starting webdriverio...
  Keyboard navigation
    ✔ Default workspace
    1) Selected block

  Clipboard test
    2) Copy and paste while block selected
    3) Cut and paste while block selected
    ✔ Copy and paste whilst dragging block (391ms)
    ✔ Cut whilst dragging block (387ms)

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

RoboErikG:
Fixes #468 

It appears the blocks used in the test shifted positions slightly leading to two issues with the tests.

1. Generating a click event on the workspace was landing on one of the blocks in the stack, which broke an assumption that focus would be on the workspace. This was fixed by clicking a specific block instead.
2. Clicking on a block that was slightly out of the window's viewport was generating a click event that the browser ignored. This was fixed by scrolling the block into view before trying to click it.

This PR also has a couple minor non-functional tweaks for correctness:

1. Changes to using 0 instead of 1 for the mouse button options (left button instead of middle).
2. Moves the fallback to getting the block's root svg inside the check that the block exists.

### DIFF
--- a/test/loadTestBlocks.js
+++ b/test/loadTestBlocks.js
@@ -182,7 +182,7 @@ const simpleCircle = {
     'blocks': [
       {
         'type': 'p5_setup',
-        'id': '5.{;T}3Qv}Awi:1M$:ut',
+        'id': 'p5_setup_1',
         'x': 0,
         'y': 75,
         'deletable': false,
--- a/test/webdriverio/test/basic_test.ts
+++ b/test/webdriverio/test/basic_test.ts
@@ -6,8 +6,13 @@
 
 import * as chai from 'chai';
 import * as Blockly from 'blockly';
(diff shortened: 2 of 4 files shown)


RoboErikG:
This seems to be necessary but not sufficient. There was a previous issue as stated in the PR description, but after syncing to head there's also a new issue that causes the same failures.

Please review this change as is and I'll look into fixing the second issue separately.

BenHenning (commented):
Thanks @RoboErikG! These fixes seem reasonable.

PR #472, opened 2025-04-25
feat: Add a visual indicator to blocks in move mode.

### LINKED ISSUE #465: Add clear indicator that a block is in move mode
For user testing we hacked in a move indicator and it really helps clarify that a block is in move mode and evoke using the keyboard to move it around. Especially useful for top-level blocks and disabled top-level blocks.

Demo: https://move-indicator.blockly-keyboard-experimentation.pages.dev/

Screengrab:

Image

We'd like to work up a proper PR for this but starting with this issue to collect feedback and any suggestions. At the moment we haven't had any real visual feedback, but I'd favour getting an indicator like this in place and then improving it.

In the linked demo I think it's possible to get the block selection outline drawn over the move indicator. It's possible #462 will fix this.

gonfunko:
This PR fixes #465 by adding a new icon and associated bubble that display a four-way-arrow move indicator attached to the corner of blocks while they are in move mode.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -17,7 +17,6 @@ import * as Constants from '../constants';
 import {Direction, getXYFromDirection} from '../drag_direction';
 import {KeyboardDragStrategy} from '../keyboard_drag_strategy';
 import {Navigation} from '../navigation';
-import {MoveIcon} from '../move_icon';
 import {clearMoveHints} from '../hints';
 
 /**
@@ -124,7 +123,6 @@ export class Mover {
     // Begin drag.
     dragger.onDragStart(info.fakePointerEvent('pointerdown'));
     info.updateTotalDelta();
-    block.addIcon(new MoveIcon(block));
     return true;
   }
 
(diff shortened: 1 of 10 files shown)


test/index.ts
```diff
@@ -90,6 +90,7 @@ function createWorkspace(): Blockly.WorkspaceSvg {
   const injectOptions = {
     toolbox,
     renderer,
+    rtl: true,
```
rachel-fenichel: Remove from test page.

src/actions/mover.ts
```diff
@@ -123,6 +124,7 @@ export class Mover {
     // Begin drag.
     dragger.onDragStart(info.fakePointerEvent('pointerdown'));
     info.updateTotalDelta();
+    block.addIcon(new MoveIcon(block));
```
rachel-fenichel: What's the reason to add the icon here instead of in the keyboard drag strategy?

test/index.ts
gonfunko: Done.

src/actions/mover.ts
gonfunko: It was the first place I ran across to do it, but agreed that makes more sense; moved it accordingly.

rachel-fenichel (approved):
Approving. One of the build failures in CI is on a line that this PR doesn't change, which makes me suspect issues with the simultaneous changes in core and here. Aaron verified that it was able to build and run locally.

@maribethb FYI for when you get back, in case you have opinions about this approach.

PR #477, opened 2025-04-29
chore: Remove now-obsolete call to `connectionPreviewer.hidePreview`

### LINKED ISSUE #8853: Hide connection previews in `revertDrag`
### Check for duplicates

### Problem

While implementing constrained drag in keyboard experimentation, I found that `revertDrag` in `BlockDragStrategy` does not call `connectionPreviewer.hidePreview`.

So far this has been safe because drags are only reverted when dropped onto an invalid location.

With the option to end a drag with escape, it's important to fully clean up in `revertDrag`.

### Request

Call `connectionPreviwer.hidePreview` in `revertDrag`. Should be a no-op if no preview was being shown.

### Alternatives considered

_No response_

### Additional context

_No response_

cpcallen:
This is no longer needed after google/blockly#8934.

Fixes google/blockly#8853.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -177,10 +177,6 @@ export class Mover {
       dragStrategy.isNewBlock,
     );
 
-    // Explicitly call `hidePreview` because it is not called in revertDrag.
-    // @ts-expect-error Access to private property connectionPreviewer.
-    dragStrategy.connectionPreviewer.hidePreview();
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
Code looks good but your issue/pr links are wrong.

rachel-fenichel (approved):

Issue #476, opened 2025-04-29
Error inserting Function block with block already selected


microbit-matt-hillsdon:
1. In the default demo page (but not the deployed version, that's too old), select a block, e.g. draw circle
2. Insert a function via the keyboard
3. Logged error

```
Cannot read properties of null (reading 'type')
TypeError: Cannot read properties of null (reading 'type')
    at KeyboardDragStrategy.forceShowPreview (webpack-internal:///./src/keyboard_drag_strategy.ts:202:47)
    at KeyboardDragStrategy.startDrag (webpack-internal:///./src/keyboard_drag_strategy.ts:42:14)
    at BlockSvg$$module$build$src$core$block_svg.startDrag (webpack-internal:///../blockly/dist/blockly_compressed.js:1003:508)
    at Dragger$$module$build$src$core$dragging$dragger.onDragStart (webpack-internal:///../blockly/dist/blockly_compressed.js:1540:799)
    at Mover.startMove (webpack-internal:///./src/actions/mover.ts:108:17)
    at EnterAction.insertFromFlyout (webpack-internal:///./src/actions/enter.ts:132:20)
    at Object.callback (webpack-internal:///./src/actions/enter.ts:57:38)
    at ShortcutRegistry$$module$build$src$core$shortcut_registry.onKeyDown (webpack-internal:///../blockly/dist/blockly_compressed.js:1380:138)
    at onKeyDown$$module$build$src$core$inject (webpack-internal:///../blockly/dist/blockly_compressed.js:447:269)
    at HTMLDivElement.f (webpack-internal:///../blockly/dist/blockly_compressed.js:98:159)
```

Core hash dee27b905dd4cfbae1f564192d594e401c19b468, plugin hash 25fe2babd594278c9917b7465023d88bb4def314

PR #479, opened 2025-04-29
feat: paste in move mode

### LINKED ISSUE #478: Put pasted blocks in move mode
We use move mode for insert from the flyout, but not for paste (or duplicate, which I'll raise a separate issue for).

Paste should be consistent with insert and share code/behaviour for e.g. top-level block placement.

See also https://github.com/google/blockly-keyboard-experimentation/issues/480 for Duplicate.

microbit-matt-hillsdon:
Share insert code with flyout insert.

This aligns top level block placement too.

Potentially we use the same approach for Duplicate too.

Closes #478

### DIFF
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -20,6 +20,7 @@ import {Navigation} from '../navigation';
 import {getShortActionShortcut} from '../shortcut_formatting';
 import * as Blockly from 'blockly';
 import {clearPasteHints, showCopiedHint, showCutHint} from '../hints';
+import {Inserter} from './inserter';
 
 const KeyCodes = blocklyUtils.KeyCodes;
 const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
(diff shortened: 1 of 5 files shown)


microbit-matt-hillsdon:
I found https://github.com/google/blockly-keyboard-experimentation/issues/476 while testing this but it predates this change.

RoboErikG:
Code looks good overall, but for now lets go with placing the block instead of putting it in move mode both to be consistent with other editors and so it doesn't cause issues with using a mouse for copy/pasting.

RoboErikG (changes requested):
Please update to place the block instead of leaving in move mode.

microbit-matt-hillsdon:
Thanks @RoboErikG, I agree this needs more thought for the impact on mouse users and it looks like we have copy/paste input inbound from user testing so I'll park this one for a bit until I get my head around that.

microbit-matt-hillsdon:
Closing in favour of keeping paste more like regular Blockly as on #501.

PR #482, opened 2025-04-29
feat: Use FocusManager for navigation & state management

### LINKED ISSUE #142: Use focus to denote cursor location

### LINKED ISSUE #481: FocusManager breaks insertion & toolbox shortcuts
### Check for duplicates

### Description

See https://github.com/google/blockly/pull/8933#pullrequestreview-2795396046 for specifics, but 'T' and 'I' break with the focus manager changes being introduced.

### Reproduction steps

Try 'I' and 'T' along with https://github.com/google/blockly/pull/8875#issuecomment-2814265889 and see that they don't work as expected.

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
Fixes #142
Fixes #481

This removes nearly all custom focus/blur management and instead leverages `FocusManager` as the source of truth and primary system for focusing/selecting elements.

This largely continues the work in Core starting with https://github.com/google/blockly/pull/8938 and ending with https://github.com/google/blockly/pull/8941 in order to bring basic parity in the keyboard navigation plugin using `FocusManager`.

Specifically:
- This replaces all cases where explicit focus/opening logic is needed with `focusNode`/`focusTree` calls via `FocusManager`.
- `Navigation` now infers state based on what's currently focused rather than tracking it separately.
- All of the existing focus/blur response logic has been made redundant due to the inferred navigation mode and focus being tightly coupled with the components themselves.
- The gesture monkey patch seems no longer necessary since, as far as I can tell, focus gets forced back to where it's supposed to be at specific lifecycle moments (which helps undo many of the problems caused by `WorkspaceSvg`'s `markFocused`).
- The passive indicator is no longer necessary now that the `FocusManager`-driven `blocklyPassiveFocus` CSS class exists for styling.
- Note that there's a lot of automatic state synchronizing that has been purposely pushed into Core to simplify the navigation pieces, particularly:
  - Blocks will auto-select/unselect themselves based on focus.
  - `Toolbox` will automatically synchronize its selected item state with its item that has active focus.
  - `Toolbox` and `Flyout` cooperate for automatically closing the `Flyout` at the correct times.
- The new CSS classes are not yet final as the overall strategy for indicating active/passive focus isn't final. There's also additional work needed to ensure selection and active focus can be properly stylized as a combined state, but this is blocked on the completion of https://github.com/google/blockly/issues/8942.
- `Toolbox` navigation was moved to using its methods directly instead of its event since selection logic now needs to be paired with `focusNode`.
- There are a number of changes in this PR that could probably go into Core, but these can wait for a future PR as they shouldn't be API breaking. Some of the changes include:
  - `Toolbox` item selection could auto-focus within `Toolbox` itself, or this could be done as part of `Toolbox`-specific navigation.
  - `Flyout` focus could automatically focus its first item if nothing is currently focused.
- For regression testing, the following has been checked:
  - 'T' and 'I' do work as expected (ensuring https://github.com/google/blockly/pull/8933#pullrequestreview-2795396046 is no longer an issue).
  - Verified #227, #200, #229, #222, and #287 have not regressed. https://github.com/google/blockly-samples/issues/2498 shouldn't be possible anymore (even with the fix) due to ephemeral focus, and we seem to be matching parity with #326 per the `handleFocus*()` functions, though there may be cases from `handleFocusWorkspace` yet to check.

The test updates were necessary because, at present, clicking does not force focus which means the tests need to rely purely on keys for navigation. I think these changes are actually an improvement since it closer aligns them with keyboard-only usage.

**Current gaps** that may need resolution prior to this PR being mergeable:
- There's a regression with clicking to create a variable closing the flyout that I'm still investigating. This is detailed more in https://github.com/google/blockly/pull/8941.
- At some point between the latest changes to this branch and the ones in https://github.com/google/blockly/pull/8941 block movement regressed slightly: finishing a movement gesture now defocuses the workspace entirely (even though this was working earlier).
- There are a lot of inconsistencies between the focus and selection styling before this PR and with this PR. It's unclear how much of this requires actual resolution.
- Much more testing is needed to understand gaps as this changes a significant amount of state handling for the plugin.


BenHenning:
I still need to push my latest changes (currently working through some focus issues with variable creation), but now that block movement is working I'm pretty happy with the stability of this PR with regards to it being testable. It's definitely not yet ready for review, but I'm hopeful the delta won't be too large.

BenHenning (commented):
Self-reviewed changes.

BenHenning:
PTAL @rachel-fenichel for an initial review pass.

If you can, I'd also really appreciate your check on this @microbit-matt-hillsdon as it makes substantial changes to focus management that you had carefully considered as part of #326. I think that I've matched parity here with the removed focus handlers, but I can't 100% certain for flyout (due to `markFocused` quirks, though it seems to be working well for me minus the variables issues) and for workspace (since it had a few extra bits that are no longer covered).

BenHenning:
The build & test failures were partly due to https://github.com/google/blockly/pull/8941 not yet being merged, but it's merged now.

There were still test updates needed which I will document shortly.

BenHenning (commented):
Self-reviewed changes.

BenHenning:
Thanks to the wonderful testing from @microbit-grace and @microbit-matt-hillsdon there are ~40 issues that have been found with this PR. However, I think we may still want to merge this as-is since, without these changes, the plugin is entirely unusable with tip-of-tree v12 (especially after https://github.com/google/blockly/pull/8941 was merged). I think we can follow up with the needed fixes in Core and the plugin to address those problems, plus a number warrant additional discussion (especially around styling).

I defer to you @rachel-fenichel on this.

src/actions/enter.ts
```diff
@@ -9,14 +9,12 @@ import {
...
   utils as BlocklyUtils,
-} from 'blockly/core';
-
-import type {
   Block,
   BlockSvg,
   Field,
   FlyoutButton,
   WorkspaceSvg,
+  getFocusManager,
```
rachel-fenichel: Should not need to change all of the `import type`s to `imports` just to add the focus manager.

rachel-fenichel (approved):
Let's merge this to move from "head is completely broken" to "head has broken flows that we can debug".

src/actions/enter.ts
BenHenning: Oops, I think I missed the type specifier and thought there were redundant import lines (which, in retrospect, I now realize would've been caught by the linter). Fixed in latest.

BenHenning (commented):
Self-reviewed latest fix.

BenHenning:
Going ahead and merging this as discussed above. Thanks for the review!

PR #486, opened 2025-05-02
chore: Implement field navigation tests.

johnnesky:
Implements several unit tests described in https://docs.google.com/document/d/1pIIX5sKEG9f_NcWDIQJJEOMM7w7pQUf8eJoT3lrAElQ involving navigating with the arrow keys when starting from a field or inline input.

I ran these tests using old code for both core blockly and keyboard experimentation. I built against a slightly older commit from core blockly rc/v12.0.0, namely this one https://github.com/google/blockly/commit/dee27b905dd4cfbae1f564192d594e401c19b468 since the most recent commit seems unstable at the moment. I also did not update to the latest commit of keyboard experimentation before writing these tests, namely I remained at: https://github.com/google/blockly-keyboard-experimentation/commit/c90aae92f45a889a3efcc84d5ee9f17f945bbd0a

Updating the code in either repo introduces compilation errors that I'm not sure how to address yet.

The last test I left commented out because it fails even though I wrote it with the expectation that it would succeed. It seems that the curly quote icons in a text block count as selection stops when pressing the right arrow key, even though they are not interactive. If the quotes are fixed, then the test can be uncommented to verify that the right arrow key navigates past them.


rachel-fenichel:
@RoboErikG please rebase this PR once sufficient changes are made in core and over here to make it build, fix lint/format, and then review the actual test contents and merge if good.

Thanks John for writing these tests!

RoboErikG:
Replaced by https://github.com/google/blockly-keyboard-experimentation/pull/496

PR #490, opened 2025-05-05
fix: remove calls to cursor.hide

rachel-fenichel:
Part of https://github.com/google/blockly/issues/8962

Related to https://github.com/google/blockly/pull/8924

The keyboard navigation plugin currently explicitly turns keyboard mode on and off, and that includes explicitly hiding the cursor. After https://github.com/google/blockly/pull/8924 we will use CSS to control whether active and passive focus are visible, and updating that CSS will be done by the new `KeyboardNavigationController`.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -885,10 +885,6 @@ export class Navigation {
       workspace.keyboardAccessibilityMode
     ) {
       workspace.keyboardAccessibilityMode = false;
-      workspace.getCursor()?.hide();
-      if (this.getFlyoutCursor(workspace)) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Ready for review now that https://github.com/google/blockly/issues/8962 landed.

maribethb (approved):
Are we planning to keep these enable/disable methods? I thought we didn't want them. Or maybe we want them in the plugin but makecode will just never use them?

rachel-fenichel:
I think we will delete them entirely as a companion to https://github.com/google/blockly/pull/8924.

PR #491, opened 2025-05-05
refactor: Update to use `INavigable` instead of `ASTNode`.

gonfunko:
This PR depends on https://github.com/google/blockly/pull/8992 in core. It updates the keyboard experiment to use the new `INavigable` interface rather than ASTNode. It also temporarily disables the workspace cursor, as it is incompatible with the new approach.

The `FieldColour` plugin will need to be updated; in the meantime, navigating to a block with a colour field will break. I believe this is also the cause of the remaining test failures.

### DIFF
--- a/src/actions/action_menu.ts
+++ b/src/actions/action_menu.ts
@@ -15,7 +15,7 @@ const createSerializedKey = ShortcutRegistry.registry.createSerializedKey.bind(
 );
 
 /**
- * Keyboard shortcut to show the action menu on Cmr/Ctrl/Alt+Enter key.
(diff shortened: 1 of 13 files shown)


rachel-fenichel:
Looks reasonable and the code feels way cleaner without all the extra casts.

Please use `.skip` on the failing tests and file a bug to fix them when the colour field works--I'd like to have tests passing on head so we can notice deltas when changes break new ones.

RoboErikG:
I'm working on getting tests fixed in https://github.com/google/blockly-keyboard-experimentation/pull/496 so please actually just leave the tests alone to avoid merge conflicts. The build errors are blocking for me, though.

Issue #488, opened 2025-05-05
Ensure keyboard shortcuts work with all focused node types


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Request

- Update keyboard shortcuts to work with focused node parameter and not just `getSelected`
- Ensure keyboard shortcuts work correctly when focused node is of different types (e.g. copy on a connection)

Issue #480, opened 2025-04-29
Consider aligning duplicate with insert and copy/paste


microbit-matt-hillsdon:
"Duplicate" is internally implemented as copy then paste, but the effect of the keyboard navigation plugin overrides for copy/paste but not for duplicate is for these two behaviours to drift apart when this plugin is in use.

Copy+paste inserts relative to the current cursor position and on #479 I propose paste should also paste in move mode. Duplicate just drops the block on the workspace offset from the current position.

Not sure if this should be another action override or a different approach that enables core to reuse the plugin's copy/paste behaviours.

RoboErikG:
We'll also want to consider the behavior with a mouse. It might be very confusing for someone if they click a block, try to copy & paste it, and end up in a keyboard move with the mouse being ignored.

I'm not sure what the right behavior would be in this case or if it makes sense to have both copy/paste and duplicate.  🤔

microbit-matt-hillsdon:
Yeah, I'm imagining this is for keyboard uses of the context menu entry and potentially a future keybinding directly for duplicate itself.

microbit-matt-hillsdon:
We discussed this some more based on user testing feedback.

Motivations for change based on user testing:
- Copy/paste or duplicate is often to use the block elsewhere (easier than finding in toolbox). Duplicate can't target the elsewhere. Paste can, but often the user will copy then paste without intending an insert location. Resources often suggest a duplicate then move approach
- Connecting on paste can pop another block out or confusingly nest blocks
- Concern that connecting on paste combined with on-by-default keyboard nav might be an unexpected change for Blockly users (this is something I hadn't previously considered but we should review carefully).

## Approach A

What if:
- Paste, duplicate and connect all dropped a block into move mode

Exposes mixed mode users to move mode. Potentially could depend on a keyboard nav heuristic? But it's a big behaviour shift. Feedback from Blockly on PR that proposed this that in other apps paste is in a fixed position so should be in Blockly too.

Avoids pop-out due to connection. Will still show previews of pop-outs that might be a bit weird.

## Approach B

What if:
- Paste didn't connect (dropped on workspace as before)
- Duplicate remained as is (doesn't connect)
- Each operation popped a first-time toast explaining move mode
    - Problem: toast would be visible to mixed-mode users (could use keyboard nav heuristic?)

Closer to mouse behaviours. 

[...]

PR #501, opened 2025-05-07
feat: delegate clipboard actions to core

### LINKED ISSUE #237: Cut/copy/paste actions hidden unavailable from right-click
https://github.com/google/blockly-keyboard-experimentation/pull/234 overrides the normal cut/copy/paste preconditions to also include the preconditions used for keyboard shortcuts. 

This includes checking whether a gesture is happening (with `!Gesture.inProgress()` and canceling if there's a gesture.

Right-clicking is a gesture and therefore the menu option is shown as disabled on blocks and completely hidden on the workspace.

The fix is to go through the uses of `Gesture.inProgress()` in `clipboard.ts` and make the gesture part only apply for the keyboard shortcut, rather than being part of the shared precondition.

### LINKED ISSUE #500: keyboard nav + cross-tab-copy-paste
### Check for duplicates

### Problem

_No response_

### Request

The plugin needs to be compatible with cross-tab-copy-paste, or other custom implementations of copy/paste, somehow

### Alternatives considered

_No response_

### Additional context

_No response_

maribethb:
Works on #488 
Fixes #237 
Fixes #500

This PR removes all of the custom copy/paste logic and simply delegates the responsibility for copy/pasting to core, or the currently registered copy/paste functions. This simplifies the plugin, and makes it possible to use the cross-tab-copy-paste plugin as long as that plugin registers its replacement shortcuts under the same shortcut names that core uses.
The only custom logic for keyboard shortcuts is that we show/hide toasts on clipboard actions, and we paste in the main workspace instead of flyout workspaces.

We add the context menu options but those also largely use the implementations from core. We can't totally reuse the preconditionFn from core for the keyboard shortcuts, because we can't use the check if a gesture is in progress (it always is during a right click/context menu being open). We can also handle additional places of showing these options in the context menu, such as showing paste on the workspace's menu and not just a block's.

Also note that I decreased the toast time from almost 2 hours (7000 seconds) to 7 seconds.

Adopting this approach would mean that any edge cases around what to do when you try to copy a connection for example would need to be added in core, not in the plugin.

One downside to this approach is that the "paste" option appears enabled even if you have never copied anything. The core implementation does not expose this information so we can't check it, and we can't use the precondition from the keyboard shortcut because that checks if there's a gesture in progress. However, it may be possible to expose the clipboard data from core. Doing so would probably be semi-ugly (I sort of hate that the clipboard data is just held in the keyboard shortcut items file instead of being its own class).

And finally, note that until https://github.com/google/blockly/issues/8793 is fixed, this PR doesn't handle cut behavior correctly. After cutting a block, nothing is focused, so you have to click the workspace again before you can paste or do something else with the workspace.

@microbit-matt-hillsdon I would appreciate if you could take a look and let me know if you think this approach would work with the makecode special pasting logic.

### DIFF
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -238,8 +238,10 @@ export class Clipboard {
         return 'disabled';
       },
       callback: (scope, menuOpenEvent) => {
-        const ws = scope.block?.workspace;
-        if (!ws) return;
+        if (!isCopyable(scope.focusedNode)) return false;
+        const ws = scope.focusedNode.workspace;
+        if (!(ws instanceof WorkspaceSvg)) return false;
+
         return this.copyCallback(ws, menuOpenEvent, undefined, scope);
       },
       id: 'blockCopyFromContextMenu',
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -168,7 +168,9 @@ export class Clipboard {
   private cutCallback(
     workspace: WorkspaceSvg,
     e: Event,
-    shortcut: ShortcutRegistry.KeyboardShortcut | undefined,
+    shortcut: ShortcutRegistry.KeyboardShortcut = {
+      name: Constants.SHORTCUT_NAMES.CUT,
+    },
     scope: ContextMenuRegistry.Scope,
   ) {
     const didCut =
@@ -260,7 +262,9 @@ export class Clipboard {
   private copyCallback(
     workspace: WorkspaceSvg,
     e: Event,
-    shortcut: ShortcutRegistry.KeyboardShortcut | undefined,
+    shortcut: ShortcutRegistry.KeyboardShortcut = {
+      name: Constants.SHORTCUT_NAMES.CUT,
+    },
     scope: ContextMenuRegistry.Scope,
   ) {
     const didCopy =
@@ -392,10 +396,14 @@ export class Clipboard {
   private pasteCallback(
     workspace: WorkspaceSvg,
     e: Event,
-    shortcut: ShortcutRegistry.KeyboardShortcut | undefined,
+    shortcut: ShortcutRegistry.KeyboardShortcut = {
+      name: Constants.SHORTCUT_NAMES.CUT,
+    },
     scope: ContextMenuRegistry.Scope,
   ) {
-    const didPaste = !!this.oldPasteCallback && this.oldPasteCallback();
+    const didPaste =
+      !!this.oldPasteCallback &&
+      this.oldPasteCallback(workspace, e, shortcut, scope);
 
     // Clear the paste hints regardless of whether something was pasted
     // Some implementations of paste are async and we should clear the hint
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -6,28 +6,22 @@
 
 import {
   ContextMenuRegistry,
-  Gesture,
   ShortcutRegistry,
-  Events,
-  Msg,
-  utils as blocklyUtils,
-  clipboard,
   ICopyData,
-  LineCursor,
-  FocusableTreeTraverser,
-  getFocusManager,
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
> Also note that I decreased the toast time from almost 2 hours (7000 seconds) to 7 seconds.

Amazing. I hadn't spotted that the unit had changed when it moved into core.

src/actions/clipboard.ts
```diff
@@ -335,54 +360,62 @@ export class Clipboard {
...
-    Events.setGroup(true);
-    const block = clipboard.paste(this.copyData, pasteWorkspace) as BlockSvg;
-    if (block) {
-      if (targetNode) {
-        this.navigation.tryToConnectBlock(targetNode, block);
-      }
-      Events.setGroup(false);
-      return true;
-    }
-    Events.setGroup(false);
```
microbit-matt-hillsdon, line 387: This removed part doesn't have an equivalent so the behaviour has reverted to the core behaviour that doesn't connect.

I'm not sure this is a bad thing - see discussion on #480 and recent user testing feedback - but checking whether it's intended here.

maribethb: Yes, I read #480 and figured that staying consistent with the behavior in core made sense, and if we wanted to change it, we could change it in core. It should be easy to do a one-time move mode toast, I should get to the "keyboard mode" heuristic soon and we could use that.

src/actions/clipboard.ts
```diff
@@ -80,43 +88,69 @@ export class Clipboard {
...
+          !workspace.isReadOnly() &&
+          isDeletable(focused) &&
+          focused.isDeletable() &&
+          isDraggable(focused) &&
+          focused.isMovable() &&
+          !focused.workspace.isFlyout
+        )
+          return 'enabled';
+
+        return 'disabled';
```
gonfunko: Could this be factored out since it appears to be shared with copy?

src/actions/clipboard.ts
```diff
@@ -125,93 +159,96 @@ export class Clipboard {
...
       },
-      callback: (scope) => {
-        const ws = scope.block?.workspace;
-        if (!ws) return;
-        return this.copyCallback(ws);
+      callback: (scope, menuOpenEvent) => {
+        if (!isCopyable(scope.focusedNode)) return false;
```
gonfunko, line 242: Can this be moved to the precondition?

gonfunko (approved):

src/actions/clipboard.ts
maribethb: it is also in the precondition, but it is here as well because if the thing is `ICopyable` it will definitely have a `workspace` property, so without that we'd have to cast in order to check the property.

src/actions/clipboard.ts
maribethb: Done

maribethb:
Updated to v12 release and addressed Aaron's suggestion

gonfunko (approved):

microbit-robert:
This has allowed us to implement MakeCode's cross tab cut / copy / paste shortcuts so no further changes are suggested for launch, however, there was a bit of work involved. 

- change the timing of the keyboard nav plugin initialisation to ensure that MakeCode's shortcut overrides happen first
- initialise the keyboard nav plugin which wraps these shortcuts and adds items to the context menu
- then override the added context menu preconditions by the plugin to match MakeCode's current customisations and to only show paste as enabled when there is clipboard data (since we have access to this in MakeCode)

This is probably not that bad by itself, but made worse by the fact that the using the keyboard nav plugin is a user setting, so all of the above happens optionally with alternatives where the plugin isn't used. 

Not sure if there's anything that can change to make this more straightforward, but I think we'd support having the clipboard exposed in core so that the plugin can determine if paste should be enabled as this would help most users with simpler cases.

PR #502, opened 2025-05-08
fix: Support now-focusable bubbles and icons.

BenHenning:
This updates `MoveIcon` and `MoveIndicatorBubble` to be non-focusable per the changes being introduced in https://github.com/google/blockly/pull/9004.


BenHenning:
FYI I think this can technically be merged before https://github.com/google/blockly/pull/9004 since JavaScript, but I plan to wait until that PR is merged since I want to make sure there are no additional API changes that would just break this again.

rachel-fenichel (approved):

PR #509, opened 2025-05-10
chore: Fix and enable field navigation unit tests.

johnnesky:
Many test methods had been disabled. In fact, only one of them was actually running (because it called `suite.only()`). I re-enabled most of the tests, adapted them to the current behavior, and temporarily reverted one line in `test_setup.ts` to `callgetCursor()?.setCurNode(field)` instead of `field.getFocusableElement()?.focus()` because the latter doesn't work yet and I figure it would be more useful for the tests to be functional and enabled.

EDIT: Whoops, it looks like existing code in this repo was broken by https://github.com/google/blockly/pull/9016 so github's automated test run fails for reasons that aren't related to my changes.

### DIFF
--- a/test/webdriverio/test/insert_test.ts
+++ b/test/webdriverio/test/insert_test.ts
@@ -15,7 +15,7 @@ import {
   testSetup,
 } from './test_setup.js';
 
-suite.only('Insert test', function () {
+suite('Insert test', function () {
   // Setting timeout to unlimited as these tests take longer time to run
   this.timeout(0);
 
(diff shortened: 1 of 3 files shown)


test/webdriverio/test/insert_test.ts
```diff
@@ -15,7 +15,7 @@ import {
   testSetup,
 } from './test_setup.js';
 
-suite.only('Insert test', function () {
```
microbit-matt-hillsdon, line 18: Oops!

Would need a separate run step for CI vs local to sanity check this but might be worth it.

https://github.com/mochajs/mocha/issues/5299

RoboErikG:
https://github.com/google/blockly-keyboard-experimentation/pull/512 should fix the build errors you're seeing.

RoboErikG (approved):

PR #510, opened 2025-05-12
fix: make insert compile against rc/v12.0.0

microbit-matt-hillsdon:
This uses the same logic as the precondition above and similar callbacks in clipboard.ts.

It builds against beta.5 and rc/v12.0.0 whereas the previous code relied on an `any` that was removed in https://github.com/google/blockly/pull/9016 (so only builds against beta.5).

I'm a bit hazy on whether the plugin is targeting beta.5 or a future release (I suspect the latter) but this change feels like progress either way.

### DIFF
--- a/src/actions/insert.ts
+++ b/src/actions/insert.ts
@@ -88,9 +88,13 @@ export class InsertAction {
         return this.insertPrecondition(ws) ? 'enabled' : 'hidden';
       },
       callback: (scope: ContextMenuRegistry.Scope) => {
-        const ws =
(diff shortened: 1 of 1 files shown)


BenHenning (approved):
Thanks @microbit-matt-hillsdon! LGTM, though it looks like there's one conflict to resolve.

BenHenning:
Ah, there may perhaps be not much left to do in this PR now that #512 is merged.

rachel-fenichel:
Closed in favor of #512

PR #516, opened 2025-05-12
refactor: Don't use the cursor to find connection positions.

gonfunko:
This PR fixes a problem that would be introduced by https://github.com/google/blockly/pull/9030: since the cursor won't visit connections, using the cursor to find connections when inserting a block no longer works. Instead, connections are just queried for directly.


RoboErikG (approved):
Good catch!

rachel-fenichel (approved):
Approved; please file an issue to take another look at the search code to see if we can use the connection db instead.

PR #520, opened 2025-05-13
refactor: Update to use `IFocusableNode` in place of `INavigable`.

gonfunko:
This PR updates the keyboard experiment for compatibility with https://github.com/google/blockly/pull/9037 by using IFocusableNode in place of INavigable.

### DIFF
--- a/test/webdriverio/test/basic_test.ts
+++ b/test/webdriverio/test/basic_test.ts
@@ -44,13 +44,12 @@ suite('Keyboard navigation on Blocks', function () {
       await this.browser.pause(PAUSE_TIME);
     }
 
-    const selectedId = await this.browser.execute(() => {
-      return Blockly.common.getSelected()?.id;
-    });
-    chai.assert.equal(selectedId, 'controls_repeat_1');
+    chai
+      .expect(await getCurrentFocusNodeId(this.browser))
+      .to.include('controls_if_2');
   });
 
-  test('Down from statement block selects next connection', async function () {
+  test('Down from statement block selects next block across stacks', async function () {
     await tabNavigateToWorkspace(this.browser);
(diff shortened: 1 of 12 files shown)


rachel-fenichel:
`FlyoutCursor` needs updates (or removal?) before this can build.

gonfunko:
Already had them, just needed to be re-run after checking in the core change.

RoboErikG:
I think the failing tests are from the change in behavior to skip connections. You can double check by increasing the pause timeout to watch it go and then update the expected end position.

test/webdriverio/test/basic_test.ts
```diff
@@ -44,13 +44,12 @@ suite('Keyboard navigation on Blocks', function () {
       await this.browser.pause(PAUSE_TIME);
     }
 
-    const selectedId = await this.browser.execute(() => {
-      return Blockly.common.getSelected()?.id;
-    });
-    chai.assert.equal(selectedId, 'controls_repeat_1');
+    chai
+      .expect(await getCurrentFocusNodeId(this.browser))
+      .to.include('controls_if_2');
```
RoboErikG: I think this be an assert.equal here and elsewhere. The include is only needed for connections and fields because they get a generated string appended to their id, but blocks the whole id is known so we can use equal.

RoboErikG (approved):
Approved. Feel free to submit as is or make the proposed tweak to use assert.equal instead. The .to.include call is only needed for fields and connections since they get a generated string appended to their id. This means technically any of the sub-elements of the block would also match the .include version and pass the test.

maribethb (approved):

test/webdriverio/test/basic_test.ts
gonfunko: Done.

PR #530, opened 2025-05-16
Finish drag on focus change

### LINKED ISSUE #444: Click outside workspace during keyboard move causes stuck move
1. Start a move of "draw circle"
2. Click outside the workspace/toolbox
3. Observe passive focus indicator
4. Fail to resume the move when focus is regained

Perhaps the move should confirm or abort (depending on what we agree elsewhere)?

RoboErikG:
Fixes #444 

Adds a blur listener to the block's element during a keyboard move that ends the move if focus is lost.


rachel-fenichel (approved):
Looks good--can you also add a test for this to guard against regressions?

PR #531, opened 2025-05-17
chore: Add tests for focus after deletion.

johnnesky:
Implements unit tests described at https://docs.google.com/document/d/1pIIX5sKEG9f_NcWDIQJJEOMM7w7pQUf8eJoT3lrAElQ/edit?tab=t.0 for deleting blocks via keyboard commands.

The document linked above expects that deleting a block that's inside an inline input would result in the parent block being selected, and that's not what happens, instead the parent block's input connection is selected. This PR expects the currently implemented behavior.

### DIFF
--- a/test/webdriverio/test/delete_test.ts
+++ b/test/webdriverio/test/delete_test.ts
@@ -19,7 +19,7 @@ import {
 } from './test_setup.js';
 import {Key} from 'webdriverio';
 
-suite.only('Deleting Blocks', function () {
+suite('Deleting Blocks', function () {
   // Setting timeout to unlimited as these tests take a longer time to run than most mocha test
   this.timeout(0);
 
--- a/test/webdriverio/test/delete_test.ts
+++ b/test/webdriverio/test/delete_test.ts
@@ -1,0 +1,222 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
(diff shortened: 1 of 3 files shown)


test/webdriverio/test/delete_test.ts
```diff
@@ -0,0 +1,212 @@
...
+  getFocusedBlockType,
+  moveToToolboxCategory,
+  testSetup,
+  testFileLocations,
+  PAUSE_TIME,
+  tabNavigateToWorkspace,
+} from './test_setup.js';
+import {Key} from 'webdriverio';
+
+suite.only('Deleting Blocks', function () {
```
rachel-fenichel: Remove `.only` so this doesn't disable other tests.

rachel-fenichel (commented):
Tests look good--please fix lint and remove `.only` to be able to merge.

> The document linked above expects that deleting a block that's inside an inline input would result in the parent block being selected, and that's not what happens, instead the parent block's input connection is selected. This PR expects the currently implemented behavior.

Works for me. ~Cursor position after delete is discussed https://github.com/google/blockly-keyboard-experimentation/issues/254 and we can update the test to match the behaviour we agree on in discussion.~

Just realized that other issue is currently closed. I meant https://github.com/google/blockly/issues/8793

PR #532, opened 2025-05-19
feat: Allow visiting and activating icons via keyboard navigation.

### LINKED ISSUE #189: Blockly Icons are not keyboard navigable
Blockly Icons are used for comments which are currently not keyboard navigable. The context menu can be used to add and remove comments, but it is only possible to edit and view comments via mouse clicks.

Image

Another use case in the test page in the `if-do` block:

Image

A third-party use case for these icons are as breakpoint on/off toggles for the debugger mode in MakeCode. There is a potential complication for this case, as the blocks are readonly in this mode, but we'd still want icon interaction.

Image

gonfunko:
In conjunction with https://github.com/google/blockly/pull/9072, this PR partly fixes #189. It allows icons to be visited and activated. Support for moving focus to icon bubbles will be forthcoming in future PRs. Navigation logic is tested in the core PR.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -14,6 +14,7 @@ import {
   RenderedConnection,
   WorkspaceSvg,
   Field,
+  icons,
   FocusableTreeTraverser,
 } from 'blockly/core';
 
@@ -106,6 +107,8 @@ export class EnterAction {
       curNode instanceof WorkspaceSvg
     ) {
       this.navigation.openToolboxOrFlyout(workspace);
+    } else if (curNode instanceof icons.Icon) {
+      curNode.onClick();
     }
   }
 
--- a/test/index.html
+++ b/test/index.html
@@ -104,7 +104,10 @@
         .blocklyActiveFocus:is(.blocklyPath, .blocklyHighlightedConnectionPath),
       .blocklyKeyboardNavigation
         .blocklyActiveFocus.blocklyField
-        > .blocklyFieldRect {
+        > .blocklyFieldRect,
+      .blocklyKeyboardNavigation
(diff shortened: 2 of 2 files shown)


src/actions/enter.ts
```diff
@@ -106,6 +107,8 @@ export class EnterAction {
       curNode instanceof WorkspaceSvg
     ) {
       this.navigation.openToolboxOrFlyout(workspace);
+    } else if (curNode instanceof icons.Icon) {
+      curNode.onClick();
```
BenHenning, line 111: I wonder if this should perhaps focus the icon, instead. Presumably we could then update icons' `onNodeFocus` implementations to their bubbles as needed.

BenHenning (commented):
Out of curiosity, which icons should now be navigable @gonfunko? I tried testing this + the Core change locally and I can't seem to navigate to mutator or comment icons (I didn't check warning icons).

src/actions/enter.ts
gonfunko: The icon is presumably already focused if this is being triggered though, no?

BenHenning: Ah you're right--my comment doesn't make sense.

Calling `onClick` does feel a bit weird, though. I suppose having an actual "doAction"-esque function would be a bit cleaner than `onClick` (since this is masquerading as an event provider a bit), but I don't have strong opinions here.

BenHenning (approved):
I'm not sure what changed but everything seems working for me in testing this morning.

src/actions/enter.ts
gonfunko: Yeah, I agree the naming is suboptimal but that feels like something more easily sorted out when we move this back into core.

PR #537, opened 2025-05-21
fix(mover): Don't heal rows when disconnecting at start of move

### LINKED ISSUE #456: Value block keyboard move moves fewer blocks than expected in some cases
This video starts a variety of value block moves then presses enter to put the block back where it came from.

Most behave as I'd expect, but for `sin` and `not`, less is "picked up" than I expected. The asdf variable and true block are left behind. This is inconsistent with mouse drag.

https://github.com/user-attachments/assets/9ff10172-0dad-4979-bc78-5c8a8e10dfb1

cpcallen:
Modify `KeyboardDragStrategy.prototype.shouldHealStack` so that only stacks will be healed.

Fixes #456.

Also add tests for starting to move both stack and value blocks, with the value blocks being both single-value-input and multiple-value-input.

### DIFF
--- a/test/webdriverio/test/move_test.ts
+++ b/test/webdriverio/test/move_test.ts
@@ -30,7 +30,7 @@ suite('Move tests', function () {
   // moved, with subsequent statement blocks below it in the stack
   // reattached to where the moving block was - i.e., that a stack
   // heal will occur.
-  test.only('Start moving statement blocks', async function () {
+  test('Start moving statement blocks', async function () {
     for (let i = 1; i < 7; i++) {
       // Navigate to statement_<i>.
       await tabNavigateToWorkspace(this.browser);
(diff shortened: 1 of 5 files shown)


gonfunko (approved):

test/webdriverio/test/move_test.ts
```diff
@@ -0,0 +1,224 @@
...
+    this.browser = await testSetup(testFileLocations.MOVE_TEST_BLOCKS);
+    await this.browser.pause(PAUSE_TIME);
+  });
+
+  // When a move of a statement block begins, it is expected that only
+  // that block (and all blocks connected to its inputs) will be
+  // moved, with subsequent statement blocks below it in the stack
+  // reattached to where the moving block was - i.e., that a stack
+  // heal will occur.
+  test.only('Start moving statement blocks', async function () {
```
rachel-fenichel: Remove `.only` so that you don't disable the rest of the tests.

rachel-fenichel (changes requested):

test/webdriverio/test/move_test.ts
rachel-fenichel: Same below.

rachel-fenichel (approved):

PR #542, opened 2025-05-22
feat: align disconnect with move + mouse behaviours

microbit-matt-hillsdon:
- Single statement block to match move. Perhaps both could do multi-block with uppercase.
- Just call unplug like the drag strategy does.

This means the disconnected block ends up the current node. Previous (presumably accidental) behaviour was to select the connection in a way that can't normally be navigated to causing it to persist oddly.

You could make a case for move mode too but it's all tangled up with the insert/paste/heuristic discussion so not considering here.

Demo: https://simpler-disconnect.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/actions/disconnect.ts
+++ b/src/actions/disconnect.ts
@@ -6,12 +6,14 @@
 
 import {
   BlockSvg,
-  RenderedConnection,
+  Events,
   ShortcutRegistry,
   utils as BlocklyUtils,
+  Connection,
+  ConnectionType,
 } from 'blockly';
 import * as Constants from '../constants';
-import type {WorkspaceSvg, IFocusableNode} from 'blockly';
+import type {WorkspaceSvg} from 'blockly';
 import {Navigation} from '../navigation';
 
 const KeyCodes = BlocklyUtils.KeyCodes;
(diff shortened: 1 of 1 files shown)


src/actions/disconnect.ts
```diff
@@ -76,66 +78,15 @@ export class DisconnectAction {
...
-    superiorConnection.disconnect();
-    inferiorConnection.bumpAwayFrom(superiorConnection);
+    if (!cursor) return;
+    const curNode = cursor.getCurNode();
+    if (!(curNode instanceof BlockSvg)) return;
 
-    const rootBlock = superiorConnection.getSourceBlock().getRootBlock();
-    rootBlock.bringToFront();
+    const healStack = !curNode.outputConnection?.isConnected();
+    Events.setGroup(true);
```
rachel-fenichel, line 86: Set the event group to false afterwards.

rachel-fenichel:
Overall I'm happy with this--matching existing behaviour means we use well-tested codepaths.

src/actions/disconnect.ts
microbit-matt-hillsdon: 79a1c22db79b3daf26770913f02424b1d335c6f8

rachel-fenichel (approved):

PR #549, opened 2025-05-23
release: bump version to 0.0.9

rachel-fenichel:
Updates the dependency on Blockly to v12.0.1-beta.1 to pull in some recent Blockly fixes, and bumps the version.

Purpose: release another version for testing by MakeCode and Code.org/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@blockly/keyboard-experiment",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@RoboErikG this is now ready for review. I had to add the `overrides` section back for `field-colour`, despite Christopher having just removed it. 

The `overrides` stanza is necessary because of this part of [how npm handles semver](https://www.npmjs.com/package/semver): "Prerelease Tags
If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy comparator sets if at least one comparator with the same [major, minor, patch] tuple also has a prerelease tag.

RoboErikG (approved):

Issue #546, opened 2025-05-23
Can no longer use keyboard actions on blocks in the flyout


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

At one point, we were able to use Ctrl + Enter to open the context menu on blocks in the flyout, and Ctrl + C to copy blocks in the flyout (the latter automatically closed the flyout). This is no longer case. This feels inconsistent with mouse users who can still open the context menu via right click. Copy is still not available when opening the context menu via mouse, but perhaps it should be?

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

RoboErikG:
Opening the menu now works, but copy still needs fixing. The plan is to fix copy/cut/paste in core using the following checks:

- Copy: precondition returns true if block.isOwnEditable (this will return true in the flyout)
- Cut: precondition returns true if block.isEditable (this will only return true on an editable workspace)
- Paste: precondition returns true if copyData exists and the workspace is editable or workspace.targetWorkspace is editable (in which case it pastes to the target workspace).

PR #550, opened 2025-05-23
fix: Fix when the context menu is allowed to open

### LINKED ISSUE #276: Context menu should work via keyboard in the flyout
The context menu should work via keyboard in the flyout in the same way as in the regular workspace (some actions may have different enablement, but it should be possible to trigger the menu).

At the moment if you trigger by mouse you get Help and Cut/Copy/Paste. It unhelpfully closes the flyout due to #200 but the menu does work.

MakeCode has additional context menu entries for variables:

Image

### LINKED ISSUE #434: Can trigger context menu in keyboard move mode

RoboErikG:
Fixes #434 
Fixes #276

#546 is partly fixed by this change but may still need some work for shortcuts other than menu opening.

This fixes the precondition checks for showing the action menu and also enables the action menu to be shown in the flyout so that it's consistent with right click.

### DIFF
--- a/test/webdriverio/test/actions_test.ts
+++ b/test/webdriverio/test/actions_test.ts
@@ -7,7 +7,7 @@
 import * as chai from 'chai';
 import {Key} from 'webdriverio';
 import {
-  checkActionPrecondition,
+  contextMenuExists,
   moveToToolboxCategory,
   PAUSE_TIME,
   setCurrentCursorNodeById,
@@ -26,12 +26,15 @@ suite('Menus test', function () {
     await this.browser.pause(PAUSE_TIME);
   });
 
-  test('Menu action returns true normally', async function () {
+  test('Menu action opens menu', async function () {
     // Navigate to draw_circle_1.
     await tabNavigateToWorkspace(this.browser);
     await setCurrentCursorNodeById(this.browser, 'draw_circle_1');
+    await this.browser.pause(PAUSE_TIME);
+    await this.browser.keys([Key.Ctrl, Key.Return]);
+    await this.browser.pause(PAUSE_TIME);
     chai.assert.isTrue(
-      await checkActionPrecondition(this.browser, 'menu'),
+      await contextMenuExists(this.browser, 'Duplicate'),
       'The menu should be openable on a block',
     );
   });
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
Assigning @BenHenning to review because I don't understand the workspace state changes.

rachel-fenichel:
Also fixes (?) https://github.com/google/blockly-keyboard-experimentation/issues/276

RoboErikG:
It partly fixes 276 by allowing the context menu to open in the flyout, but making copy/paste work correctly requires fixing some things in core or undoing Maribeth's change to depend on the core callbacks.

RoboErikG:
I mixed up bugs, this does fix #276 and partially fixes #546. Updated the description to match.

BenHenning (commented):
Thanks @RoboErikG. I think the solution makes sense. Is it possible add tests for these cases?

test/webdriverio/test/test_setup.ts
```diff
@@ -431,3 +431,40 @@ export async function isDragging(
...
+    }
+    const actionItem = Blockly.ShortcutRegistry.registry.getRegistry()[action];
+    if (!actionItem || !actionItem.preconditionFn) {
+      throw new Error(
+        `No registered action or missing precondition: ${action}`,
+      );
+    }
+    return actionItem.preconditionFn(workspace, {
+      focusedNode:
+        Blockly.FocusManager.getFocusManager().getFocusedNode() ?? undefined,
```
BenHenning: Can `node` from above be reused here? I think technically it's guaranteed to be defined at this point, too, but you could always amend the conditional on line 456 to check for node's truthiness to placate the compiler (if needed).

test/webdriverio/test/actions_test.ts
```diff
@@ -0,0 +1,65 @@
...
+    // Navigate to draw_circle_1.
+    await tabNavigateToWorkspace(this.browser);
+    await setCurrentCursorNodeById(this.browser, 'draw_circle_1');
+    // Navigate to a toolbox category
+    await moveToToolboxCategory(this.browser, 'Functions');
+    // Move to flyout.
+    await this.browser.keys(Key.ArrowRight);
+
+    chai.assert.isTrue(
+      await checkActionPrecondition(this.browser, 'menu'),
```
BenHenning: Is it possible to actually open the menu and verify that it's open rather than checking the precondition? That would be a lot closer to what the user experiences (and thus perhaps a bit more robust against implementation changes and refactors that don't change the end user behavior).

BenHenning (approved):
Thanks @RoboErikG! Approving to unblock, but I did have some thoughts on the tests. Feel free to re-add me if you'd like another review pass.

PR #557, opened 2025-05-29
chore: Introduce tests for flyout & toolbox

BenHenning:
Fixes part of https://github.com/google/blockly/issues/8915
Fixes part of https://github.com/google/blockly/issues/9020

This PR does a few things:
- It introduces some new helpers to simplify arrow key inputs across all of the webdriver tests.
- It introduces a new focusable div element that exists in the tab order before toolbox (so that back navigation can be tested).
- It renames a few functions for clarity: `setCurrentCursorNodeById` -> `focusOnBlock` and `setCurrentCursorNodeByIdAndFieldName` -> `focusOnBlockField`. This is closer to what the functions are actually doing, and it moves away from cursor verbiage (which could become confusing in the future once the cursor is removed).
- It introduces some robustness sanity checks for test utility functions. These should fail if an assumption isn't met rather than return null or undefined--failing fast is really useful in tests to avoid hiding legitimate failures.
- It introduces a whole test suite for toolbox and flyout (though a lot more tests can be added). This specifically emphasizes regressions fixed by https://github.com/google/blockly/pull/9079.

### DIFF
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -18,7 +18,7 @@ import {
   keyRight,
 } from './test_setup.js';
 
-suite.only('Toolbox and flyout test', function () {
+suite('Toolbox and flyout test', function () {
   // Clear the workspace and load start blocks
   setup(async function () {
     this.browser = await testSetup(testFileLocations.BASE);
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -11,15 +11,15 @@ import {
   testFileLocations,
   PAUSE_TIME,
   tabNavigateForward,
-  getFocusedBlockType,
   keyDown,
   tabNavigateBackward,
   tabNavigateToWorkspace,
   keyRight,
   getCurrentFocusNodeId,
+  getCurrentFocusedBlockId,
 } from './test_setup.js';
 
-suite('Toolbox and flyout test', function () {
+suite.only('Toolbox and flyout test', function () {
   // Clear the workspace and load start blocks
   setup(async function () {
     this.browser = await testSetup(testFileLocations.BASE);
@@ -46,8 +46,8 @@ suite('Toolbox and flyout test', function () {
 
     // The top block of the category should be automatically selected. See:
     // https://github.com/google/blockly/issues/8978.
-    const blockType = await getFocusedBlockType(this.browser);
-    chai.assert.strictEqual(blockType, 'controls_if');
+    const blockId = await getCurrentFocusedBlockId(this.browser);
+    chai.assert.strictEqual(blockId, 'if_block');
   });
 
   test('Tab navigating to toolbox then right arrow key should auto-select first block in flyout', async function () {
@@ -60,8 +60,8 @@ suite('Toolbox and flyout test', function () {
 
     // The top block of the category should be automatically selected. See:
     // https://github.com/google/blockly/issues/8978.
-    const blockType = await getFocusedBlockType(this.browser);
-    chai.assert.strictEqual(blockType, 'controls_if');
+    const blockId = await getCurrentFocusedBlockId(this.browser);
+    chai.assert.strictEqual(blockId, 'if_block');
   });
 
   test('Keyboard nav to different toolbox category should auto-select first block', async function () {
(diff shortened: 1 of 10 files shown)


BenHenning (commented):
Basic self-review (mostly a quick spot check).

BenHenning:
This can be reviewed now, but it must wait until https://github.com/google/blockly/pull/9079 is merged before it can also be merged (since it relies on those changes in order for CI to pass).

test/webdriverio/test/test_setup.ts
```diff
@@ -418,6 +420,75 @@ export async function tabNavigateForward(browser: WebdriverIO.Browser) {
...
+ * Sends the specified key for the specified number of times, waiting between
+ * each key press to allow changes to keep up.
+ *
+ * @param browser The active WebdriverIO Browser object.
+ * @param key The WebdriverIO representative key value to press.
+ * @param times The number of times to repeat the key press.
+ */
+async function sendKeyAndWait(
+  browser: WebdriverIO.Browser,
+  key: string,
```
cpcallen, line 483: Any reason not to make this accept everything that `browser.keys` does?  Then you could call it from `tabNavigateForward` and `tabNavigateBackward` (and other places too, if exported).

test/webdriverio/test/flyout_test.ts
```diff
@@ -0,0 +1,290 @@
...
+  // Clear the workspace and load start blocks
+  setup(async function () {
+    this.browser = await testSetup(testFileLocations.BASE);
+    await this.browser.pause(PAUSE_TIME);
+  });
+
+  test('Tab navigating to toolbox should open flyout', async function () {
+    // Two tabs should navigate to the toolbox (initial element is skipped).
+    await tabNavigateForward(this.browser);
+
```
cpcallen, line 32: Delete seemingly spurious blank line?  (Here and on several more test cases below).

test/webdriverio/test/flyout_test.ts
```diff
@@ -0,0 +1,290 @@
...
+    // Note that two tabs are needed here to move past the flyout.
+    await keyDown(this.browser, 3);
+    await tabNavigateBackward(this.browser);
+    await tabNavigateForward(this.browser);
+    await tabNavigateForward(this.browser);
+
+    // The previously selected block should be retained upon returning. See:
+    // https://github.com/google/blockly/issues/8965#issuecomment-2900479280.
+    const blockType = await getFocusedBlockType(this.browser);
+    chai.assert.strictEqual(blockType, 'p5_draw');
```
cpcallen: Maybe I'm being excessively picky, especially since the draw block is supposed to be unique, but since this is a block loaded from the scenario I think I'd check the block's ID rather than type—and to facilitate that, make sure that the JSON being loaded specified a nice human one.

cpcallen (approved):
A few minor suggestions—but definitely acceptable as-is.

rachel-fenichel:
Because this change refactors tests and several in-flight changes add or modify tests, please coordinate with the rest of the team on merge order.

Also, I just updated the PR title to match conventional commits

test/webdriverio/test/flyout_test.ts
BenHenning: These actually aren't spurious. I usually format my tests always into two or three blocks for distinct arrange-act-assert sections (to make it very clear to readers which part of the test is setup, the actual operation being verified, and the code for doing that verifying).

test/webdriverio/test/flyout_test.ts
BenHenning: I think I was noticing that the IDs were being generated and not retained in the workspace for whatever reason, but I will give this a try and follow up.

BenHenning:
> Because this change refactors tests and several in-flight changes add or modify tests, please coordinate with the rest of the team on merge order.
> 
> Also, I just updated the PR title to match conventional commits

I somehow forgot to update the title--nice catch and thanks for fixing.

test/webdriverio/test/test_setup.ts
BenHenning: I'm not opposed, but I think I'll do that in a follow-up if there's no concerns there (since I'd rather move everything over and that's a bit of a scope expansion for this PR).

I also don't think I want to exactly mimic the contract of `browser.keys` since it can do some inadvertent key repeats (which we actually use elsewhere), but allowing it to include meta keys is a good point.

test/webdriverio/test/flyout_test.ts
BenHenning: Looks like this was as simple as just adding IDs to the toolbox JSON. Fixed in latest.

BenHenning:
#9079 is merged and CI is now passing. Just double checking the latest changes before merging this.

BenHenning (commented):
Self-reviewed latest (spot check).

test/webdriverio/test/flyout_test.ts
cpcallen: That totally makes sense, and I approve of the arrangment.

Were it my test I might have had a separate comments for the "arrange" and "act" sections.

PR #560, opened 2025-05-29
onShortcut monkey-patch

microbit-matt-hillsdon:
Possible temporary, plugin-only fix for #437 and https://github.com/google/blockly/issues/9101

Needs discussion before merge but at least lets you play with behaviour.

### DIFF
--- a/src/actions/move.ts
+++ b/src/actions/move.ts
@@ -7,6 +7,7 @@
 import {
   BlockSvg,
   ContextMenuRegistry,
+  FocusManager,
   Msg,
   ShortcutRegistry,
   utils,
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
Demo: https://on-shortcut-patch.blockly-keyboard-experimentation.pages.dev/

rachel-fenichel:
@gonfunko is working on a change for this that doesn't involve a monkeypatch and should be ready today.

maribethb:
Aaron's change is only for the first issue, not the second. I'm working on the second issue though. Going to have to go with the precondition approach because I think not firing them at all if the widget div is open would be a breaking change.

maribethb:
We shouldn't need this due to this behavior being implemented by 

https://github.com/google/blockly/pull/9110
https://github.com/google/blockly-keyboard-experimentation/pull/568 
https://github.com/google/blockly-keyboard-experimentation/pull/564

PR #561, opened 2025-05-29
fix: share abort/finish clean-up so both remove blur listener

### LINKED ISSUE #553: cancelling move mode breaks block dragging
### Check for duplicates

### Description

Cancelling a move mode move with "escape" breaks block dragging. See video

### Reproduction steps

1. Press M to move a block
2. Move it with arrow keys
3. Press Escape to cancel
4. The block is put in the upper left corner, not where it started
5. Try to drag the block with the mouse. It picks up another block that wasn't connected to it, and then you have to put the mouse in a weird spot to get it to connect.
3.

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/85ce9ef7-ea97-4595-a58e-25a8e31548e3

### Browsers

_No response_

microbit-matt-hillsdon:
Previously calling abort called finish via setParent (causing a blur) which caused the strategy to be unpatched too soon.

Fixes #553

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -152,28 +152,14 @@ export class Mover {
    * @returns True iff move successfully finished.
    */
   finishMove(workspace: WorkspaceSvg) {
-    clearMoveHints(workspace);
-
-    const info = this.moves.get(workspace);
-    if (!info) throw new Error('no move info for workspace');
-
-    // Remove the blur listener before ending the drag
-    info.block
-      .getFocusableElement()
-      .removeEventListener('blur', info.blurListener);
+    const info = this.preDragEndCleanup(workspace);
 
     info.dragger.onDragEnd(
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
Demo: https://mover-blur-fix.blockly-keyboard-experimentation.pages.dev/

rachel-fenichel (commented):
Ben is doing the primary review, but: it would help to have a test for this.

BenHenning (approved):
The fix makes sense to me plus the approach is much cleaner.

I spent about 15 minutes trying to write a test to suggest, but unfortunately ran into some issues that would take longer to figure out.

rachel-fenichel:
I'm okay with getting a test as a follow-up. It's a regression guard, as usual. I know that this is a release blocker so I'm fine with merging it as is.

PR #564, opened 2025-05-29
feat: Commit moves before running keyboard shortcuts.

### LINKED ISSUE #534: Confirm / finish keyboard moves on subsequent actions
### Check for duplicates

### Description

It is still possible to trigger other actions while a keyboard move is in progress leading to unexpected results or errors. 

Moves are now confirmed when a user clicks outside of the workspace while a move is in progress (see https://github.com/google/blockly-keyboard-experimentation/pull/530). Other actions should probably follow suit and confirm the move before triggering the next action, this is supported by feedback we have received so far.

Scenarios that error while in move mode:
- Cut (Ctrl + X)
- Opening context menu (Ctrl + Enter)
- Clean up workspace (C) - no error but this can be confusing due to the amount of movement of other blocks

Other actions such as delete have checks in their precondition functions to avoid triggering the action. It would be good to align all of these to have the same behaviour with an approach that will also work for shortcuts added by apps using Blockly via the shortcut registry.

[This PR](https://github.com/google/blockly-keyboard-experimentation/pull/437) suggests a couple of approaches to handling this problem, but there may well be other solutions.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
This PR fixes #534 by registering a keyboard shortcut under the key combos of all existing keyboard shortcuts that commits the in-progress move before returning control to the corresponding real keyboard shortcut. To fully function, this depends on https://github.com/google/blockly/pull/9109; without it, moves will be commited, but the underlying/real shortcut action may not run.


rachel-fenichel (commented):
Please add this test:

- start a move with keys
- press an arrow key twice
- press `c`
- press an arrow key twice
- check that the block is in the expected location

src/actions/mover.ts
```diff
@@ -140,6 +146,47 @@ export class Mover {
...
+      .flatMap((shortcut) => shortcut.keyCodes)
+      .filter((keyCode) => {
+        return (
+          keyCode &&
+          ![
+            utils.KeyCodes.RIGHT,
+            utils.KeyCodes.LEFT,
+            utils.KeyCodes.UP,
+            utils.KeyCodes.DOWN,
+            utils.KeyCodes.ENTER,
```
microbit-matt-hillsdon, line 164: Also needs `KeyCodes.SPACE` from the finish move action alongside `ENTER`.

Potentially these could be pulled from a list of the move actions from `move.ts`? Would be helpful in future as we hope to add UI with the ability to rebind shortcuts (though these ones are perhaps less likely to change than some).

microbit-matt-hillsdon:
Would be great to get #561 in first and clean up this new shortcut in the new common clean-up method, as presumably it should be cleaned up on abort too. Or call in both places here and I can fix conflicts with the other PR.

src/actions/mover.ts
rachel-fenichel: And we do anticipate that when all of the keyboard nav code makes it into core, the shortcut actions will directly cancel moves. This setup is necessary because some shortcuts are in core and some are in the plugin. We will keep UI rebinding in mind for design work on that.

gonfunko: We don't actually need space, because this shortcut finishes the move itself; Enter is there to support command-enter to bring up the contextual menu, not to commit the move. But we do need Escape to abort the move, so I added that in (and cleaned up this shortcut in the new common cleanup method introduced in #561).

rachel-fenichel (approved):
Looks good, and I'm okay with merging as-is since this is a blocker. 

And a tag for https://github.com/google/blockly/issues/8915 in case you don't get to writing that test today.

PR #566, opened 2025-05-29
release: v1.0.0 of @blockly/keyboard-navigation

### LINKED ISSUE #497: Publish under `@blockly/keyboard-navigation`
### Check for duplicates

### Problem

_No response_

### Request

When we're ready to publish the 1.0.0 release, we should publish under the `@blockly/keyboard-navigation` name. That's the name of the old plugin in samples which currently has a 0.6 version number.

This may require jumping through some hoops to change the code location or publish under the `@blockly` namespace, needs investigation.

We should also be cautious that once we publish the plugin for real, any change to the minimum required version of Blockly will be a breaking change to the plugin. Thus, we should avoid making many small paired changes that add APIs to core and then use them in the plugin, requiring a major version bump. Users will not automatically get new major versions and we don't want to leave all of our new users behind on old versions or ask them to update repeatedly.

### Alternatives considered

_No response_

### Additional context

_No response_

maribethb:
Fixes #497 

Updates the package name, version, and minimum version of blockly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@blockly/keyboard-navigation",
-  "version": "1.0.0-beta.0",
+  "version": "1.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@blockly/keyboard-navigation",
-      "version": "1.0.0-beta.0",
+      "version": "1.0.0",
       "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


maribethb:
This will work fine once the blockly version needed isn't a beta so it won't conflict with the colour field. So I think we just leave this be until blockly 12.1.0 is fully published.

rachel-fenichel (approved):

PR #568, opened 2025-05-30
fix: dont do keyboard shortcuts when field editors open

### LINKED ISSUE #9101: Keyboard shortcuts handled with the wrong scope
### Check for duplicates

### Description

There is a [keydown listener on the widget div](https://github.com/google/blockly/blob/develop/core/inject.ts#L303-L318). When we use arrow keys, this listener is triggered. The listener checks to see if the element is a text input or similar and ignores the event if that's the case. Otherwise, the keydown event is passed to the shortcut listener. The shortcut listener attempts to handle the keydown event, and the scope is the currently focused node. This code ignores whether anything has taken ephemeral focus. The focused node is the last field that was focused before the field editor was opened. Therefore, arrow keys move the cursor around the workspace in the background. This is problematic for a complex field editor in MakeCode that uses the widget div.

This doesn't currently happen with dropdown divs, but this is due to another bug where the keydown listener is removed from the dropdown div and not reinstated. I think that when [this PR](https://github.com/google/blockly/pull/9085) goes in, regular dropdown fields will also exhibit this problem.

I think a possible fix for this that the call to [getFocusedNode here](https://github.com/google/blockly/blob/develop/core/shortcut_registry.ts#L262-L265) must take into account ephemeral focus.

### Reproduction steps

The easiest way to reproduce this is probably to run the [blockly-keyboard-experimentation repo](https://github.com/google/blockly-keyboard-experimentation) locally with a linked Blockly repo that has the code changes in https://github.com/google/blockly/pull/9085 checked out.

You can then:
- add a `show icon` block 
- navigate to the icon field with keyboard
- hit 'Enter' to open the field editor
- move the cursor on the workspace in the background with the left and right arrow buttons (minor visible change for the first move, invisible after that)
- hit 'Enter' again to trigger an action whether the cursor now is

See video below for an example of this.

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/fed9b288-d041-446c-8d6c-3b734256532a

### Browsers

_No response_

maribethb:
Fixes https://github.com/google/blockly/issues/9101

Depends on https://github.com/google/blockly/pull/9110

Adds a check to the `canCurrentlyNavigate` function to make sure nothing has ephemeral focus.

This is safe to do because this function is only used by action preconditions.

It's even safe to do for the escape shortcut because that shortcut only applies when the flyout or toolbox is opened, and in that case it's not possible for a field editor to be open. The escape shortcut in core is the one that closes chaff generally (e.g. when field editors are open) and that one does not have this precondition, so it can still be used to close field editors.

### DIFF
--- a/test/webdriverio/test/clipboard_test.ts
+++ b/test/webdriverio/test/clipboard_test.ts
@@ -15,6 +15,8 @@ import {
   ElementWithId,
   tabNavigateToWorkspace,
   focusOnBlock,
+  focusOnBlockField,
+  blockIsPresent,
 } from './test_setup.js';
 import {Key, KeyAction, PointerAction, WheelAction} from 'webdriverio';
 
@@ -109,6 +111,24 @@ suite('Clipboard test', function () {
       initialWsBlocks,
       await serializeWorkspaceBlocks(this.browser),
       'Blocks on the workspace should not have changed',
+    );
+  });
+
+  test('Do not cut block while field editor is open', async function () {
+    // Open a field editor
+    await focusOnBlockField(this.browser, 'draw_circle_1_color', 'COLOUR');
+    await this.browser.pause(PAUSE_TIME);
+    await this.browser.keys(Key.Enter);
+    await this.browser.pause(PAUSE_TIME);
+
+    // Try to cut block while field editor is open
+    await this.browser.keys(Key.Ctrl);
+    await this.browser.keys('x');
+    await this.browser.keys(Key.Ctrl); // release ctrl key
+
+    // Block is not deleted
+    chai.assert.isTrue(
+      await blockIsPresent(this.browser, 'draw_circle_1_color'),
     );
   });
 });
--- a/test/webdriverio/test/clipboard_test.ts
+++ b/test/webdriverio/test/clipboard_test.ts
@@ -122,9 +122,7 @@ suite('Clipboard test', function () {
     await this.browser.pause(PAUSE_TIME);
 
     // Try to cut block while field editor is open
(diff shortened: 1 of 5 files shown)


test/webdriverio/test/clipboard_test.ts
```diff
@@ -111,6 +113,24 @@ suite('Clipboard test', function () {
...
+    // Open a field editor
+    await focusOnBlockField(this.browser, 'draw_circle_1_color', 'COLOUR');
+    await this.browser.pause(PAUSE_TIME);
+    await this.browser.keys(Key.Enter);
+    await this.browser.pause(PAUSE_TIME);
+
+    // Try to cut block while field editor is open
+    await this.browser.keys(Key.Ctrl);
+    await this.browser.keys('x');
+    await this.browser.keys(Key.Ctrl); // release ctrl key
```
BenHenning: Could you just use `await this.browser.keys([Key.Ctrl, 'x']);` instead? I think we do that elsewhere.

BenHenning (approved):
Thanks @maribethb, LGTM! Just had one comment that might lead to a slight test simplification.

test/webdriverio/test/clipboard_test.ts
maribethb: Oh, I guess so. The [webdriver documentation](https://webdriver.io/docs/api/browser/keys/) includes a very confusing statement that modifier keys will be considered held down until you trigger them again, but then their example with copy and paste doesn't do anything related to that, so... I guess it's fine.

PR #575, opened 2025-05-31
chore: Added context menu item tests.

johnnesky:
Implements the unit tests described in https://docs.google.com/document/d/1pIIX5sKEG9f_NcWDIQJJEOMM7w7pQUf8eJoT3lrAElQ for context menu items in various situations.

The set of expected available context menu items depends on the focused object, whether it's in a flyout, and how the menu was opened. The tests are currently configured to expect the set of menu items that the currently implemented behavior produces.

### DIFF
--- a/test/webdriverio/test/actions_test.ts
+++ b/test/webdriverio/test/actions_test.ts
@@ -22,6 +22,68 @@ import {
   contextMenuItems,
 } from './test_setup.js';
 
+const isDarwin = process.platform === 'darwin';
+const blockActionsViaKeyboard = [
+  {'text': 'Duplicate D'},
+  {'text': 'Add Comment'},
+  {'text': 'External Inputs'},
+  {'text': 'Collapse Block'},
+  {'text': 'Disable Block'},
+  {'text': 'Delete 2 Blocks Delete'},
+  {'text': 'Move Block M'},
+  {'text': 'Edit Block contents Right'},
+  {'text': isDarwin ? 'Cut ⌘ X' : 'Cut Ctrl + X'},
+  {'text': isDarwin ? 'Copy ⌘ C' : 'Copy Ctrl + C'},
+  {'disabled': true, 'text': isDarwin ? 'Paste ⌘ V' : 'Paste Ctrl + V'},
+];
+
+const blockActionsViaMouse = [
+  {'text': 'Duplicate D'},
+  {'text': 'Add Comment'},
+  {'text': 'External Inputs'},
+  {'text': 'Collapse Block'},
+  {'text': 'Disable Block'},
+  {'text': 'Delete 2 Blocks Delete'},
+  {'text': isDarwin ? 'Cut ⌘ X' : 'Cut Ctrl + X'},
+  {'text': isDarwin ? 'Copy ⌘ C' : 'Copy Ctrl + C'},
+  {'disabled': true, 'text': isDarwin ? 'Paste ⌘ V' : 'Paste Ctrl + V'},
+];
+
+const shadowBlockActionsViaKeyboard = [
+  {'text': 'Add Comment'},
+  {'text': 'Collapse Block'},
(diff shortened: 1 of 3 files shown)


test/webdriverio/test/actions_test.ts
```diff
@@ -27,6 +31,18 @@ suite('Menus test', function () {
...
+    await tabNavigateToWorkspace(this.browser);
+    await this.browser.pause(PAUSE_TIME);
+    await this.browser.keys([Key.Ctrl, Key.Return]);
+    await this.browser.pause(PAUSE_TIME);
+
+    chai.assert.isTrue(
+      await contextMenuExists(this.browser, 'Undo', /* reverse= */ true),
+      'The menu should not be openable on the workspace',
+    );
+  });
```
cpcallen: This may be the current behaviour but only because we have not yet correctly implemented keyboard access to the workspace context menu (which requires first implementing the workspace cursor properly).  Does it make sense to add a test to assert broken behaviour?

test/webdriverio/test/actions_test.ts
```diff
@@ -70,4 +84,133 @@ suite('Menus test', function () {
...
+        'External Inputs',
+        'Collapse Block',
+        'Disable Block',
+        'Delete 2 Blocks',
+        'Move Block',
+        'Edit Block contents',
+        'Cut',
+        'Copy',
+        'Paste',
+      ],
```
cpcallen: This test appears to enforce an ordering.  If we are concerned about the order, should we not also be concerned about presence of spacers, shortcut keys, etc.?

test/webdriverio/test/actions_test.ts
johnnesky: I deleted this test.

test/webdriverio/test/actions_test.ts
johnnesky: I rebased my branch and now I'm using the same strategy to verify menu items as other tests, including shortcut keys.

test/webdriverio/test/actions_test.ts
```diff
@@ -64,14 +68,75 @@ suite('Menus test', function () {
...
+
+  test('Block menu via mouse displays expected items', async function () {
+    await tabNavigateToWorkspace(this.browser);
+    await clickBlock(this.browser, 'draw_circle_1', {button: 'right'});
+
+    chai.assert.deepEqual(
       await contextMenuItems(this.browser),
+      process.platform === 'darwin'
+        ? [
+            {'text': 'Duplicate D'},
```
RoboErikG: What do you think of pulling these lists of context menu items up into some constants to make it easier to compare kb and mouse items? I think it could be helpful to have both together to make it more obvious when changing one to also consider changing the other.

RoboErikG (approved):
Tests look good. One suggestion about making it easier to update the expected menu items.

test/webdriverio/test/actions_test.ts
johnnesky: I moved the lists of action labels into constants (and refactored the `platform === 'darwin'` checks to reduce redundancy).

However, when the automated github action runs, one of the test methods fails, which I can't reproduce on my local macbook. The test that fails is `'Shadow block menu via keyboard displays expected items'`, which opens a context menu when a shadow block is in focus. The error is that the `'Add Comment'` option is hidden, while the rest of the context menu options are as expected. According to the core blockly repo, this action has the precondition `scope.workspace?.isMutator ? 'hidden' : 'enabled'`. The test is using a color picker as the shadow block, and color pickers do have a mutator when you click on them, although I'm not sure why the mutator is being opened in this situation and why the test failure only happens on github. Maybe I'll try a different kind of shadow block.

johnnesky: Okay yeah I loaded different test blocks and used a text shadow block instead, that fixed it!

PR #588, opened 2025-06-10
feat: Allow passing HTML elements to display in toasts.

gonfunko:
This PR fixes #[467](https://github.com/google/blockly-keyboard-experimentation/issues/467) by registering a custom toast implementation that checks for an `element` key in the `ToastOptions` object. If present, the HTML element referenced by that key will be displayed in the body of the toast. This allows callers to construct abitrary HTML content and display it in a toast, while still providing support for normal plaintext-based toasts.

### DIFF
--- a/src/html_toast.ts
+++ b/src/html_toast.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Configuration options for toasts.
+ */
+interface HtmlToastOptions extends Blockly.ToastOptions {
+  element?: HTMLElement;
+}
+
+/**
+ * Custom toast implementation that supports HTML elements in toast messages.
+ */
+class HtmlToast extends Blockly.Toast {
+  /**
+   * Creates the body of the toast for display.
+   *
+   * @param workspace The workspace the toast will be displayed on.
+   * @param options Configuration options for toast appearance/behavior.
+   * @returns The body for the toast.
+   */
+  protected static override createDom(
+    workspace: Blockly.WorkspaceSvg,
+    options: Blockly.ToastOptions,
+  ) {
+    const dom = super.createDom(workspace, options);
+    const contents = dom.querySelector('div');
+    if (
+      contents &&
+      'element' in options &&
+      options.element instanceof HTMLElement
+    ) {
+      contents.innerHTML = '';
+      contents.appendChild(options.element);
+    }
+    return dom;
+  }
+}
+
+/**
+ * Registers HtmlToast as the default toast implementation for Blockly. */
+export function registerHtmlToast() {
+  Blockly.dialog.setToast(HtmlToast.show.bind(HtmlToast));
+}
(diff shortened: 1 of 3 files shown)


src/html_toast.ts
```diff
@@ -0,0 +1,49 @@
...
+      contents.innerHTML = '';
+      contents.appendChild(options.element);
+    }
+    return dom;
+  }
+}
+
+/**
+ * Registers HtmlToast as the default toast implementation for Blockly. */
+export function registerHtmlToast() {
```
BenHenning, line 52: Should this also be unregistered?

src/html_toast.ts
```diff
@@ -0,0 +1,49 @@
...
+    ) {
+      contents.innerHTML = '';
+      contents.appendChild(options.element);
+    }
+    return dom;
+  }
+}
+
+/**
+ * Registers HtmlToast as the default toast implementation for Blockly. */
```
BenHenning: Looks like there's a formatting issue here per the eslint error. Please fix.

src/actions/move.ts
```diff
@@ -36,7 +36,7 @@ export class MoveActions {
     const shortcuts: ShortcutRegistry.KeyboardShortcut[] = [
       // Begin and end move.
       {
-        name: Msg['START_MOVE'],
+        name: 'start_move',
```
BenHenning: Are these `Msg` changes intended and correct?

src/html_toast.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Configuration options for toasts.
+ */
+interface HtmlToastOptions extends Blockly.ToastOptions {
```
BenHenning, line 12: Did you intend to use this? It appears unused, so I'm wondering if it should be removed or if `HtmlToast` should be updated to use it.

src/html_toast.ts
```diff
@@ -0,0 +1,49 @@
...
+  element?: HTMLElement;
+}
+
+/**
+ * Custom toast implementation that supports HTML elements in toast messages.
```
BenHenning, line 17: This might be worth updating to include an example of how to use `HtmlToast` since it seems that the options object needs to be very particularly set up in order for this to work correctly.

BenHenning (changes requested):
Thanks @gonfunko! Left some thoughts.

The approach makes sense. It may well be worth adding tests for this, as well, if the goal is ultimately to move it into Core.

src/html_toast.ts
gonfunko: Done.

src/actions/move.ts
gonfunko: Yes, but in https://github.com/google/blockly-keyboard-experimentation/pull/587, not here. Rebased to remove.

src/html_toast.ts
gonfunko: Yes, it's used by `HtmlToast`, updated the types to reflect that.

src/html_toast.ts
gonfunko: Done.

src/html_toast.ts
gonfunko: We can't usefully do so because we can't know what the originally-registered toast function was. And in any case, this one falls back to the stock behavior if you don't pass `element` in the options object.

BenHenning (approved):

PR #594, opened 2025-06-13
feat: Render keyboard shortcuts nicely in menus.

### LINKED ISSUE #590: Pretty-print keyboard shortcuts in context menus
### Check for duplicates

### Problem

Keyboard shortcuts in context menus are currently appended to the name of the shortcut. This is not pretty. They should instead be right-aligned and light grey, to match standard visual language for displaying keyboard shortcuts.

### Request

https://github.com/google/blockly-keyboard-experimentation/issues/466#issuecomment-2902388242

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
This PR fixes #590 by rendering menu items that have keyboard shortcuts with their corresponding shortcut off to the side in grey:

Blockly contextual menu with keyboard shortcuts shown

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -161,6 +161,35 @@
         .blocklyToolboxCategoryContainer:focus-visible {
         outline: none;
       }
+
+      .blocklyRTL
+        .blocklyKeyboardNavigation
+        .blocklyMenuItemContent
+        .blocklyShortcutContainer {
+        flex-direction: row-reverse;
+      }
+      .blocklyKeyboardNavigation
+        .blocklyMenuItemContent
+        .blocklyShortcutContainer {
+        width: 100%;
+        display: flex;
+        justify-content: space-between;
+      }
+      .blocklyKeyboardNavigation
+        .blocklyMenuItemContent
+        .blocklyShortcutContainer
+        .blocklyShortcut {
+        color: #ccc;
+        margin-left: 16px;
+      }
+      .blocklyRTL
+        .blocklyKeyboardNavigation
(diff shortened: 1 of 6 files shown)


gonfunko:
Filed https://github.com/google/blockly/issues/9144 in core to track updating the messages.

test/index.html
```diff
@@ -161,6 +161,35 @@
         .blocklyToolboxCategoryContainer:focus-visible {
         outline: none;
       }
+
```
maribethb, line 164: Is it intentional to only put this CSS in the test page? Then users of the plugin will have to add this CSS to their applications themselves. I realized I don't know if we're following that approach with other style decisions in this plugin but it seems a bit strange if so since we want to provide good defaults. Does this still look good if you don't have this CSS?

maribethb (approved):

test/index.html
gonfunko: AFAICT this is where all the CSS lives at the moment, so I put it with the rest of it; presumably when we move this back to core it'll get merged in there, but perhaps we should file an issue to register all of the plugin's CSS now? And no, it looks pretty bad without the styling.

maribethb: Yeah, I'll file an issue for that. We want it to be overridable because i'm assuming makecode has extensive custom styling, but it should look ok by default. maybe have a parameter to control whether to register any css or not. but maybe that's not necessary as you can just re-style it yourself if you don't like it.

gonfunko: Sounds good, thanks! I'll merge as-is then.

PR #597, opened 2025-06-16
fix: Allow some things to handle the enter key on read only workspaces

### LINKED ISSUE #585: Blockly icons (e.g., comments) cannot be opened via keyboard in readonly mode
### Check for duplicates

### Description

[Blockly icons are inaccessible in readonly mode via keyboard](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/actions/enter.ts#L57-L58), but they are otherwise clickable. This can be observed in the [Blockly playground page with advanced settings on](https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html):

- Add a block
- Add a block comment
- Go to the right hand panel -> options -> basic -> toggle readOnly on
- Observe that the comment is clickable

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

RoboErikG:
Fixes #585 

This updates the enter handler to do more selective checks when deciding whether or not to handle an enter key. Also simplifies getState() slightly to remove an unnecessary parameter.

In general, this is aiming to reach parity between mouse and keyboard behavior when on a read-only workspace. Specific changes to behavior:

- Enables creating a workspace cursor ('w' key) while in read-only mode.
- Enables enter key to "click" on icons while in read-only mode.
- Enabled navigating to the toolbox while in read-only mode.
- Factors out most of the checks for whether or not an enter will be handled to the precondition.
- Returns false from the precondition and callback more often when the shortcut isn't actually handled.

TODO for https://github.com/google/blockly/issues/8915 
Add additional tests for enter and click with a read-only workspace and ensure both modalities are consistent.

### DIFF
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -207,7 +207,9 @@ export class Clipboard {
       !!this.oldCutShortcut?.callback &&
       this.oldCutShortcut.callback(workspace, e, shortcut, scope);
     if (didCut) {
-      this.copyWorkspace = workspace;
+      this.copyWorkspace = workspace.isFlyout
+        ? workspace.targetWorkspace
+        : workspace;
       showCutHint(workspace);
     }
     return didCut;
@@ -285,7 +287,9 @@ export class Clipboard {
       !!this.oldCopyShortcut?.callback &&
       this.oldCopyShortcut.callback(workspace, e, shortcut, scope);
     if (didCopy) {
-      this.copyWorkspace = workspace;
+      this.copyWorkspace = workspace.isFlyout
+        ? workspace.targetWorkspace
+        : workspace;
       showCopiedHint(workspace);
     }
     return didCopy;
--- a/src/actions/clipboard.ts
+++ b/src/actions/clipboard.ts
@@ -207,9 +207,7 @@ export class Clipboard {
       !!this.oldCutShortcut?.callback &&
       this.oldCutShortcut.callback(workspace, e, shortcut, scope);
     if (didCut) {
-      this.copyWorkspace = workspace.isFlyout
-        ? workspace.targetWorkspace
-        : workspace;
+      this.copyWorkspace = workspace;
       showCutHint(workspace);
     }
     return didCut;
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -115,6 +115,12 @@ export class EnterAction {
     const curNode = cursor?.getCurNode();
     if (!curNode) return false;
     if (curNode instanceof Field && !curNode.isClickable()) return false;
+    if (
+      curNode instanceof RenderedConnection ||
+      curNode instanceof WorkspaceSvg
+    ) {
+      return !workspace.isReadOnly();
+    }
     // Returning true is sometimes incorrect for icons, but there's no API to check.
     return true;
   }
@@ -131,19 +137,23 @@ export class EnterAction {
     if (!curNode) return false;
     if (curNode instanceof Field) {
       curNode.showEditor();
+      return true;
     } else if (curNode instanceof BlockSvg) {
       if (!this.tryShowFullBlockFieldEditor(curNode)) {
         showHelpHint(workspace);
       }
+      return true;
     } else if (
       curNode instanceof RenderedConnection ||
       curNode instanceof WorkspaceSvg
     ) {
       this.navigation.openToolboxOrFlyout(workspace);
+      return true;
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
+      return true;
     }
-    return true;
+    return false;
   }
 
   /**
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -114,7 +114,7 @@ export class EnterAction {
     const cursor = workspace.getCursor();
     const curNode = cursor?.getCurNode();
     if (!curNode) return false;
-    if (curNode instanceof Field && !curNode.isClickable()) return false;
+    if (curNode instanceof Field) return curNode.isClickable();
     if (
       curNode instanceof RenderedConnection ||
       curNode instanceof WorkspaceSvg
@@ -122,7 +122,8 @@ export class EnterAction {
       return !workspace.isReadOnly();
     }
     // Returning true is sometimes incorrect for icons, but there's no API to check.
-    return true;
+    if (curNode instanceof icons.Icon) return true;
+    return false;
   }
 
   /**
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -56,9 +56,7 @@ export class EnterAction {
       preconditionFn: (workspace): boolean => {
         switch (this.navigation.getState()) {
           case Constants.STATE.WORKSPACE:
-            // The main workspace may or may not handle it depending on what's
-            // selected, so always pass it through to the callback.
-            return true;
+            return this.shouldHandleEnterForWS(workspace);
           case Constants.STATE.FLYOUT: {
             // If we're in the flyout the only supported actions are inserting
             // blocks or clicking buttons, so don't handle this if the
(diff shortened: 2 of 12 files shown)


src/actions/clipboard.ts
```diff
@@ -207,7 +207,9 @@ export class Clipboard {
       !!this.oldCutShortcut?.callback &&
       this.oldCutShortcut.callback(workspace, e, shortcut, scope);
     if (didCut) {
-      this.copyWorkspace = workspace;
+      this.copyWorkspace = workspace.isFlyout
```
maribethb: just checking, is this change supposed to be here? it doesn't seem related to the PR title

src/actions/enter.ts
```diff
@@ -54,33 +53,51 @@ export class EnterAction {
...
       name: Constants.SHORTCUT_NAMES.EDIT_OR_CONFIRM,
-      preconditionFn: (workspace) =>
-        this.navigation.canCurrentlyEdit(workspace),
-      callback: (workspace, event) => {
+      preconditionFn: (workspace): boolean => {
+        switch (this.navigation.getState()) {
+          case Constants.STATE.WORKSPACE:
+            // The main workspace may or may not handle it depending on what's
+            // selected, so always pass it through to the callback.
+            return true;
```
maribethb: The precondition should really try to give an accurate answer. The keyboard shortcut help display could be depending on the precondition being accurate. Rachel envisions a contextual help where only keyboard shortcuts that would currently do anything are shown.

src/actions/enter.ts
```diff
@@ -54,33 +53,51 @@ export class EnterAction {
...
+          case Constants.STATE.FLYOUT: {
+            // If we're in the flyout the only supported actions are inserting
+            // blocks or clicking buttons, so don't handle this if the
```
maribethb, line 62: Are we sure clicking buttons shouldn't be allowed? I feel like on one hand, this is too simple, but on the other hand, having a flyout at all in a read-only workspace is unlikely as that's just confusing, so maybe this is fine.

src/actions/enter.ts
```diff
@@ -54,33 +53,51 @@ export class EnterAction {
...
           case Constants.STATE.WORKSPACE:
-            this.handleEnterForWS(workspace);
-            return true;
+            return this.handleEnterForWS(workspace);
           case Constants.STATE.FLYOUT:
-            if (!workspace.targetWorkspace) return false;
-            flyoutCursor = this.navigation.getFlyoutCursor(
-              workspace.targetWorkspace,
-            );
+            if (targetWorkspace.isReadOnly()) return false;
```
maribethb: Do you need to check this again since you're already checking it in the precondition?

src/actions/enter.ts
```diff
@@ -114,6 +132,7 @@ export class EnterAction {
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
     }
+    return true;
```
maribethb: Despite the comment in the precondition, it seems like now we're saying the keyboard shortcut for enter will always be handled regardless of whether the workspace is in read-only mode or not. This doesn't seem correct to me, e.g. the field editors shouldn't open if the workspace is in read-only mode. 

Based on #585 and playing with the current mouse behavior for read-only mode, I think only icons should be special cased, and the rest of these should continue not doing anything.

It would be helpful if you can add in the PR descriptions which cases you intended to make interactable in read-only, as then I know if we're on the same page about the intentions which may be different than being on the same page about the actual result of the code (which I may be misreading, for example).

src/actions/enter.ts
```diff
@@ -278,7 +297,7 @@ export class EnterAction {
     if (block.isSimpleReporter()) {
       for (const input of block.inputList) {
         for (const field of input.fieldRow) {
-          if (field.isClickable() && field.isFullBlockField()) {
+          if (field.isFullBlockField()) {
```
maribethb: why remove this check? if the field isn't clickable, it shouldn't be "enter"able either, I think.

src/actions/clipboard.ts
RoboErikG: Whoops, not sure how this slipped in.

src/actions/enter.ts
RoboErikG: For some cases there's no API to ask if it would be allowed and some of them we don't even know if it did anything. For example, the Icon.onClick() method doesn't have a return and there's no isClickable() method, but it only opens the mutator if the workspace is editable.

Because of that, it didn't seem like it was worth making the precondition overly complex if we're going to have to handle it anyway, but I'll factor out the logic to check all the known cases.

src/actions/enter.ts
RoboErikG: Yeah, there may be corner cases where this is an over-simplification, but I couldn't come up with any good use cases to check for so this seemed appropriate.

src/actions/enter.ts
RoboErikG: Removed. Thanks for catching!

src/actions/enter.ts
RoboErikG: The problem is we don't know if it was handled or not for all cases, specifically icons which have an on click() method that doesn't return anything. I've added additional return false statements where we do know.

src/actions/enter.ts
RoboErikG: I'd removed it because field.showEditor() calls this.isClickable() before showing the editor so it had seemed redundant. That said, I'm adding it back so that this method returns whether or not it handled the return correctly.

src/actions/enter.ts
```diff
@@ -90,16 +104,31 @@ export class EnterAction {
...
+   * @param workspace The workspace to check.
+   * @returns True if the enter action should be handled.
+   */
+  private shouldHandleEnterForWS(workspace: WorkspaceSvg): boolean {
+    const cursor = workspace.getCursor();
+    const curNode = cursor?.getCurNode();
+    if (!curNode) return false;
+    if (curNode instanceof Field && !curNode.isClickable()) return false;
+    // Returning true is sometimes incorrect for icons, but there's no API to check.
+    return true;
```
maribethb: This is still going to be a behavior change for pressing enter on other cases, if I'm reading this correctly. If the workspace is read-only then pressing enter shouldn't open the toolbox. It doesn't before this change but I think this change will cause it to open because of this precondition + you're not checking it in the callback.

src/actions/enter.ts
maribethb: OK, that's fine for icons.

For connections and workspace, I don't think the enter key should do anything in read-only mode, but I think this does actually cause them to do something. I can't tell if you intended them to change behavior and we should discuss that intention, or if that is unintentional, or if I'm wrong and the behavior hasn't changed.

src/actions/enter.ts
```diff
@@ -114,6 +143,7 @@ export class EnterAction {
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
     }
+    return true;
```
maribethb: Also, you shouldn't return true as the default, because if the current node isn't an instance of any of these things (e.g. custom focusable object, or workspace comment), you shouldn't say we've handled the shortcut because we haven't. Those custom objects could register their own shortcut handler for enter and if you return true that could cause theirs to not be executed. 

I realize this is the old behavior, but I think that's a bug.

src/actions/enter.ts
RoboErikG: Yes, that was intentional to make it consistent with mouse which still opens the toolbox on click.

src/actions/enter.ts
RoboErikG: Ah, I'd misunderstood that path. Added a check to shouldHandle for this case.

src/actions/enter.ts
RoboErikG: Updated to return false by default and true for each case we handle.

src/actions/enter.ts
RoboErikG: err, ignore my previous comment. It's intentional that pressing 't' opens the toolbox on a read only workspace. I've updated it so pressing enter on the ws or a connection won't since that's an insert action.

maribethb (approved):
lgtm, but it would be helpful if you can make the pr description more informative for future reference when reading the changelog, e.g.:

- Enables creating a workspace cursor ('w' key) while in readonly mode
- Enables enter key to "click" on icons while in readonly mode
- Returns false from the keyboard shortcut callback more often when the shortcut isn't actually handled

PR #598, opened 2025-06-16
fix: Handle shortcuts being triggered on non-main workspace.

gonfunko:
This PR updates the shortcuts/actions to handle being triggered on a non-main workspace (e.g. the flyout or a mutator bubble) after https://github.com/google/blockly/pull/9137 updated core to fire keyboard events against the workspace they actually occurred in.

### DIFF
--- a/src/actions/arrow_navigation.ts
+++ b/src/actions/arrow_navigation.ts
@@ -95,7 +95,7 @@ export class ArrowNavigation {
     ): boolean => {
       const toolbox = workspace.isFlyout
         ? workspace.targetWorkspace?.getToolbox()
-        : (workspace.getToolbox() as Toolbox);
+        : workspace.getToolbox();
       let isHandled = false;
       switch (this.navigation.getState(workspace)) {
         case Constants.STATE.WORKSPACE:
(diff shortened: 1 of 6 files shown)


src/actions/arrow_navigation.ts
```diff
@@ -91,7 +93,9 @@ export class ArrowNavigation {
       e: Event,
       shortcut: ShortcutRegistry.KeyboardShortcut,
     ): boolean => {
-      const toolbox = workspace.getToolbox() as Toolbox;
+      const toolbox = workspace.isFlyout
+        ? workspace.targetWorkspace?.getToolbox()
+        : (workspace.getToolbox() as Toolbox);
```
maribethb: nit: do you need this cast? if you don't need it on the line above I don't know why you'd need it here. seems like it should be both or neither.

maribethb (approved):

src/actions/arrow_navigation.ts
gonfunko: Good catch, indeed it's not necessary. Removed!

maribethb: There's a couple other places in the same file you have the same cast which is probably unnecessary there too then

PR #599, opened 2025-06-17
fix(mover): Pass drag delta to `dragger.onDrag` in pixels

### LINKED ISSUE #446: Constrained move after unconstrained move translates the block in MakeCode
This issue only occurs in MakeCode, and not the blockly-keyboard-experimentation test page.

Reproduction:
- Insert a top-level block without any constrained positions
- Move the block using unconstrained move (modifier key and arrows)
- Attempt to the move the block using constrained move (arrows only)
- The block unexpectedly moves in the same direction you previously moved the block in unconstrained move mode
- The amount the block moves increases on each subsequent arrow key press

Please see screen capture below:
Image

What is happening:
The `newLoc` passed to the overridden `drag` method is being unexpectedly updated during constrained moves after an unconstrained move has occurred. This then calls `drag.super` and in turn `this.block.moveDuringDrag` which actually translates the block. It is unclear why there is a difference in behaviour between MakeCode and the test page.

A potential fix:
This can be fixed by removing the call to `info.updateTotalDelta()` inside the `moveConstrained` method, however, the consequences of doing this are not fully understood by us and we want to avoid any negative side effects.

cpcallen:
There was an error in the initial implementation of `DragMover` (in PR #317), where the drag delta was being passed in workspace coordinates instead of pixels.  This lead to various problems, notably #446.

Fixes #446.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -302,7 +302,7 @@ export class Mover {
 
     info.dragger.onDrag(
       info.fakePointerEvent('pointermove', direction),
-      info.totalDelta,
+      info.totalDelta.clone().scale(workspace.scale),
     );
 
     info.updateTotalDelta();
@@ -327,7 +327,10 @@ export class Mover {
     info.totalDelta.x += x * UNCONSTRAINED_MOVE_DISTANCE * workspace.scale;
     info.totalDelta.y += y * UNCONSTRAINED_MOVE_DISTANCE * workspace.scale;
 
-    info.dragger.onDrag(info.fakePointerEvent('pointermove'), info.totalDelta);
+    info.dragger.onDrag(
+      info.fakePointerEvent('pointermove'),
+      info.totalDelta.clone().scale(workspace.scale),
+    );
     this.scrollCurrentBlockIntoView(workspace);
     return true;
   }
--- a/test/webdriverio/test/move_test.ts
+++ b/test/webdriverio/test/move_test.ts
@@ -13,6 +13,9 @@ import {
   tabNavigateToWorkspace,
   testFileLocations,
   testSetup,
+  sendKeyAndWait,
+  keyDown,
+  keyRight,
 } from './test_setup.js';
 
 suite('Move tests', function () {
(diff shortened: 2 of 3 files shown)


src/actions/mover.ts
```diff
@@ -302,7 +302,7 @@ export class Mover {
 
     info.dragger.onDrag(
       info.fakePointerEvent('pointermove', direction),
-      info.totalDelta,
+      info.totalDelta.clone().scale(workspace.scale),
```
BenHenning, line 305: This makes me wonder if we should have distinct types for different coordinate spaces, e.g. ScreenCoordinate and WorkspaceCoordinate. It's not obvious from code that scaling performs the space conversion, or what coordinate system we're currently in at this point (since if `totalDelta` is already in screen space then scaling again would be wrong).

test/webdriverio/test/move_test.ts
```diff
@@ -13,6 +13,9 @@ import {
   tabNavigateToWorkspace,
   testFileLocations,
   testSetup,
+  sendKeyAndWait,
+  keyDown,
+  keyRight,
```
BenHenning: Looks like this can be removed.

test/webdriverio/test/move_test.ts
```diff
@@ -218,3 +277,22 @@ function getConnectedBlockInfo(browser: Browser, id: string, index: number) {
     index,
   );
 }
+
+/**
+ * Given a block ID, get the coordinates of that block, as returned by
+ * .getRelativeTosSurfaceXY().
```
BenHenning:
```suggestion
 * getRelativeToSurfaceXY().
```

Nit: typo fix, plus removed the '.' since it doesn't seem necessary for the code reference.

test/webdriverio/test/move_test.ts
```diff
@@ -145,6 +148,62 @@ suite('Move tests', function () {
       await this.browser.keys(Key.Escape);
     }
   });
+
+  // When a top-level block with no previous, next or output
+  // connections is subject to a constrained move, it should not move.
+  //
+  // This includes a regression test for issue #446 (fixed in PR #599)
+  // where, due to an implementation error in Mover, constrained
+  // movement following unconstrained movement it would result in the
```
BenHenning:
```suggestion
  // movement following unconstrained movement would result in the
```

Extra 'it' if I'm reading this correctly.

test/webdriverio/test/move_test.ts
```diff
@@ -145,6 +148,62 @@ suite('Move tests', function () {
...
+  // block unexpectedly moving (unless workspace scale was === 1)
```
BenHenning:
```suggestion
  // block unexpectedly moving (unless workspace scale was === 1).
```

Super nit on punctuation for consistency.

BenHenning (approved):
Thanks @cpcallen! Seems like a really nice fix. Just had some nits and one core design consideration, but otherwise the PR LGTM.

src/actions/mover.ts
cpcallen: This would be a great idea; I've filed google/blockly#9156 for future consideration.

Relatedly, one of the things I somewhat dislike about TypeScript (and also Go) is duck typing: it should be possible to give different names to the same type specification and have the compiler flag any mixed use (absent explicit casts).  For object types it's pretty easy to make things different shapes (adding a dummy method to the class definition changes the type without changing the actuals shape of the objects in memory), but a classic use of nominal typing is to distinguish between `string`s containing HTML and ones containing plain text, and enforce calling a function to do escaping when converting the latter to the former.  TypeScript doesn't let you do that—you'd have to use subclasses of the `String` object type rather than plain `string`s.

test/webdriverio/test/move_test.ts
cpcallen: Good spot.  This comment got reworded several times to try to keep it to a reasonable length; looks like I can shorten it by one more word!

PR #600, opened 2025-06-17
feat: Add support for navigating in mutator workspaces.

### LINKED ISSUE #9002: Make mutator bubbles focusable and navigable
See parent issue.

gonfunko:
In conjunction with https://github.com/google/blockly/pull/9151, this PR fixes https://github.com/google/blockly/issues/9002 by adding support for navigating into and manipulating blocks within mutator workspaces using keyboard navigation.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -154,7 +154,7 @@ export class EnterAction {
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
       renderManagement.finishQueuedRenders().then(() => {
-        cursor.in();
+        cursor?.in();
       });
       return true;
     }
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -152,6 +152,9 @@ export class EnterAction {
       this.navigation.openToolboxOrFlyout(workspace);
       return true;
     } else if (curNode instanceof icons.Icon) {
+      // Calling the icon's click handler will trigger its action, generally
+      // opening a bubble of some sort. We then need to wait for the bubble to
+      // appear before attempting to navigate into it.
       curNode.onClick();
       renderManagement.finishQueuedRenders().then(() => {
         cursor?.in();
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -17,6 +17,7 @@ import {
   Field,
   icons,
   FocusableTreeTraverser,
+  renderManagement,
 } from 'blockly/core';
 
 import type {Block} from 'blockly/core';
@@ -113,6 +114,9 @@ export class EnterAction {
       this.navigation.openToolboxOrFlyout(workspace);
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
+      renderManagement.finishQueuedRenders().then(() => {
+        cursor.in();
+      });
     }
   }
 
--- a/src/actions/exit.ts
+++ b/src/actions/exit.ts
@@ -58,6 +58,7 @@ export class ExitAction {
                 return true;
               }
             }
+            return false;
           }
           default:
             return false;
--- a/src/actions/exit.ts
+++ b/src/actions/exit.ts
@@ -9,6 +9,7 @@ import {
   utils as BlocklyUtils,
   getFocusManager,
   Gesture,
+  icons,
 } from 'blockly/core';
 
 import * as Constants from '../constants';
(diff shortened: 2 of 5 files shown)


test/webdriverio/test/mutator_test.ts
```diff
@@ -0,0 +1,107 @@
...
+    await this.browser.pause(PAUSE_TIME);
+    // Hit enter to enter insert mode
+    await this.browser.keys(Key.Enter);
+    await this.browser.pause(PAUSE_TIME);
+    // Hit enter again to lock it into place on the connection
+    await this.browser.keys(Key.Enter);
+
+    const topBlocks = await this.browser.execute(() => {
+      const focusedTree = Blockly.getFocusManager().getFocusedTree();
+      if (!(focusedTree instanceof Blockly.WorkspaceSvg)) return [];
```
BenHenning: Can we make this throw a hard error instead of being a condition? It seems that we probably don't want to default behavior in this case and instead just fail the test.

src/navigation.ts
```diff
@@ -824,9 +824,12 @@ export class Navigation {
    * @returns whether keyboard navigation is currently allowed.
    */
   canCurrentlyNavigate(workspace: Blockly.WorkspaceSvg) {
-    const accessibilityMode = workspace.isFlyout
-      ? workspace.targetWorkspace?.keyboardAccessibilityMode
-      : workspace.keyboardAccessibilityMode;
+    const accessibilityMode = (
+      workspace.getRootWorkspace() ??
+      workspace.targetWorkspace?.getRootWorkspace() ??
```
BenHenning, line 836: These two are new. I'm assuming `targetWorkspace` is only used for flyouts which is why that fallback below is a correct logical replacement to the previous `isFlyout` check, but are these root workspace checks always correct and in this order?

It might help to add a line comment before this explaining when each condition can occur, and why the fallback is correct if more than 2 can occur at once (since `workspace` will always be present).

src/actions/exit.ts
```diff
@@ -39,6 +40,26 @@ export class ExitAction {
               workspace.hideChaff();
             }
             return true;
+          case Constants.STATE.WORKSPACE: {
```
BenHenning, line 43: How does escape work when the mutator's flyout is open? At present it seems like it will navigate away from the mutator workspace rather than closing the flyout, or does this work because the navigation state correctly represents flyout/toolbox as needed?

This seems like a good test to add, as well, since it's a specific edge case that could be missed by existing tests if it broke in the future.

src/actions/enter.ts
```diff
@@ -152,6 +153,9 @@ export class EnterAction {
       return true;
     } else if (curNode instanceof icons.Icon) {
       curNode.onClick();
+      renderManagement.finishQueuedRenders().then(() => {
```
BenHenning, line 159: I'm slightly surprised this works in all cases, and am wondering if there are cases where it doesn't per this line:

https://github.com/google/blockly/blob/4977b4bc9bd70412456c2d125f5044cd34b625e0/core/icons/mutator_icon.ts#L171

It seems that updating the visibility of the bubble itself blocks on `finishQueuedRenders`, but then queues more? I'm actually unsure of the semantics when calling an async function from a non-async function without using `await`. https://stackoverflow.com/a/53730574 suggests that `finishQueuedRenders` could actually resolve after the rest of the logic occurs, but there's subtlety there with the return value that may not be relevant here.

I'm not sure if there's a better solution, but I think it might be helpful at least to add a comment here explaining why the `finishQueuedRenders` call is both necessary and correct following `onClick` in case we do catch a case in the future where this doesn't work and need to investigate it.

BenHenning (approved):
Thanks @gonfunko! Had a few thoughts, but the PR largely LGTM. Feel free to re-add me if you'd like another review pass.

src/actions/exit.ts
gonfunko: Mutator flyouts aren't closable, but if the mutator flyout is focused the navigation state is indeed `.FLYOUT`, so Escape moves focus to the mutator workspace. A second Escape will then dismiss the mutator bubble.

test/webdriverio/test/mutator_test.ts
gonfunko: Done.

src/navigation.ts
gonfunko: That's correct, added a comment explaining the logic here.

src/actions/exit.ts
gonfunko: Added the test as well.

src/actions/enter.ts
gonfunko: Experimentally it seems to, but I agree this does seem a bit fragile. Added the comment, and if we do run across a situation where this doesn't work we can investigate further at that point.

PR #604, opened 2025-06-18
fix: scroll into view used the wrong position

### LINKED ISSUE #577: Scroll block away from view when inserting a block
### Check for duplicates

### Description

When the project is long (blocks are far apart vertically), inserting a new block to the bottom of the block stack seems to scroll the newly-added block away from view. The expected behaviour is for the newly-added block to be in view after insertion.

### Reproduction steps

1. Drag the "draw" block to away from the "setup" block so that it is at the bottom of the screen.
2. Press "T" to open toolbox.
3. Navigate to a block in the flyout.
4. Hit "ENTER" on a block to insert.
5. See the newly added block be in move mode and out of view.

Moving the newly-added block around, or confirming move then navigating around with arrow keys moves focused block correctly back into view

### Stack trace

```shell

```

### Screenshots

**Video of reproduction**
https://github.com/user-attachments/assets/3ffe2822-2817-49c9-9214-ce24aad1babe

### Browsers

_No response_

microbit-matt-hillsdon:
It used the position from positionNewTopLevelBlock, not the initial insert
position. Delay the setting of current node until the position is correct.

Fixes #577

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -146,8 +146,6 @@ export class EnterAction {
 
     workspace.setResizesEnabled(true);
 
-    getFocusManager().focusTree(workspace);
-    workspace.getCursor()?.setCurNode(newBlock);
     this.mover.startMove(workspace, newBlock, insertStartPoint);
 
     const isStartBlock =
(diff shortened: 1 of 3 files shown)


src/actions/enter.ts
```diff
@@ -146,8 +146,6 @@ export class EnterAction {
 
     workspace.setResizesEnabled(true);
 
-    getFocusManager().focusTree(workspace);
```
microbit-matt-hillsdon, line 149: This line was pointless since `setCurNode` started moving focus.

gonfunko (approved):

PR #610, opened 2025-06-18
fix: Fix display of workspace focus ring in move mode.

### LINKED ISSUE #569: move mode focus issues
### Check for duplicates

### Description

Move mode has some unexpected issues around focus or focus visualization. See repro steps.

Possibly related but possibly not, https://github.com/google/blockly/issues/9112. That one is the more serious of the issues so it should be looked at first

## Workspace visualization
2. Focus a block
3. Press M for move mode
4. Notice that the workspace no longer has the blue focus ring.
5. Enter to commit move.
6. Notice the workspace once again has the blue focus ring.

I don't think it should disappear while move mode is active. This could just be a css issue.

## When field editors are open
This one relies on https://github.com/google/blockly/pull/9110 and #568 being in
1. Focus a field with a dropdown editor
2. Press enter to open the editor
3. Press M for move mode
4. Move mode isn't activated, because the precondition fails now, this is the correct behavior
5. However, the passive focus indicator on the field that has its editor open has disappeared.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
This PR fixes #569 (the first half) by updating the stylesheet as suggested by @microbit-matt-hillsdon . The latter half of the bug appears to have been incidentally fixed at some earlier point.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -146,6 +146,9 @@
         .blocklyWorkspace:has(.blocklyActiveFocus)
         .blocklyWorkspaceFocusRing,
       .blocklyKeyboardNavigation
+        .blocklySvg:has(~ .blocklyBlockDragSurface .blocklyActiveFocus)
+        .blocklyWorkspaceFocusRing,
+      .blocklyKeyboardNavigation
         .blocklyWorkspace.blocklyActiveFocus
         .blocklyWorkspaceFocusRing {
         stroke: var(--blockly-active-tree-color);


test/index.html
```diff
@@ -145,6 +145,9 @@
       .blocklyKeyboardNavigation
         .blocklyWorkspace:has(.blocklyActiveFocus)
         .blocklyWorkspaceFocusRing,
+      .blocklyKeyboardNavigation
+        .blocklySvg:has(~ .blocklyBlockDragSurface .blocklyActiveFocus)
```
maribethb, line 149: seems fine if it works, but what does the `~` do when it's at the front like that? all the examples I found have it in between two other selectors, so I don't know what this is actually doing

maribethb (approved):

test/index.html
gonfunko: I just copied it from @microbit-matt-hillsdon's comment, but it appears to let you target the first part of the ~ relationship, i.e. an element with the blocklySvg class that has a sibling drag surface that has active focus. https://developer.mozilla.org/en-US/docs/Web/CSS/:has#with_the_sibling_combinator discusses this for +, but the idea is the same, just without the strict-next-sibling qualification.

maribethb: Thanks for the link!

PR #611, opened 2025-06-19
chore: Add `type: "module"` to `package.json`

cpcallen:
This means that `.js` files will by default be interpreted as ESM rather than CJS—which is correct, since almost all our source files are in fact ES modules using `import`.

This allows removal of `test/webdriverio/test/package.json`, which was preventing npm scripts (e.g. `npm test`) from working properly in that directory.

A few config files that do in fact use `require()` are renamed to `.cjs` to ensure they are correctly interpreted.


maribethb (approved):
LGTM, you mentioned you wanted to check the packaged output to make sure that was correct still, were you able to do that?

maribethb:
Closing this since it's out of date and we're moving active development to core.

PR #612, opened 2025-06-19
Ensure black on white for move indicator.

microbit-matt-hillsdon:
This fixes the icon in MakeCode's nascent dark theme, which sets color to white at a high level in the DOM which is then used on semi-transparent white here via currentColor. Seems reasonable to hardcode the foreground if we do so for the white background.

Example of broken state in MakeCode:

image

If there are further styling demands here then a CSS class on the group might be helpful. Having said that, I think this icon looks OK in dark and light modes as-is.

### DIFF
--- a/src/move_indicator.ts
+++ b/src/move_indicator.ts
@@ -53,7 +53,7 @@ export class MoveIndicatorBubble
       Blockly.utils.Svg.PATH,
       {
         'fill': 'none',
-        'stroke': 'currentColor',
+        'stroke': 'black',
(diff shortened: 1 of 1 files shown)


cpcallen:
@gonfunko: Can you take a look at this?  I'm not quite sure why you used `currentColor` here, so I'm not sure if it's safe to change it.

(Also, tsk for misspelling "colour" ;-)

gonfunko (approved):

PR #613, opened 2025-06-20
test: allow testing of workspace comments

microbit-matt-hillsdon:
Call `registerCommentOptions` so we can add workspace comments which are worth testing with keyboard navigation.
Add the delete icon as that's a factor in navigation too.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -99,6 +99,10 @@
         box-sizing: border-box;
       }
 
+      .blocklyDeleteIcon {
+        display: block;
+      }
+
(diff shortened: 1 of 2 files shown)


RoboErikG (approved):

microbit-matt-hillsdon:
Thanks for the review. Just a reminder that someone else will need to merge 🙏

RoboErikG:
Whoops, I kicked off the tests and forgot to merge when they finished. Thanks for the reminder!

PR #617, opened 2025-06-24
chore: upgrade field colour to buggy version

microbit-matt-hillsdon:
This illustrates that the latest version, based on field-grid-dropdown, does not work with keyboard navigation enabled. This is because Enter propagates to the global shortcut handler, causing the field editor to be retriggered or block toast messages to be shown.

On reflection I think this is actually a plugin bug surfaced by recent changes to field-colour. I'll open a PR shortly.

Demo link: https://upgrade-field-colour.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "devDependencies": {
         "@blockly/dev-scripts": "^4.0.8",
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
I'm going to close this in favour of a PR that has a sequence of commits including this one and a fix. Closing.

PR #619, opened 2025-06-24
fix: account for ephemeral focus in enter precondition 

### LINKED ISSUE #618: Enter in dropdown div can trigger keyboard nav enter action
Note: Essentially the same as my closed https://github.com/google/blockly-samples/issues/2568 but I decided it belonged here.

### Check for duplicates

### Description

field-grid-dropdown and field-colour (since #2559), propagate Enter/Space breaking keyboard navigation because the global handler does unhelpful things like try to retrigger the field editor or show a toast.

But I think this is actually showing up a bug in the plugin which should not be handling enter on a button in a widget/dropdown div in its enter shortcut handler.

### Reproduction steps

[Demo here](https://upgrade-field-colour.blockly-keyboard-experimentation.pages.dev/) which upgrades the keyboard nav experiment to 6.0.2 which has the rewrite of the colour field in terms of the grid dropdown. 

#### Scenario 1

1. Trigger the colour field editor with the mouse
2. Choose a colour with the keyboard and press Enter
3. Get an unexpected toast

#### Scenario 2

1. Trigger the colour field editor with the keyboard
2. Choose a colour with the keyboard and press Enter
3. Get a weird version of the colour picker back again. For reasons I can't fathom, this appears to be a version that loses the FieldColour subclass changes:

Image

The difference between the two scenarios is just which block the cursor is on.

### Stack trace

You get a trace in scenario 2 but more to the point you're trying to open an editor at the wrong time.

```shell
Uncaught Error: Attempted to take ephemeral focus when it's already held, with new element: [object HTMLDivElement].
    at jl.takeEphemeralFocus (bundle.js:2:1182134)
    at so (bundle.js:2:1101714)
    at no (bundle.js:2:1101433)
    at Object.to [as showPositionedByField] (bundle.js:2:1100978)
    at a.showEditor_ (bundle.js:2:6929)
    at a.showEditor_ (bundle.js:2:11301)
    at a.showEditor (bundle.js:2:1240079)
    at Hr.tryShowFullBlockFieldEditor (bundle.js:2:1937267)
    at Hr.handleEnterForWS (bundle.js:2:1935163)
    at Object.callback (bundle.js:2:1934569)
```

### Screenshots

_No response_

microbit-matt-hillsdon:
Previously bugs could occur if the field editor contained e.g. a button but did not prevent keydown events propagating.

This is true of field-colour from 6.0.2 so this PR upgrades to demonstrate that.

I've tried to make the enter action's preconditions more similar to the others, using canCurrentlyNavigate/canCurrentlyEdit.

I've also changed where we account for ephemeral focus so that the Constants.STATE value is NOWHERE when ephemeral focus is in progress. I don't think anything good can come of the state value representing somewhere that doesn't have focus.

Demos:
- [First commit that just upgrades field colour to demo the bug](https://upgrade-field-colour.blockly-keyboard-experimentation.pages.dev/)
- [This branch with fix](https://ephemeral-enter.blockly-keyboard-experimentation.pages.dev/)

Closes https://github.com/google/blockly-keyboard-experimentation/issues/618

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -96,7 +96,12 @@ export class Navigation {
    * @returns The state of the given workspace.
    */
   getState(): Constants.STATE {
-    const focusedTree = Blockly.getFocusManager().getFocusedTree();
+    const focusManager = Blockly.getFocusManager();
+    if (focusManager.ephemeralFocusTaken()) {
+      return Constants.STATE.NOWHERE;
+    }
+
+    const focusedTree = focusManager.getFocusedTree();
     if (focusedTree instanceof Blockly.WorkspaceSvg) {
       if (focusedTree.isFlyout) {
         return Constants.STATE.FLYOUT;
@@ -837,11 +842,7 @@ export class Navigation {
       workspace.targetWorkspace ??
       workspace
     ).keyboardAccessibilityMode;
-    return (
-      !!accessibilityMode &&
-      this.getState() !== Constants.STATE.NOWHERE &&
-      !Blockly.getFocusManager().ephemeralFocusTaken()
-    );
+    return !!accessibilityMode && this.getState() !== Constants.STATE.NOWHERE;
   }
 
   /**
@@ -854,7 +855,7 @@ export class Navigation {
    * @returns whether keyboard navigation and editing is currently allowed.
    */
   canCurrentlyEdit(workspace: Blockly.WorkspaceSvg) {
-    return this.canCurrentlyNavigate(workspace) && !workspace.options.readOnly;
+    return this.canCurrentlyNavigate(workspace) && !workspace.isReadOnly();
   }
 
   /**
--- a/test/webdriverio/test/basic_test.ts
+++ b/test/webdriverio/test/basic_test.ts
@@ -375,15 +375,14 @@ suite('Keyboard navigation on Fields', function () {
     await keyRight(this.browser);
     // Enter to choose.
     await this.browser.keys(Key.Enter);
-    await this.browser.pause(1000);
 
-    // Check that browser focus is back on the colour block, not e.g. a widget
-    // or dropdown div. Not sufficient to check with the focus manager.
-    // TODO: fix and flip.
-    chai.assert.isFalse(
-      await this.browser.execute(() =>
-        document.activeElement?.classList.contains('blocklyActiveFocus'),
-      ),
+    // Focus seems to take longer than a single pause to settle.
+    await this.browser.waitUntil(
+      () =>
+        this.browser.execute(() =>
+          document.activeElement?.classList.contains('blocklyActiveFocus'),
+        ),
+      {timeout: 1000},
     );
   });
 });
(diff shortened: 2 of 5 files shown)


src/navigation.ts
```diff
@@ -854,7 +855,7 @@ export class Navigation {
    * @returns whether keyboard navigation and editing is currently allowed.
    */
   canCurrentlyEdit(workspace: Blockly.WorkspaceSvg) {
-    return this.canCurrentlyNavigate(workspace) && !workspace.options.readOnly;
+    return this.canCurrentlyNavigate(workspace) && !workspace.isReadOnly();
```
microbit-matt-hillsdon, line 858: Changed for clarity as I had to read the code to learn that `options.readOnly` was mutated when `setIsReadOnly` was called.

microbit-matt-hillsdon:
I think this regressed in #597 when the relevant preconditions started just checking the state not calling the common canXXX methods. I think calling the navigate one as I have done maintains the needed behaviour, but it would be safe written either way with the changes to `getState` in this PR. I've checked you can still open comments if you mark the workspace read-only after creating one.

test/webdriverio/test/basic_test.ts
```diff
@@ -363,4 +363,26 @@ suite('Keyboard navigation on Fields', function () {
...
+    // Move right to pick a new colour.
+    await keyRight(this.browser);
+    // Enter to choose.
+    await this.browser.keys(Key.Enter);
+
+    // Focus seems to take longer than a single pause to settle.
+    await this.browser.waitUntil(
+      () =>
+        this.browser.execute(() =>
+          document.activeElement?.classList.contains('blocklyActiveFocus'),
```
BenHenning, line 383:
```suggestion
          document.activeElement?.classList?.contains('blocklyActiveFocus'),
```

Is this needed to avoid a potential null?

BenHenning (approved):
Thanks @microbit-matt-hillsdon. This change makes sense, and the new test will hopefully keep this working if this code has to change again.

No concerns on my end, just one comment that may lead to a slight test improvement.

test/webdriverio/test/basic_test.ts
microbit-matt-hillsdon: I don't think so, e.g. testing in Node:

```node
> document = { activeElement: null }
{ activeElement: null }
> document.activeElement?.classList.contains('foo')
undefined
```

maribethb: To add to matt's answer, if the activeElement exists, the classList will exist too (it might be an empty list, but it won't be null) https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

BenHenning: Ah. This behaves differently in Kotlin land (where I'm much more familiar). There the '?' actually cascades. In this case, `document.activeElement?.classList` would evaluate to either `null` or the class list which is why the additional `?` would be needed (since it would be trying to call `contains` on `null` rather than avoiding the call altogether).

This will take me some time to get used to. :)

cpcallen: Yeah, the way optional chaining works in JS is very pragmatic from the point of view of programmers but decidedly weird from the point of view of JS implementers.  I found the parse trees that result from expressions with optional chaining to be _extremely_ unintuitive—despite being (or perhaps because I was) already very familiar with ES5 parsing.

PR #620, opened 2025-06-25
test: add scenario with workspace and block comments

microbit-matt-hillsdon:
Aim is to illustrate a bug where a new workspace comment has different behaviour to one that's been loaded (the new Cmd+Enter shortcut seems not to work). Will also be useful once navigation is supported for comments themselves.

Demo: https://test-comments.blockly-keyboard-experimentation.pages.dev/?scenario=comments&toolbox=toolbox&renderer=zelos

### DIFF
--- a/test/loadTestBlocks.js
+++ b/test/loadTestBlocks.js
@@ -862,6 +862,121 @@ const moveTestBlocks = {
   },
 };
 
+const comments = {
+  'workspaceComments': [
+    {
+      'height': 100,
+      'width': 146.63990783691406,
+      'id': 'workspace_comment_1',
+      'x': 96.5390625,
+      'y': 531.42578125,
+      'text': 'Workspace comment',
+    },
+  ],
+  'blocks': {
+    'languageVersion': 0,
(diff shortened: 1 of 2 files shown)


test/loadTestBlocks.js
```diff
@@ -870,11 +985,12 @@ const moveTestBlocks = {
 export const load = function (workspace, scenarioString) {
   const scenarioMap = {
```
microbit-matt-hillsdon, line 986: Alphabetised so I don't have to think too hard where new ones go.

microbit-matt-hillsdon:
[Started to consider comment navigation more systematically on this doc](https://docs.google.com/document/d/10xXnx2QyCKpGeLEwUipdBVy4DHVzpH1N7hstYESVfX4/edit?tab=t.0#heading=h.ez9febbvnzd7), will get more micro:bit team input tomorrow. But I think we need to do something soon about the fact you can get trapped in block and workspace comments if you don't use the mouse.

RoboErikG (approved):

Issue #614, opened 2025-06-20
Make workspace comments keyboard-navigable


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Request

As of https://github.com/google/blockly/pull/9154

workspace comments are focusable & usable for mouse users

additional work is needed for keyboard use - following from @microbit-matt-hillsdon:

> Some way to navigate away from a comment to the workspace seems essential for keyboard nav users. Once you have a workspace or block comment focused (add one via context menu, press tab) you can't move to a block without resorting to the mouse. At least when the textarea has focus, arrows aren't an option!
> 
> The focus position is subtle after adding a workspace comment and not visible after adding a block comment. Unclear to me whether it should be on the textarea or not, especially for workspace comments with the collapse/delete button.
> 
> There's a weird scenario where tab in a comment textarea generally leaves the workspace, except if you have a workspace comment and a block comment and the focus is on the workspace comment. In that case it navigates to the block comment. If you do it repeatedly (clicking back in the workspace comment) you go to a different block comment each time in sequence.

I think for finishing editing a comment, we should use the same keys that apply to the multiline text input field (since that's a plugin, i'm not sure if it has been updated to be keyboard-navigable or not). my thoughts would be escape should work, and possibly ctrl/cmd+enter, which is what the platforms i'm familiar with use to commit text when the enter key adds a return.

The focus style is css so customizable by host applications but would certainly accept updates to the default style especially before the next release

microbit-matt-hillsdon:
[Some notes on this doc](https://docs.google.com/document/d/10xXnx2QyCKpGeLEwUipdBVy4DHVzpH1N7hstYESVfX4/edit?tab=t.0#heading=h.17zwshndk3mu).

PR #625, opened 2025-06-27
feat: Add support for acting on workspace comments and icons.

gonfunko:
In conjunction with https://github.com/google/blockly/pull/9182, this PR adds support for editing workspace comments, activating their icons via keyboard navigation, and moving them.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -127,15 +127,10 @@ export class EnterAction {
       return !workspace.isReadOnly();
     }
     // Returning true is sometimes incorrect for icons, but there's no API to check.
-    if (
-      curNode instanceof BlockSvg ||
+    return curNode instanceof BlockSvg ||
       curNode instanceof icons.Icon ||
       curNode instanceof comments.CommentBarButton ||
-      curNode instanceof comments.RenderedWorkspaceComment
-    ) {
-      return true;
-    }
-    return false;
+      curNode instanceof comments.RenderedWorkspaceComment;
   }
 
   /**
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -130,7 +130,7 @@ export class EnterAction {
     if (
       curNode instanceof BlockSvg ||
       curNode instanceof icons.Icon ||
-      curNode instanceof comments.CommentIcon ||
+      curNode instanceof comments.CommentBarButton ||
       curNode instanceof comments.RenderedWorkspaceComment
     ) {
       return true;
@@ -171,11 +171,10 @@ export class EnterAction {
         cursor?.in();
       });
       return true;
-    } else if (curNode instanceof comments.CommentIcon) {
+    } else if (curNode instanceof comments.CommentBarButton) {
       curNode.performAction();
       return true;
     } else if (curNode instanceof comments.RenderedWorkspaceComment) {
-      curNode.setEditable(true);
       curNode.setCollapsed(false);
       getFocusManager().focusNode(curNode.getEditorFocusableNode());
       return true;
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -16,6 +16,8 @@ import {
   icons,
   FocusableTreeTraverser,
   renderManagement,
+  comments,
+  getFocusManager,
 } from 'blockly/core';
 
 import type {Block} from 'blockly/core';
@@ -124,9 +126,15 @@ export class EnterAction {
     ) {
       return !workspace.isReadOnly();
     }
-    if (curNode instanceof BlockSvg) return true;
     // Returning true is sometimes incorrect for icons, but there's no API to check.
-    if (curNode instanceof icons.Icon) return true;
+    if (
+      curNode instanceof BlockSvg ||
+      curNode instanceof icons.Icon ||
+      curNode instanceof comments.CommentIcon ||
+      curNode instanceof comments.RenderedWorkspaceComment
+    ) {
+      return true;
+    }
     return false;
   }
 
@@ -162,6 +170,14 @@ export class EnterAction {
       renderManagement.finishQueuedRenders().then(() => {
         cursor?.in();
       });
+      return true;
+    } else if (curNode instanceof comments.CommentIcon) {
+      curNode.performAction();
+      return true;
+    } else if (curNode instanceof comments.RenderedWorkspaceComment) {
+      curNode.setEditable(true);
+      curNode.setCollapsed(false);
+      getFocusManager().focusNode(curNode.getEditorFocusableNode());
       return true;
     }
     return false;
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -127,10 +127,12 @@ export class EnterAction {
       return !workspace.isReadOnly();
     }
     // Returning true is sometimes incorrect for icons, but there's no API to check.
-    return curNode instanceof BlockSvg ||
+    return (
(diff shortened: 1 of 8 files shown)


src/actions/enter.ts
```diff
@@ -163,6 +171,14 @@ export class EnterAction {
         cursor?.in();
       });
       return true;
+    } else if (curNode instanceof comments.CommentIcon) {
+      curNode.performAction();
+      return true;
+    } else if (curNode instanceof comments.RenderedWorkspaceComment) {
+      curNode.setEditable(true);
```
maribethb: why change the editable state? I think this is serialized, so someone could have placed an uneditable workspace comment e.g. as part of a tutorial. I don't think you should change that state. Maybe instead check if it's not editable and return false?

gonfunko: Good catch, this was leftover from a hack I got rid of. Removed.

src/actions/enter.ts
```diff
@@ -124,9 +126,15 @@ export class EnterAction {
...
-    if (curNode instanceof icons.Icon) return true;
+    if (
+      curNode instanceof BlockSvg ||
+      curNode instanceof icons.Icon ||
+      curNode instanceof comments.CommentBarButton ||
+      curNode instanceof comments.RenderedWorkspaceComment
+    ) {
+      return true;
+    }
     return false;
```
BenHenning:
```suggestion
    return curNode instanceof BlockSvg ||
      curNode instanceof icons.Icon ||
      curNode instanceof comments.CommentBarButton ||
      curNode instanceof comments.RenderedWorkspaceComment;
```

Could we do this, instead?

BenHenning (commented):
Generally looks good, though is it possible to add tests for both comment cases?

gonfunko:
This PR fixes #614.

gonfunko:
Tests have been added.

test/webdriverio/test/test_setup.ts
```diff
@@ -263,6 +263,27 @@ export async function focusOnBlock(
...
+ * @param commentId The ID of the workspace comment to select.
+ */
+export async function focusOnWorkspaceComment(
+  browser: WebdriverIO.Browser,
+  commentId: string,
+) {
+  return await browser.execute((commentId) => {
+    const workspaceSvg = Blockly.getMainWorkspace() as Blockly.WorkspaceSvg;
+    const comment = workspaceSvg.getCommentById(commentId);
+    if (!comment)
```
BenHenning: Nit: Please add braces for convention.

test/webdriverio/test/workspace_comment_test.ts
```diff
@@ -0,0 +1,216 @@
...
+import {Key} from 'webdriverio';
+
+suite('Workspace comment navigation', function () {
+  // Setting timeout to unlimited as these tests take a longer time to run than most mocha test
+  this.timeout(0);
+
+  // Setup Selenium for all of the tests
+  setup(async function () {
+    this.browser = await testSetup(testFileLocations.NAVIGATION_TEST_BLOCKS);
+    [this.comment1, this.comment2] = await this.browser.execute(() => {
```
BenHenning: Perhaps instead: ``commentId1`` and ``commentId2`` since they're IDs, not comment objects.

test/webdriverio/test/workspace_comment_test.ts
```diff
@@ -0,0 +1,216 @@
...
+    await keyLeft(this.browser);
+    const focusedNodeId = await getCurrentFocusNodeId(this.browser);
+    chai.assert.equal(focusedNodeId, `${this.comment1}_collapse_bar_button`);
+  });
+
+  test('Activate workspace comment button', async function () {
+    await focusOnWorkspaceComment(this.browser, this.comment1);
+    await keyRight(this.browser);
+    await this.browser.keys(Key.Enter);
+    await this.browser.pause(PAUSE_TIME);
```
BenHenning: I think you can use `sendKeyAndWait` instead for a slight simplification:

https://github.com/google/blockly-keyboard-experimentation/blob/a51caa5089fdcf6c9456b6ecc8a7c5c912d604e5/test/webdriverio/test/test_setup.ts#L506

Ditto for all other cases below.

test/webdriverio/test/workspace_comment_test.ts
```diff
@@ -0,0 +1,216 @@
...
+  test('Navigate backward from workspace comment button to workspace comment button', async function () {
+    await focusOnWorkspaceComment(this.browser, this.comment1);
+    await keyRight(this.browser);
+    await keyRight(this.browser);
+    await keyLeft(this.browser);
+    const focusedNodeId = await getCurrentFocusNodeId(this.browser);
+    chai.assert.equal(focusedNodeId, `${this.comment1}_collapse_bar_button`);
+  });
+
+  test('Activate workspace comment button', async function () {
```
BenHenning, line 142: Maybe add a test for toggling to show that it can be expanded.

BenHenning (approved):
Thanks @gonfunko. LGTM, though happy to take another pass if you'd like. I did have a few suggestions, but I think they're all hopefully non-controversial.

microbit-matt-hillsdon:
Feedback from @microbit-grace testing workspace comments nav today:

- there's a tab stop for a new workspace comment textarea immediately after creation, is that still intended?
- we think the constrained/unconstrained distinction doesn't make sense for workspace comments so move mode should just do unconstrained without a modifier
- focus styles are particularly unclear (but can fix in MakeCode CSS)

microbit-matt-hillsdon:
I've used the following styles in MakeCode on our preview branch which includes this PR. Maybe use the same blue as for `--blockly-active-tree-color` in the plugin if there's enough contrast (MakeCode uses black for that due to a blue top nav bar).

```css
/* Comments are yellow default highlight is orange */
--blockly-active-workspace-comment-color: black;
```
...
```css
/* Workspace comments */
.blocklyKeyboardNavigation .blocklyComment.blocklyActiveFocus .blocklyCommentHighlight {
    stroke: var(--blockly-active-workspace-comment-color);
    stroke-width: 4px;
}
.blocklyCommentTopbar .blocklyActiveFocus {
    outline: 2px solid var(--blockly-active-workspace-comment-color);
    border-radius: 2px;
    outline-offset: -1px;
}
```

image

image

Yellow on yellow or the orange outline seem too subtle.

microbit-matt-hillsdon:
Also noticed that comments are not scrolled into view when you navigate to them.

test/webdriverio/test/workspace_comment_test.ts
gonfunko: I think that'd be better handled in the core workspace comment tests; my goal here is just to ensure that the keyboard nav infrastructure is hooked up to the underlying views/buttons.

test/webdriverio/test/workspace_comment_test.ts
gonfunko: Done.

test/webdriverio/test/workspace_comment_test.ts
gonfunko: Done.

test/webdriverio/test/test_setup.ts
gonfunko: Done, would be great if the autoformatter could do this.

gonfunko:
Thanks for the feedback! Agreed that all broadly sounds reasonable; I'll merge this as-is to unblock #626 and address those points in a followup later today.

PR #628, opened 2025-07-01
fix: remove undo and redo shortcuts in favor of those in core

### LINKED ISSUE #548: KeyboardNavigation.dispose() does not fully clean up, leading to shortcut collision on reinitialization
### Check for duplicates

### Description

When calling `KeyboardNavigation.dispose()` and then creating a new `KeyboardNavigation` instance, a crash occurs due to a collision in keyboard shortcut registration, specifically for the "undo" and "redo" actions.
Blockly throws an error for trying to register keyboard shortcuts that already exist:

```
Uncaught (in promise) Error: Shortcut named "undo" collides with shortcuts "undo,undo"
```

`UndoRedoAction` backs up the original shortcuts before registering the new ones, but it doesn't actually unregister the old ones first. The root of the issue is that `UndoRedoAction.uninstall()` attempts to re-register the original shortcuts, but those don’t have `allowCollision`, causing registration to fail. Note that there are two 'undo' shortcuts in the error shown above because plugin only installs a new shortcut. Creating the instance should unregister the original shortcuts. Disposing of a `KeyboardNavigation` instance should also unregister all shortcuts cleanly.

 I suggest updating `UndoRedoAction` to explicitly unregister the undo/redo shortcuts before registering the keyboard navigation versions and before restoring the original ones.

In the meantime, we can workaround this locally by unregister undo and redo from the `ShortcutRegistry`. We have to do this before creating the instance (to remove the originals) and before disposing (to remove the plugin shortcuts).

### Reproduction steps

1. Create an instance of `KeyboardNavigation`.
2. Call `.dispose()` on that instance.
3. Create a second instance of `KeyboardNavigation`.

### Stack trace

```shell
addKeyMapping	@	shortcut_registry.ts:117
register	@	shortcut_registry.ts:64
uninstall	@	undo_redo.ts:70
dispose	@	navigation_controller.ts:276
dispose	@	index.ts:138
```

### Screenshots

Image

### Browsers

_No response_

maribethb:
Fixes #548 

Depends on https://github.com/google/blockly/pull/9194

Removes the shortcut from kbe now that the precondition in core matches the override done here.


cpcallen (approved):
Yes!  Delete it all!!!!11!

RoboErikG (approved):

PR #631, opened 2025-07-02
fix: dont publish test directory, update ghpages

### LINKED ISSUE #580: No valid typings in published package
Although the published package contains `.d.ts` files, `package.json` isn't set up so the package counts as having types.

If I adapt a copy of the plugin test folder to use the released plugin I can only get it to work by supplying my own types. Otherwise I get:

```
➜  test git:(main) ✗ npx tsc
src/main.ts:11:36 - error TS7016: Could not find a declaration file for module '@blockly/keyboard-navigation'. '/Users/mth/tmp/test/node_modules/@blockly/keyboard-navigation/dist/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/blockly__keyboard-navigation` if it exists or add a new declaration (.d.ts) file containing `declare module '@blockly/keyboard-navigation';`

11 import { KeyboardNavigation } from '@blockly/keyboard-navigation';
                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in src/main.ts:11
```

Repro:
1. `git clone git@github.com:microbit-matt-hillsdon/test-blockly-keyboard-navigation.git` # [this project](https://github.com/microbit-matt-hillsdon/test-blockly-keyboard-navigation)
2. `cd test-blockly-keyboard-navigation`
3. `npm i`
4. `rm src/blockly-keyboard-navigation.d.ts` # local types
5. `npx tsc`

If you skip step 4 then the project compiles with tsc and you can see a working dev server via `npm run dev`.

One option to review the types is [this tool](https://arethetypeswrong.github.io/?p=%40blockly%2Fkeyboard-navigation%401.0.0) (node16 might not be a priority as not sure blockly itself is in a good place here).

maribethb:
Fixes #580 

- Changes the tsconfig to only compile the `src` directory which in turn means only the src directory is published.
- Updates the ghpages workflow to use the `predeploy` script instead of maintaining a separate webpack config

Tested via:
- Running the workflow on my own fork, results here: https://maribethb.github.io/blockly-keyboard-experimentation/
- Running `npm pack` and uploading the results to https://arethetypeswrong.github.io/


.github/workflows/pages.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         uses: actions/upload-pages-artifact@v3
         with:
           # Upload build folder
-          path: './blockly-keyboard-experimentation/dist'
+          path: './blockly-keyboard-experimentation/build'
```
cpcallen, line 67: AFAICT `blockly-scripts predeploy` should only be putting output in `build/` if it's being run in development mode, rather than production mode.

Do we actualy want to be run it in development mode when building GitHub pages?

gulpfile.js
```diff
@@ -0,0 +1,34 @@
...
+  fs.writeFileSync('build/index.html', testPage, 'utf-8');
+  console.log(
+    `Open the 'build/index.html in a browser to see results, or upload the 'build' directory to ghpages.`,
+  );
+  return done();
+}
+
+module.exports = {
+  prepareGhPages,
+};
```
cpcallen: It seems silly to add a dependency on gulp just to run a script that doesn't use any gulp features.  You could just replace the contents of this file with the following (and rename it to e.g. `scripts/deploy.js`:

```suggestion
console.log(`Preparing test page for gh-pages deployment.`);

execSync(`npm run build && npm run predeploy`, {stdio: 'pipe'});

// Copy test/index.html to build/ directory.
// Update the path at which the test_bundle can be found.
let testPage = fs.readFileSync('./test/index.html').toString();
testPage = testPage.replace('../build/test_bundle.js', 'test_bundle.js');
fs.writeFileSync('build/index.html', testPage, 'utf-8');
console.log(
    `Open the 'build/index.html in a browser to see results, or upload the 'build' directory to ghpages.`,
);
```

and then change the NPM script to invoke it directly as noted below.

package.json
```diff
@@ -5,12 +5,12 @@
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
     "build": "blockly-scripts build",
-    "ghpages": "webpack",
     "clean": "blockly-scripts clean",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "ghpages": "gulp prepareGhPages",
```
cpcallen:
```suggestion
    "ghpages": "node scripts/deploy.js",
```

package.json
```diff
@@ -64,6 +64,7 @@
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-jsdoc": "^46.8.2",
     "globals": "^15.4.0",
+    "gulp": "^5.0.1",
```
cpcallen:
```suggestion
```

cpcallen (approved):
Approving this since even as-is this PR is a strict improvement.  I do have two suggestions, however:

1. What's different about this plugin's setup compared to the other pugins?  Presumably at least some of them are written in TypeScript, have published typings, are published without including the contents of `test/`, and have GitHub page to demo them.  Is it not possible to fix #580 (and make sure we publish only the plugin itself) by bringing this package's setup more closely in line with those ones?
2. If that's not straight-forward for some reason, then please try to avoid adding an unnecessary dependency on `gulp`—`fs` and `execSync` work just fine from any old Node script!

.github/workflows/pages.yml
maribethb: Yes we do because we want the test bundle that includes the test page JavaScript. The custom webpack script that was here before was effectively doing development mode.

maribethb:
Thanks, I’ll remove gulp. In my defense when I started I thought it was going to be a lot more complicated than it turned out to be :p

This is moving this package as much inline with other plugins as it can be. The other plugins are put on GitHub pages in a very different manner, so we can’t just reuse that setup. The GitHub pages setup in samples is a lot more complicated and builds the entire site then puts it manually in the gh-pages directory. This plugin is much simpler and this is basically the bare minimum required to simply host the test page on GitHub pages.

gulpfile.js
maribethb: Done, thanks!

PR #634, opened 2025-07-02
chore: lint for floating promises

microbit-matt-hillsdon:
Fix missing awaits in the webdriver tests.

Needed tsconfig information so I've added the project setting - see https://typescript-eslint.io/blog/parser-options-project-true/#introducing-true

It's all but inevitable that folks miss some awaits in webdriver tests and it's easy to mask the issue most of the time with a pause leading to tests that are both slow and flaky.

This is just a suggestion, feel free to close it if the team finds this to be more trouble than its worth and I'll open a version that just fixes the missing awaits we have right now. I personally find this kind of support really useful when writing async code like the webdriver tests.

Rules:
- https://typescript-eslint.io/rules/no-floating-promises
- https://typescript-eslint.io/rules/no-misused-promises

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -259,4 +259,17 @@ module.exports = [
       '@typescript-eslint/no-require-imports': 'error',
     },
   },
+  {
+    files: ['test/webdriverio/test/*.ts'],
+    languageOptions: {
+      parserOptions: {
+        project: true,
+        tsconfigRootDir: __dirname,
+      },
+    },
+    rules: {
+      '@typescript-eslint/no-floating-promises': 'error',
+      '@typescript-eslint/no-misused-promises': 'error',
+    },
+  },
 ];
--- a/test/webdriverio/test/keyboard_mode_test.ts
+++ b/test/webdriverio/test/keyboard_mode_test.ts
@@ -86,7 +86,7 @@ suite(
 
       chai.assert.isFalse(await isKeyboardNavigating(this.browser));
 
-      this.browser.execute(() => {
+      await this.browser.execute(() => {
         Blockly.keyboardNavigationController.setIsActive(true);
       });
 
@@ -106,7 +106,7 @@ suite(
 
       chai.assert.isFalse(await isKeyboardNavigating(this.browser));
 
-      this.browser.execute(() => {
+      await this.browser.execute(() => {
         Blockly.keyboardNavigationController.setIsActive(true);
       });
 
@@ -126,7 +126,7 @@ suite(
 
       await this.browser.pause(PAUSE_TIME);
       // Right click a block
-      clickBlock(this.browser, 'controls_if_1', {button: 'right'});
+      await clickBlock(this.browser, 'controls_if_1', {button: 'right'});
       await this.browser.pause(PAUSE_TIME);
 
       chai.assert.isFalse(await isKeyboardNavigating(this.browser));
--- a/test/webdriverio/test/scroll_test.ts
+++ b/test/webdriverio/test/scroll_test.ts
@@ -24,7 +24,7 @@ suite('Scrolling into view', function () {
   setup(async function () {
     this.browser = await testSetup(testFileLocations.BASE);
     // Predictable small window size for scrolling.
-    this.browser.setWindowSize(800, 600);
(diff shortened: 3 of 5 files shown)


tsconfig.json
```diff
@@ -13,5 +13,5 @@
   // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
   // Only src matters for production builds.
   "include": ["./src", "./test"],
-  "exclude": ["./test/webdriverio/"]
+  "exclude": ["./test/webdriverio/test"]
```
microbit-matt-hillsdon: Before this change, test/webdriverio/index.ts was covered by neither tsconfig.json file.

microbit-matt-hillsdon:
Fixed conflicts.

tsconfig.json
maribethb: I'm removing this line in #631 as we don't want the test output in the published package. Does that file *need* to be covered by one of the tsconfig.json files? I think since webpack is compiling it it doesn't, but I haven't looked in depth at how the wdio tests are set up -- I'm surprised to see we invoke tsc directly instead of letting webpack handle it.

microbit-matt-hillsdon: I think the tsc build is for the wdio tests themselves rather than the app under test.

Your change leaves the demo app and the test app used for wdio tests without a matching tsconfig. I think they both should have one.

The downside is that tooling might use default configuration (e.g. IDEs, not sure what your webpack config does).

It  came up on this PR because the eslint rules that need type information (like the floating promise ones) choose to fail hard in this scenario:

```
/Users/mth/Development/microbit/makecode/blockly-keyboard-experimentation/test/index.ts
  0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/test/index.ts` using `parserOptions.project`: /users/mth/development/microbit/makecode/blockly-keyboard-experimentation/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

There are equivalent errors for the other app files.

I think it is made more confusing than it needs to be by nesting all the projects inside each other. There is a demo/test app in "test/" containing another app used for the webdriver io tests in webdriverio/ which in turn contains the actual tests in another "test/".

I think if this was flattened into e.g. "demo", "wdio/app" and "wdio/tests" where "wdio" itself was empty and each of those folders contained a tsconfig it would be less puzzling and each could have a tsconfig with no need for excludes.

I'll revisit this later in the week, it likely isn't worth the churn now.

maribethb: OK, thanks for the info. In blockly-samples the test page (which I think is what you're calling the demo app) are not covered by our tsconfig. Maybe that's an error with our setup for the other plugins, but it works because the webpack config compiles the `test` directory anyway (for development and test mode) and we never invoke tsc directly. 

Flattening out the structure of the test directory seems fine, but will think harder about it later when the release is done.

microbit-matt-hillsdon: I've pushed and rebased a simplified version of this change that only adds the new lint checks to the wdio tests to decouple it from the tsconfig.json discussion because on reflection I don't think it's that helpful for me to propose changes here.

I'd tried to do sometime simpler like this initially and somehow convinced myself that a bigger config change was needed but the different sections with different file globs combine just as you'd hope.

maribethb:
@microbit-matt-hillsdon Is this ready to be reviewed again? I had been thinking that you said you would hold off on it, but re-reading the latest comments I might have been mistaken. Sorry for letting it sit!

microbit-matt-hillsdon:
> @microbit-matt-hillsdon Is this ready to be reviewed again? I had been thinking that you said you would hold off on it, but re-reading the latest comments I might have been mistaken. Sorry for letting it sit!

Yes, I think this is good to review now and narrower in scope than when previously reviewed.

maribethb (approved):
Thanks, this is much more manageable from a config perspective and will definitely make tests less flaky!

PR #637, opened 2025-07-03
refactor(CSS): move plugin styling from `test/index.html` to `src/index.js`

### LINKED ISSUE #595: include necessary styles in plugin source code
### Check for duplicates

### Description

All the css styles we provide for the test page are in the test page code, not the plugin source code. So if you use the plugin you won't get any default styling and some things look pretty bad with no styling.

We should move the css to the plugin source instead of the test page. We might need a flag that controls whether or not to include the default styles, but we might not since it's css and can be overridden. This is probably a breaking change to include them by default, but it's what we should do.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

cpcallen:
### Resolves

Fixes #595.

### Proposed Changes

Move styling of:

* Comment delete icon
* Focus indicators
* Context menu items

from `test/index.html` to `src/index.ts`.

Additionally, via google/blockly#9201, move a fix for the toolbox focus styling from the plugin to core.

### Reason for Changes

It should not be necessary to manually add styling to projects using the `@blockly/keyboard-navigation` plugin in order for the plugin to function as expectd.

### Test Coverage

Manually verified that there were no obvious styling differences after making these changes.

### Additional Information

Much of the styling added to `src/index.ts`, particularly the focus-related styling, should be moved to core.  Unfortunately because it intersects with the existing styling of rendered elements, it should probably be done by breaking it up and moving it to appropriate places in `core/renderers/`, rather than just dumping it in `core/css.ts`, and I'm just not sure where it should all go yet, so in the interest of expediency I am satisfying #595 and will file a separate issue for the move to core.


cpcallen:
@microbit-matt-hillsdon: Please take a look at this and google/blockly#9201.

cpcallen:
I can't enable auto-squash on this but am happy to have it merged on my behalf once successfully reviewed.

gonfunko:
Rather than having this run at load, can the registering be moved into a (static?) method `registerKeyboardNavigationStyles()` or the like? We've regularly run into issues with accidental re-registering of styles, which IIRC throws if it happens after `inject()` is called, and also issues with multiple `KeyboardNavigation` instances.

rachel-fenichel:
> Rather than having this run at load, can the registering be moved into a (static?) method `registerKeyboardNavigationStyles()` or the like? We've regularly run into issues with accidental re-registering of styles, which IIRC throws if it happens after `inject()` is called, and also issues with multiple `KeyboardNavigation` instances.

If I recall correctly, this has to be run at load (or in a static method called on load) because you have to call `Css.register` before inject.

gonfunko:
That's correct, but I think confining it to a static method that needs to be called will avoid the other issues and is helpful on net.

rachel-fenichel:
Okay, Maribeth asked me to rebase this and make that change, so I'm also assigning it to myself.

rachel-fenichel:
Replaced by https://github.com/google/blockly-keyboard-experimentation/pull/650 (to complete the change while Christopher is offline).

PR #646, opened 2025-07-07
fix: Allow workspace comments to be moved in constrained mode.

gonfunko:
This PR allows workspace comments to be moved in "constrained" mode via the arrow keys without modifiers. Since workspace comments can be placed anywhere, this just passes through to unconstrained movement. This addresses a point of feedback raised in https://github.com/google/blockly-keyboard-experimentation/pull/625.

### DIFF
--- a/src/actions/mover.ts
+++ b/src/actions/mover.ts
@@ -318,6 +318,10 @@ export class Mover {
     const info = this.moves.get(workspace);
     if (!info) throw new Error('no move info for workspace');
 
+    if (info.draggable instanceof comments.RenderedWorkspaceComment) {
+      return this.moveUnconstrained(workspace, direction);
+    }
+
     info.dragger.onDrag(
       info.fakePointerEvent('pointermove', direction),
       info.totalDelta.clone().scale(workspace.scale),
--- a/test/webdriverio/test/workspace_comment_test.ts
+++ b/test/webdriverio/test/workspace_comment_test.ts
@@ -189,7 +189,7 @@ suite('Workspace comment navigation', function () {
     );
   });
 
-  test('Workspace comments can be moved', async function () {
+  test('Workspace comments can be moved in unconstrained mode', async function () {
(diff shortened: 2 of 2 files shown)


src/actions/mover.ts
```diff
@@ -318,6 +318,10 @@ export class Mover {
     const info = this.moves.get(workspace);
     if (!info) throw new Error('no move info for workspace');
 
+    if (info.draggable instanceof comments.RenderedWorkspaceComment) {
```
BenHenning, line 321: I assume tests already account for this?

BenHenning (approved):
It'd be nice to see at least 1 test that would break without this change, but going ahead and approving to unblock.

src/actions/mover.ts
gonfunko: Added a test that moves comments without the Alt modifier, so we have coverage for both (identical) move modes.

test/webdriverio/test/workspace_comment_test.ts
```diff
@@ -203,4 +203,19 @@ suite('Workspace comment navigation', function () {
...
+
+  test('Workspace comments can be moved in constrained mode', async function () {
+    await focusOnWorkspaceComment(this.browser, this.commentId1);
+
+    const initialPosition = await this.getCommentLocation(this.commentId1);
+    chai.assert.deepEqual(initialPosition, [200, 200]);
+
+    await sendKeyAndWait(this.browser, 'm');
+    await sendKeyAndWait(this.browser, Key.ArrowUp, 2);
+    await sendKeyAndWait(this.browser, Key.ArrowLeft);
```
BenHenning: Nit: can use `keyUp` and `keyLeft` for these, instead.

BenHenning (approved):
Thanks for adding the test!

test/webdriverio/test/workspace_comment_test.ts
gonfunko: Done.

PR #649, opened 2025-07-07
chore: Add new tests for ephemeral focus

BenHenning:
This adds 3 new tests for cases that were explicitly found as broken in https://github.com/google/blockly/issues/9203 to accompany the fixes in https://github.com/google/blockly/pull/9213.

These tests have been validated as failing without the fix from https://github.com/google/blockly/pull/9213 in place.

These are being tested here instead of in core Blockly because they are more integration tests than testing specific components, and https://github.com/google/blockly/pull/9213 includes a more targeted test for the specific regression that was fixed.

Some notes:
- The drop-down test fails on its own without the fix, but not when run with other tests. Surprisingly this means that there's some state bleeding between tests, but it's not clear where exactly this is coming from.
- There are some unrelated cleanups in the tests to make things a bit nicer.
- There's a new method for validating ephemeral focus (though it ought to be rarely used when testing at this level).
- `getFocusedFieldName` was updated to properly account for a field not actually being focused for better error propagation in a failing test.


BenHenning (commented):
Quick spot check self-review.

BenHenning:
(Retriggering CI with an empty commit in order to validate these do pass correctly now that https://github.com/google/blockly/pull/9213 is merged).

RoboErikG (approved):

BenHenning:
FYI I will be adding more tests to this (actually did locally, but one is flaky and need more investigation) before asking for a re-review.

PR #651, opened 2025-07-08
test: move registerCommentOptions before plugin

### LINKED ISSUE #648: webdriver test weirdness with duplicate shortcut
### Check for duplicates

### Description

In #642 a duplicate shortcut was added. this code required the workspace comment options to be registered before the duplicate options were registered. this was fixed correctly in the test page for the plugin, but in the webdriver test page, there are now *two* calls to `registerCommentOptions` which I missed during review. registering them twice throws an error and causes all the webdriver tests to fail.

Aaron removed one of those register calls in https://github.com/google/blockly-keyboard-experimentation/commit/0d6acb802609d9e24484c1915d8b7885351437c6 . I would have expected this to fail, because according to the PR above, the comment options need to be registered *before* the new Duplicate action is run (which happens with `new KeyboardNavigation()`).

However, aaron said that some menu-related tests failed if the comment options were registered before the plugin instead of after. 

This needs investigation + cleanup to make sure the tests are testing the right thing. @microbit-matt-hillsdon fyi

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

microbit-matt-hillsdon:
Update tests so contextMenuExists does not care whether there's a shortcut defined.

Separately assert the full context menu contents including shortcuts for blocks in the workspace and toolbox, workspace comments and the workspace itself.

Add a space to separate the keyboard shortcut from the context menu label as it was only visually separated and a screen reader would have heard e.g. "DuplicateD".

Fixes #648

### DIFF
--- a/test/webdriverio/test/actions_test.ts
+++ b/test/webdriverio/test/actions_test.ts
@@ -60,9 +60,9 @@ suite('Menus test', function () {
             {'text': 'Delete 2 Blocks Delete'},
             {'text': 'Move Block M'},
             {'text': 'Edit Block contents Right'},
-            {'text': 'Cut Ctrl+X'},
-            {'text': 'Copy Ctrl+C'},
-            {'disabled': true, 'text': 'Paste Ctrl+V'},
+            {'text': 'Cut Ctrl + X'},
+            {'text': 'Copy Ctrl + C'},
+            {'disabled': true, 'text': 'Paste Ctrl + V'},
           ],
       await contextMenuItems(this.browser),
     );
@@ -91,9 +91,9 @@ suite('Menus test', function () {
         : [
             {'text': 'Help'},
             {'disabled': true, 'text': 'Move Block M'},
-            {'disabled': true, 'text': 'Cut Ctrl+X'},
-            {'text': 'Copy Ctrl+C'},
-            {'disabled': true, 'text': 'Paste Ctrl+V'},
+            {'disabled': true, 'text': 'Cut Ctrl + X'},
+            {'text': 'Copy Ctrl + C'},
+            {'disabled': true, 'text': 'Paste Ctrl + V'},
           ],
       await contextMenuItems(this.browser),
     );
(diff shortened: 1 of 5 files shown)


microbit-matt-hillsdon:
Test failures due to spaces in the Windows/Linux option formats. Will fix shortly.

microbit-matt-hillsdon:
> Test failures due to spaces in the Windows/Linux option formats. Will fix shortly.

Hopefully got this right in 824f059. Please approve re-run.

test/webdriverio/test/actions_test.ts
```diff
@@ -51,13 +79,60 @@ suite('Menus test', function () {
...
 
-    chai.assert.isTrue(
-      await contextMenuExists(this.browser, 'Help'),
-      'The menu should be openable on a block in the toolbox',
+    chai.assert.deepEqual(
+      process.platform === 'darwin'
+        ? [
+            {'text': 'Help'},
+            {'disabled': true, 'text': 'Move Block M'},
+            {'disabled': true, 'text': 'Cut ⌘ X'},
```
cpcallen, line 87: Probably a topic for a separate issue / PR, but: shouldn't this be `'⌘X'`, not `'⌘ X'`?  Or is the spurious space just an artefact of how the displayed HTML is being converted to text?

cpcallen (approved):

test/webdriverio/test/actions_test.ts
microbit-matt-hillsdon: I think there is a real space in these context menus. There is some spacing in Mac native menus... perhaps less than a full space?

image

cpcallen: Hmm, yes, you are quite correct.  That is most curious: they certainly [used to look rather different](https://i0.wp.com/morrick.me/wp-content/uploads/2021/02/001-Menu-transparency.png?ssl=1) and indeed looked more or less like from 1984 to 2021, modulo a few font changes, but apparently this changed in macOS 12, and I've not had cause to actually look at the contents of a menu on my own laptop since I upgraded to that version three years ago.

PR #660, opened 2025-07-10
test: Make tests faster, more robust, more consistent

cpcallen:
A bunch of small fixes and refactors of tests:

* fix: Have `scroll_test.ts` restore window size after it's done.
* docs: Use "WebdriverIO" instead of "Selenium" (since we don't use Selenium!)
* chore: Remove unneeded `timeout(0)` calls.
* fix: Set `PAUSE_TIME` to 0, making tests run about 50% faster by default.
* fix: Fix one flakey right-click test in `keyboard_mode_test.ts` that was previously depending on a fixed-duration pause.
* refactor: Use `sendKeyAndWait` in almost all places instead of `browser.keys`.

### DIFF
--- a/test/webdriverio/test/actions_test.ts
+++ b/test/webdriverio/test/actions_test.ts
@@ -14,6 +14,7 @@ import {
   tabNavigateToWorkspace,
   testFileLocations,
   testSetup,
+  sendKeyAndWait,
   keyRight,
   contextMenuItems,
 } from './test_setup.js';
@@ -30,8 +31,7 @@ suite('Menus test', function () {
     await tabNavigateToWorkspace(this.browser);
     await focusOnBlock(this.browser, 'draw_circle_1');
     await this.browser.pause(PAUSE_TIME);
-    await this.browser.keys([Key.Ctrl, Key.Return]);
-    await this.browser.pause(PAUSE_TIME);
+    await sendKeyAndWait(this.browser, [Key.Ctrl, Key.Return]);
 
     chai.assert.deepEqual(
       process.platform === 'darwin'
@@ -73,8 +73,7 @@ suite('Menus test', function () {
     await moveToToolboxCategory(this.browser, 'Functions');
     // Move to flyout.
     await keyRight(this.browser);
-    await this.browser.keys([Key.Ctrl, Key.Return]);
-    await this.browser.pause(PAUSE_TIME);
+    await sendKeyAndWait(this.browser, [Key.Ctrl, Key.Return]);
 
     chai.assert.deepEqual(
       process.platform === 'darwin'
@@ -97,9 +96,8 @@ suite('Menus test', function () {
   test('Menu on workspace', async function () {
     // Navigate to draw_circle_1.
     await tabNavigateToWorkspace(this.browser);
-    await this.browser.keys('w');
-    await this.browser.keys([Key.Ctrl, Key.Return]);
-    await this.browser.pause(PAUSE_TIME);
+    await sendKeyAndWait(this.browser, 'w');
+    await sendKeyAndWait(this.browser, [Key.Ctrl, Key.Return]);
 
     chai.assert.deepEqual(
       process.platform === 'darwin'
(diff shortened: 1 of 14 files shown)


test/webdriverio/test/basic_test.ts
```diff
@@ -239,23 +237,18 @@ suite('Keyboard navigation on Blocks', function () {
...
     chai.assert.isFalse(await isDragging(this.browser));
   });
 });
 
 suite('Keyboard navigation on Fields', function () {
-  // Setting timeout to unlimited as these tests take a longer time to run than most mocha test
-  this.timeout(0);
-
-  // Setup Selenium for all of the tests
+  // Clear the workspace and load start blocks.
```
maribethb, line 251: Note that unlike the rest of the test files, this one only does this before the entire suite, not before each test. Possibly worth either changing that or adding a comment.

test/webdriverio/test/scroll_test.ts
```diff
@@ -17,25 +18,31 @@ import {
 } from './test_setup.js';
 
 suite('Scrolling into view', function () {
-  // Setting timeout to unlimited as these tests take longer time to run
-  this.timeout(0);
-
-  // Clear the workspace and load start blocks
-  setup(async function () {
+  // Clear the workspace and load start blocks.
+  suiteSetup(async function () {
```
maribethb, line 22: By changing this to `suiteSetup` instead of `setup` the workspace won't be reset between tests anymore. I think the resize stuff should be in `suiteSetup` but leave the `testSetup` call in just `setup`.

test/webdriverio/test/actions_test.ts
```diff
@@ -14,14 +14,12 @@ import {
...
   testFileLocations,
   testSetup,
+  sendKeyAndWait,
   keyRight,
   contextMenuItems,
 } from './test_setup.js';
 
 suite('Menus test', function () {
-  // Setting timeout to unlimited as these tests take longer time to run
-  this.timeout(0);
```
maribethb, line 23: I do wonder about this change as it feels like there's a good chance some longer tests will take longer to run than the default, and this will get really annoying when you're debugging locally... maybe that's not a good enough reason to put unreasonably long timeouts, but I'm a bit hesitant to do something that might make it even more annoying to write tests than it already is.

maribethb (commented):
Tests are not passing, not sure if it's due to actual timeout or if the timeout is due to the test hanging after an error

test/webdriverio/test/basic_test.ts
cpcallen: Good point re: documentation.  I had noticed the `setup` vs. `suiteSetup` inconsistency.  While looking in to that I realised that the tests in this suite are almost certainly passing only because of a combination of #632 and the fact that there happen to be a total of exactly the right number of tab stops in the document that `tabNavigateToWorkspace` loops focus back to where it started.

test/webdriverio/test/scroll_test.ts
cpcallen: Since the tests in this suite are in fact destructive of the workspace contents I concur..

test/webdriverio/test/actions_test.ts
cpcallen: Hmm.  It is true that the default timeout does make debugging tests annoying.  Maybe we should set the timeout to zero iff `PAUSE_TIME` is nonzero?  ~~Or set it to a suitable multiple of `PAUSE_TIME`?~~—no that doesn't help if you've got the debugger out.

cpcallen:
> Tests are not passing, not sure if it's due to actual timeout or if the timeout is due to the test hanging after an error

Unfortunately the test failure is the same in both runs, is not due to a timeout, is not transient, and does not reproduce locally.  :-(

Now attempting to bisect by pushing some partial reverts to this branch.

test/webdriverio/test/actions_test.ts
maribethb: Setting the timeout if pause time is nonzero seems like a good idea. If for some reason that isn't possible then I would probably lean towards having the infinite timeout to make it easier to write tests personally.

cpcallen:
Superseded by RaspberryPiFoundation/blockly-keyboard-experimentation#692.

PR #663, opened 2025-07-10
Added shortcuts help to readme

mscherotter:
The readme fails to have a section to tell users to add a div to host the shortcuts popup.  Without the div, the popup does not display when pressing the / key:

`<div id="shortcuts"></div>`


maribethb (approved):
Thank you so much for this change, apologies for the delay in review!

PR #666, opened 2025-07-11
fix: use normal shortcut modifier order and prefer shorter shortcuts

### LINKED ISSUE #647: Shortcut dialog does not reflect Redo keyboard shortcut (Ctrl+Y in Windows)
### Check for duplicates

### Description

Redo shows Ctrl+Shift+Z on the dialog although Ctrl+Y is binded as well. 

### Reproduction steps

1. Press / to see the Keyboard Shortcuts dialog
2. Ctrl+Shift+Y is shown

### Stack trace

```shell

```

### Screenshots

Image Showing Redo Shortcut

### Browsers
_No response_

microbit-matt-hillsdon:
Behaviour changes:
- Sort modifiers so we don't say Shift+Cmd+Z for redo on Mac ([this change didn't help](https://github.com/google/blockly/pull/9169), presumably because of normalization in the shortcut registry)
- Ctrl+Y is the preferred redo shortcut on Windows and not all apps support the more cross-platform Ctrl+Shift+Z so more helpful to teach Ctrl+Y.
- Add guard in case plugin shortcuts are unbound.

Implementation change:
- Map to user-facing key names last to keep earlier modifier logic using the shortcut registry representation of the modifiers. This the same point we do translations of modifiers in the equivalent code in MakeCode. This made it obvious that "Alt" didn't identify a Mac shortcut so removed it from the platform check (no change in shortcut dialog). 

This PR is equivalent to changes we've had in MakeCode for a while now.

Fixes #647

Demo: https://shortcut-formatting-tweaks.blockly-keyboard-experimentation.pages.dev/

### DIFF
--- a/src/shortcut_formatting.ts
+++ b/src/shortcut_formatting.ts
@@ -79,38 +79,52 @@ function getActionShortcutsAsKeys(
   modifierNames: Record<string, string>,
 ): string[][] {
   const shortcuts = ShortcutRegistry.registry.getKeyCodesByShortcutName(action);
+  if (shortcuts.length === 0) {
+    return [];
+  }
   // See ShortcutRegistry.createSerializedKey for the starting format.
-  const named = shortcuts.map((shortcut) => {
-    return shortcut
-      .split('+')
-      .map((maybeNumeric) => keyNames[maybeNumeric] ?? maybeNumeric)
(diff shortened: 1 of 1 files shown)


gonfunko (approved):

microbit-matt-hillsdon:
Just a reminder that you need to merge.

Issue #636, opened 2025-07-02
Keyboard navigation for block comments


microbit-matt-hillsdon:
Similar to #625 and the related blockly change, it would be good to support navigation for block comments.

The most critical issue is that you can get stuck with the comment as the only keyboard reachable position in the workspace once you tab to a comment. A fix for that that left us short of full navigation would still be valuable but full navigation might be the most viable route.

MakeCode [actually has a custom implementation more like workspace comments](https://github.com/microsoft/pxt/blob/master/pxtblocks/plugins/comments/textinput_bubble.ts) so perhaps work done for workspace nav can help us fix this in MakeCode itself independently of support in the regular block comment? Any advice welcome.

[Some thoughts on comment navigation more broadly in this doc.](https://docs.google.com/document/d/10xXnx2QyCKpGeLEwUipdBVy4DHVzpH1N7hstYESVfX4/edit?tab=t.0).

microbit-matt-hillsdon:
Discussed with the MakeCode team. We're going to support Escape to collapse a block comment (in MakeCode's custom comment bubble) as an interim approach to getting stuck, then consider this further after the release.

PR #668, opened 2025-07-14
feat: Add support for keyboard navigation in block comments. 

gonfunko:
This PR fixes #[636](https://github.com/google/blockly-keyboard-experimentation/issues/636) by adding support for using keyboard navigation with block comments. Most of the logic was added to core Blockly in https://github.com/google/blockly/pull/9227, but this PR (1) enables navigating into the bubbles of all bubble-having icons, adds styling to indicate that bubbles are focused, and adds tests. The UX for interacting with block comments is:

1. Focus the comment icon.
2. Enter will then open the comment bubble and focus the comment editor.
3. Comment text can be edited as normal.
4. On Escape or Command Enter, the edits will be committed and the block comment bubble will be focused.
5. On a second Escape, the block comment bubble will be dismissed and focus will be restored to the comment icon.

The intermediate focus state on the bubble is to allow moving/resizing block comment bubbles, which is not yet implemented but will be in the future (along with the same behavior for mutator bubbles).


microbit-robert:
We have been testing block comment navigation with this PR and found that you can get into a situation where the block comment  is open and the block comment icon has focus. When pressing `Enter` or `Space` at this point, the block comment is closed, but the focus unexpectedly moves forward to the next field/input in the block. Please see the video below for an example of this:

https://github.com/user-attachments/assets/782de94f-802c-47b9-a92b-703ddd62d4f2

gonfunko:
Thanks for flagging, that's been resolved.

microbit-matt-hillsdon:
More/broader feedback [in this doc](https://docs.google.com/document/d/1wBFiFZLRKCKpx2m-7dVBbWkfs1AGabzsdf-X3tda3Q0/edit?usp=sharing).

rachel-fenichel (approved):
I'm approving this so that it's easier to play with and modify this part of navigation.

If we need to do a release and specifically don't want this, it's easy to pull out.

None of the current proposals for navigation of block comments feels fully baked right now, so we'll need more iterating on behaviour.

PR #669, opened 2025-07-14
fix: Remove null checks for cursor.

### LINKED ISSUE #471: Handle cases when the workspace's cursor is null, or disallow that state
### Check for duplicates

### Problem

_No response_

### Request

`workspace.getCursor()` can return `null`, which leads to lots of null checks throughout the code. This should be handled in a single place or `getCursor` should be made non-nullable.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
This PR fixes #471 by removing checks/conditions for a null cursor. The cursor was made always non-null in https://github.com/google/blockly/pull/9210, so these are now unnecessary.

### DIFF
--- a/src/actions/action_menu.ts
+++ b/src/actions/action_menu.ts
@@ -86,9 +86,7 @@ export class ActionMenu {
     // TODO(#362): Pass this through the precondition and callback instead of making it up.
     const menuOpenEvent = new KeyboardEvent('keydown');
 
-    const cursor = workspace.getCursor();
-    if (!cursor) throw new Error('workspace has no cursor');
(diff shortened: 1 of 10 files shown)


BenHenning (approved):
LGTM, thanks @gonfunko!

Just had one additional clean-up suggestion.

BenHenning (commented):
Ah, I didn't realize my comment hadn't posted since the line isn't part of the diff.

I suggest removing the EsLint failure in `edit.ts` for removing the `LineCursor` import since there are already cursor clean-ups happening here, anyway.

Issue #183, opened 2025-01-29
Install shortcut dialog CSS from KeyboardNavigation constructor instead of on load


rachel-fenichel:
- Wrap [this call](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/shortcut_dialog.ts#L179) to `Blockly.Css.register` in a function
- Export that function from `shortcut_dialog.ts`
- Import that function in `index.ts`
- Call that function in the `KeyboardNavigation` constructor

rachel-fenichel:
Status update: moving this to after micro:bit testing after confirming that it is not required for Code.org right now.

Calling `Blockly.Css.register` after Blockly injection is not allowed; instead, the plugin should manage its own CSS. This can be worked around by loading the plugin code only a single time, and before injection.

PR #671, opened 2025-07-15
fix: Remove side effects from keyboard-experiment.

### LINKED ISSUE #327: Remove side effects
Currently just loading this plugin has side effects, which should be avoided. These side effects include, but may not be limited to:

- loading monkey patches for toolbox and gesture
- registering a LineCursor
- registering a FlyoutCursor

These tasks should only be done when `new KeyboardNavigation()` or a similar `init` method is called.

gonfunko:
This PR fixes #327 by removing side effects associated with loading (but not instantiating) the keyboard experiment. #183 has been closed, the gesture monkeypatch is no more, `LineCursor` is handled in core, which leaves the toolbox monkeypatch and `FlyoutCursor` registration. The toolbox monkeypatch has been removed in favor of a toolbox subclass with explicit registration, and the automatic registration for the `FlyoutCursor` has been replaced with explicit registration as well. AFAICT there are no other remaining side effects.


RoboErikG (approved):

cpcallen:
Should this not have been marked as a breaking change?

PR #684, opened 2025-07-29
feat: Introduce initial screen reader support  (experimental)

BenHenning:
Fixes part of https://github.com/google/blockly/issues/3370
Fixes part of https://github.com/google/blockly/issues/9283

This PR introduces the plugin corresponding changes needed by https://github.com/google/blockly/pull/9280.

In particular, this:
- Introduces CI workflow updates so that "tip-of-tree" is redefined as the corresponding experimental branch in core for screen reader support (to ensure CI passes here for changes that happen there).
- Removes some access error overrides since the corresponding field was made public in the core PR (which is the root of the CI failures here).
- Introduces some general instructions in the navigation plugin playground for folks testing Blockly with a screen reader (and points to https://github.com/google/blockly-keyboard-experimentation/discussions/673 for discussion).

**Important**: Note that this PR's changes will NOT be merged into the `main` branch as-is. They will be reconsidered entirely (and likely largely reimplemented) separately from this experimental branch and submitted as a separate, future review for `main`. See https://github.com/google/blockly/pull/9280 and Fixes part of https://github.com/google/blockly/issues/9283 for more specifics on why the separate branch thing is being set up this way.


BenHenning:
This now includes all the active functionality from https://github.com/google/blockly/pull/9262.

src/aria_monkey_patches.js
```diff
@@ -0,0 +1,512 @@
+/**
```
microbit-matt-hillsdon: I'd like to just check that you're not doing this to make life easier for micro:bit, as there's a degree of lost clarity around the changes in the monkey patching.

If desired we can work with a long-lived branch of blockly for user testing and similar if it makes this work more manageable, rather than pushing everything into the plugin. That's effectively where we were at towards the end of the recent work (just with blockly develop).

Actually getting upgrades into MakeCode is another scenario and would need releases, but that might be a while away.

If you value this approach over the trade offs I'm not trying to dissuade you, just give options.

BenHenning: So my thinking is that it will be relatively easier to quickly iterate on aria behaviors by changing only this repo rather than needing to deal with synchronizing core and the plugin (both of which will probably require changes to get everything in a good place).

Given that most of the aria logic will need to be significantly changed for the long-term, this is pretty heavily code that will be thrown away in its current form.

I do strongly agree that this form of altering aria properties is much less readable. I took a stab at pulling apart all of the stub overrides and moving them to TypeScript for proper typing--do you find this more readable? It's still far from perfect largely because:
1. It's still monkey patching something under the surface.
2. _Where_ aria changes happen is still heavily implementation-sensitive, and is why some behaviors have actually regressed (e.g. toolbox item position and level).

It wouldn't be impossible to pivot from this back to something akin to https://github.com/google/blockly/pull/9262 in a temporary core branch if we find this approach is more painful than beneficial. I'm also definitely able to be convinced to move to the temporary branch approach.

BenHenning:
Unfortunately, this won't be able to be merged today--I decided to spend the day rearranging the implementation to try and make it a bit better given the concerns multiple folks have surfaced around the difficulty of managing so many monkey patches. I want people to be generally happy with the approach before committing to it since pivoting, while possible, will use up unnecessary time later on if we need to do that.

Beyond finalizing the decision, this code breaks existing tests (mostly due to it disabling all of core Blockly's aria logic). This is actually a good thing: it demonstrates one of the key benefits of approaching the experimentation in this way which is quickly seeing how the keyboard navigation behaviors change with aria alterations (though most of these are technically transient failures due to the tests unexpectedly depending on aria labels and roles for non-screen reader cases).

BenHenning:
Changes have been removed and moved back into core Blockly: https://github.com/google/blockly/pull/9280.

src/aria_monkey_patches.js
BenHenning: So the monkey patches have been completely removed and the corresponding ARIA changes are checked into core Blockly now (see https://github.com/google/blockly/pull/9280). As a result, marking this as resolved since there aren't any concerns left to address here (as I understand them).

gonfunko (approved):

PR #688, opened 2025-08-06
fix: clean up shortcut help dialog

### LINKED ISSUE #654: Shortcut dialog is out of date and looks a bit broken
Missing at least:
- Move

A general review vs actions seems prudent.

Out of date actions at least:
- Insert

Missing shortcuts (since clipboard rejig to delegate to core actions?):
- Cut
- Copy
- Paste

Image

Adding Move to the middle column might leave it a bit unbalanced.

maribethb:
Fixes #654 

This PR has multiple commits that may be easier to review individually.

- Rearranges the HTML to use some template tags and variables instead of concatenating HTML. I find this easier to read and make sure tags are all closed, etc.
- Hides shortcuts from the table that don't have any keys displayed. This makes the dialog more resilient to things going out of date (e.g. the insert shortcut being removed) and also makes it more useful for people who have customized things by removing certain shortcuts for whatever reason.
- Introduces a conversion between shortcut name and display name, to account for the fact that clipboard shortcuts are called `keyboard_nav_copy` etc. but shouldn't be displayed that way.
- Uses `Blockly.Msg` for the header of the dialog
- Removes the insert shortcut which no longer exists
- Adds the jump to workspace & move shortcuts
- Fixes the cut/copy/paste shortcuts so they display correctly

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -35,6 +35,7 @@ export enum SHORTCUT_NAMES {
   LEFT = 'left',
   NEXT_STACK = 'next_stack',
   PREVIOUS_STACK = 'previous_stack',
+  // Unused.
   INSERT = 'insert',
   EDIT_OR_CONFIRM = 'edit_or_confirm',
   DISCONNECT = 'disconnect',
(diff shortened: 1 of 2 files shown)


BenHenning (approved):
Thanks @maribethb! The changes largely LGTM, but I wonder if it may be worth adding a basic test for validating the contents? That might also act as a decent guard against the UI unintentionally changing in the future, even with the extra robustness you've added here.

I also wonder if tests would be a good way to verify that custom shortcut changes also correct update the dialog as expected.

I defer to you on whether you've already considered adding tests or if you think they might not be worth it here.

maribethb:
Thanks! Filed #691 so I can submit this now and work on tests next sprint. I'm not convinced it's worth a test that just asserts all the current ones are there or not, but it's worth at least making sure empty ones don't appear in the table.

PR #689, opened 2025-08-06
feat: Deploy GH pages with screen reader experiment

### LINKED ISSUE #686: Add automatic deployment for screen reader staging branch
### Check for duplicates

### Problem

https://github.com/google/blockly-keyboard-experimentation/pull/684 is the initial PR introducing work to a new, long-lived branch: `add-screen-reader-support`. It's important to be able to easily test these changes and, given that they are in a staging branch for the next number of months, one can't simply use the existing GitHub pages link (https://google.github.io/blockly-keyboard-experimentation/) or self-deploy (easily).

### Request

@microbit-matt-hillsdon suggested that we introduce a regular CI deployment for the screen reader branch.

This could be as simple as deploying a subdirectory of the GitHub pages site, e.g. https://google.github.io/blockly-keyboard-experimentation/screen-reader, that includes the pre-released changes (possibly linked against tip-of-tree core Blockly).

### Alternatives considered

Most other solutions are probably harder, including:
- Deploying outside pages.
- Deploying to a fork (which is what we initially did with https://benhenning.github.io/blockly-keyboard-experimentation/ but then it was automatically overwritten by the same actions that make the main repo work correctly).

### Additional context

_No response_

BenHenning:
Fixes #686

This updates the GH pages workflow to:
- Automatically deploy with changes to both `main` and `add-screen-reader-support-experimental` (the screen reader experimental branch--see https://github.com/google/blockly/issues/9283).
- Upload the screen reader version of the playground to a new `/screenreader` subdirectory in the plugin repo's GitHub pages deployment.

This will approximately double the amount of time needed for the GH pages deployment since it requires building two distinct copies of core Blockly and the plugin (tip-of-tree versions and the screen reader experimentation versions).

Conveniently, this will act as a fast way to see screen reader changes since they will go live a few minutes after a change to the plugin. Unfortunately, that does not reciprocate for changes to core Blockly (where most of the screen reader changes will actually occur), so oftentimes a manual run of the workflow will be needed (which should suffice in those cases).

Note that this needs to be approached in this way (building and uploading both projects for changes to either branch) since the action being used for deployment doesn't support partial deployment. See: https://github.com/actions/deploy-pages/issues/349.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,14 +6,14 @@ on:
   pull_request:
   push:
     branches:
-      - main
+      - add-screen-reader-support-experimental
 
 permissions:
   contents: read
 
 jobs:
   build_tip_of_tree_v12:
-    name: Build test (against tip-of-tree core develop)
+    name: Build test (against add-screen-reader-support-experimental core develop)
     runs-on: ubuntu-latest
     steps:
       - name: Checkout experimentation plugin
@@ -25,7 +25,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: 'google/blockly'
-          ref: 'develop'
+          ref: 'add-screen-reader-support-experimental'
           path: core-blockly
 
       - name: Use Node.js 20.x
@@ -42,7 +42,7 @@ jobs:
           npm install
           cd ..
 
-      - name: Link latest Blockly develop
+      - name: Link latest Blockly add-screen-reader-support-experimental
         run: |
           cd core-blockly
           npm run package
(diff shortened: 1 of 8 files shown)


BenHenning:
I will put this in review once I can validate it on my fork (which will require #684 to be merged, first).

BenHenning (commented):
Self-checked the code, and updated one line that was wrong.

BenHenning:
I've verified that this is working correctly in a fork: https://benhenning.github.io/blockly-keyboard-experimentation/screenreader/ deploys with https://github.com/BenHenning/blockly-keyboard-experimentation/pull/4. I'll leave that live until this PR is ready to merge (since otherwise removing my direct change to my fork's `main` branch will undo the deployment).

BenHenning (commented):
Self-reviewed latest.

gonfunko (approved):

PR #692, opened 2025-08-08
test: Make tests faster, more robust, more consistent

cpcallen:
A bunch of small fixes and refactors of tests:

* fix: Have `scroll_test.ts` restore the original browser window size after it's done.
* docs: Use "WebdriverIO" instead of "Selenium" (since we don't use Selenium!)
* chore: Make `timeout(0)` calls only when `PAUSE_TIME` is nonzero (and make them in every suite).
* fix: Set `PAUSE_TIME` to `0`, making tests run about 50% faster.
* refactor: Use `sendKeyAndWait` almost everywhere (instead of `browser.keys`).
* fix(tests): Make `tabNavigateToWorkspace` idempotent, to avoid relying on a bug #632.
* chore(tests): Remove `tabNavigateToWorkspace` calls that are immediately followed by a `focusOnBlock` call.

This is a rebase, with updates, of the changes from (and supercedes) PR #660, and incorporates changes based on the comments thereto.

### DIFF
--- a/test/webdriverio/test/test_setup.ts
+++ b/test/webdriverio/test/test_setup.ts
@@ -27,10 +27,22 @@ import {fileURLToPath} from 'url';
 let driver: webdriverio.Browser | null = null;
 
 /**
- * The default amount of time to wait during a test. Increase this to make
- * tests easier to watch; decrease it to make tests run faster.
+ * The default amount of time to wait during a test, in ms.  Increase
+ * this to make tests easier to watch; decrease it to make tests run
+ * faster.
+ *
+ * The _test.js files in this directory are set up to disable timeouts
+ * automatically when PAUSE_TIME is set to a nonzero value via
+ *
+ *     if (PAUSE_TIME) this.timeout(0);
+ *
+ * at the top of each suite.
+ *
+ * Tests should pass reliably even with this set to zero; use one of
+ * the browser.wait* functions if you need your test to wait for
+ * something to happen after sending input.
  */
-export const PAUSE_TIME = 50;
+export const PAUSE_TIME = 0;
 
 /**
  * Start up WebdriverIO and load the test page. This should only be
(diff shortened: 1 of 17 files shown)


test/webdriverio/test/test_setup.ts
```diff
@@ -440,8 +455,14 @@ export async function tabNavigateToWorkspace(
   hasToolbox = true,
   hasFlyout = true,
 ) {
-  // Navigate past the initial pre-injection focusable div element.
-  await tabNavigateForward(browser);
+  // Move focus to initial pre-injection focusable div element.
+  //
+  // Ideally we'd just rest focus state to the state it is in when the
```
maribethb: nit: reset

test/webdriverio/test/workspace_comment_test.ts
```diff
@@ -23,10 +23,7 @@ import {
 import {Key} from 'webdriverio';
 
 suite('Workspace comment navigation', function () {
-  // Setting timeout to unlimited as these tests take a longer time to run than most mocha test
-  this.timeout(0);
```
maribethb, line 27: Should this test get the timeout extended for nonzero pause time?

test/webdriverio/test/toast_test.ts
```diff
@@ -9,6 +9,7 @@ import * as Blockly from 'blockly/core';
 import {PAUSE_TIME, testFileLocations, testSetup} from './test_setup.js';
 
 suite('HTML toasts', function () {
+  // Clear the workspace and load start blocks.
```
maribethb, line 15: Should this test get the timeout extended for nonzero pause time?

maribethb (approved):
also 1 lint warning

cpcallen:
@maribethb: If this LGTY then please go ahead and merge it so it does not get stale again.

test/webdriverio/test/workspace_comment_test.ts
cpcallen: Good point.  I think I missed this because I initially deleted all the `timeout(0)` calls and may have only reinstated the ones for suites that had tests fail when I set `PAUSE_TIME` to a large value.

test/webdriverio/test/toast_test.ts
cpcallen: Yup, here too.

Issue #699, opened 2025-08-22
Move tests: finishing a move


cpcallen:
For blocks, test that:
* [x] Pressing enter finishes a move.
  * [ ] If there is no candidate connection, the block becomes a top block on the workspace at its current location.
  * [x] If there is a candidate connection, the block is attached in the expected position.
  * [ ] If there was an existing block attached to the candidate connection, that heal behaviour matches that expected for mouse drags:
    * [ ] Stacks are healed.
    * [ ] A row is healed iff the value block dragged has exactly one value input.
* [x] Pressing esc aborts a move and returns the block to its original location.

For:
* [X] Workspace comments
* [ ] Bubbles

Test that:
* [X] Pressing enter finishes a move.
  * [ ] The item being moved remains on the workspace at its current location.
* [ ] Pressing esc aborts a move and returns the item to its original location.

cpcallen:
I noticed what I thought was a difference in behaviour between mouse drags and keyboard moves: when placing a block as the new top block of a stack (either on top of an existing stack, or for blocks with a statement input _around_ an existing stack), a keyboard drag cause the existing stack to shift position slightly.

That seemed like a bug so I added a few test items to detect it:

  * [ ] If there is a candidate connection, that the moving block is positioned at the same coordinates as the insertion marker.  Specifically, check that this is true when the moving block:
    * [ ] is the new top block of a stack,
    * [ ] is inline in a stack,
    * [ ] is at bottom of stack,
    * [ ] surrounds an existing stack (i.e. candidate connection was on a statement input of the moving block), and
    * [ ] is a value block.

Unfortunately upon closer examination I realise that keyboard drags behave identically, it's just that with the mouse it's possible to place the moving block exactly on top of the insertion marker (resulting in no movement or at most minimal movement of the existing stack blocks), whereas with keyboard moves the moving block is always and deliberately offset from the insertion marker.

I've therefore removed the aforementioned checks to this comment, since they won't be applicable unless we change mouse dragging behaviour in core (in which case keyboard move behaviour should be changed—and probably will be changed automatically—to match).

Issue #696, opened 2025-08-21
Move tests: constrained movement (of connected block)


cpcallen:
Test constrained movement of:

* [x] Simple statement block (with no statement input)
* [x] E-shaped statement block (with multiple unconnected statement inputs)
* [x] Simple reporter block (only output)
* [x] Unary value expression (with a single unconnected input)
* [x] N-ary value expression, n>1 (with multiple unconnected inputs)

when moving:
 
* [x] Left/right (should visit all possible candidate connection)
* [ ] ~~Up/down (should only visit candidate connections that could be visited by navigating up/down~~ [Moved to #712]
and verify that:

* [x] Statement blocks connect using all available statement connections (next/prev and statement inputs), in the same order they would if dragged using the mouse
* [x] ~~Value blocks connect only using the output connection~~
  * ~~N.B.: value row heal may reconnect a value block being displaced, but this is not the same as a moving value block attempting to connect via one of its inputs to a value block with an unconnected output.~~
  * Uh oh: turns out that when dragged with mouse, value blocks can connect using their inputs as well as their output.  🤔  I had thought this might be the case only for geras/thrasos, or only for value blocks with a single input (the usual rule for row healing upon delete) but it seems to apply to all value inputs in all three renderers.
* [x] Value blocks can connect via their output connection or any of their empty input connections, in the same order they would if using the mouse

cpcallen:
Hmm.  I realise that the exact desired behaviour of up/down movement of value blocks is not well defined, in part because up/down navigation seems to behave in unexpected ways.  I'm going to file a separate bug for those tests, to be completed once the expected behaviour has been clarified.

PR #706, opened 2025-08-28
refactor(tests): Refactor move tests to simplify `moveTest` helper

cpcallen:
Two tweaks to PR #704 in preparation for adding more move tests:

* Add a `// @ts-ignore` to suppress an Unintended Comparison error that can occur when `PAUSE_TIME` is set to a value other than `0`.
* Modify the `moveTest` helper: by assuming that the moving block always finishes at the same location it starts, it's possible to considerably simplify the tests that call it.
  * Tests to verify correct reconnection after end of move (e.g. healing) will be done in a separate suite as part of #699.

Part of #696.

### DIFF
--- a/test/webdriverio/test/test_setup.ts
+++ b/test/webdriverio/test/test_setup.ts
@@ -558,6 +558,7 @@ export async function sendKeyAndWait(
   keys: string | string[],
   times = 1,
 ) {
+  // @ts-ignore: Unintentional comparison error
   if (PAUSE_TIME === 0) {
     // Send all keys in one call if no pauses needed.
     keys = Array(times).fill(keys).flat();
--- a/test/webdriverio/test/test_setup.ts
+++ b/test/webdriverio/test/test_setup.ts
@@ -558,6 +558,7 @@ export async function sendKeyAndWait(
   keys: string | string[],
   times = 1,
 ) {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore: Unintentional comparison error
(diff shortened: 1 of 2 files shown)


test/webdriverio/test/test_setup.ts
```diff
@@ -558,6 +558,7 @@ export async function sendKeyAndWait(
   keys: string | string[],
   times = 1,
 ) {
+  // @ts-ignore: Unintentional comparison error
```
RoboErikG, line 562: Instead of this I think you can just declare PAUSE_TIME as the number type:
export const PAUSE_TIME: number = 0;

That should prevent its type from getting set to the literal value and avoid the unintentional comparison error when it's not 0.

cpcallen: Alas no: if I do that then eslint complains:
```
  46:14  error  Type number trivially inferred from a number literal, remove type annotation  @typescript-eslint/no-inferrable-types
```
I could disable eslint to allow the declaration, but it seems more straightforward to disable it at the point the ts/lint errors actually occur:
```JS
  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore: Unintentional comparison error
  if (PAUSE_TIME === 0) {
    // …
```

So many well-intentioned rules making it so difficult to just change the value of a numeric constant!

cpcallen:
OK, I had to re-run the flaky test a few too many times for my liking but I believe this is now ready to go, pending review.

cpcallen:
Superseded by #710.

PR #708, opened 2025-08-28
release: release version 3.0.1.

gonfunko:
This PR merges the rc/v3.0.0 branch back into main after bumping the Blockly dependency to 12.3.0 and this package's version to 3.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "@types/p5": "^1.7.6",
     "@typescript-eslint/eslint-plugin": "^6.7.2",
     "@typescript-eslint/parser": "^6.7.2",
-    "blockly": "^12.3.0",
+    "blockly": "12.3.0",
     "chai": "^5.2.0",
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):

package.json
```diff
@@ -59,7 +59,7 @@
     "@types/p5": "^1.7.6",
     "@typescript-eslint/eslint-plugin": "^6.7.2",
     "@typescript-eslint/parser": "^6.7.2",
-    "blockly": "^12.2.0",
+    "blockly": "^12.3.0",
```
maribethb: you need to update the peerDependency version as well, if this wasn't done in the released version of 3.0.0 then we need to do a patch to update it. the peer dependency is what matters for users of the plugin. bumping the devDependency version is not actually a breaking change.

gonfunko: Whoops! Done, thank you for flagging. I also sent a CL to update the docs to this effect.

Issue #709, opened 2025-08-29
`Uncaught Error: Block not present in workspace's list of top-most blocks` when moving unary value expression block (geras/thrasos)


cpcallen:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

When doing a constrained move of a value block with one input and one output, a "Block not present in workspace's list of top-most blocks" error is thrown when attempting to move out of a connection candidate associated with a shadow block.

### Reproduction steps

1. Load test playground (any scenario, but default simple circle is perfect).
2. Use the renderer drop-down to select geras or thrasos (*not* zelos).
3. Use mouse to drag all blocks out of statement input of top-level `p5_draw` ("draw") block.
4. Attach a stack of three `text_print` ("print") blocks to the draw block.
5. To the first print block, attach a `text_join` ("create text with").
6. Use join block's mutator to remove the second input, leaving a block with a single, unconnected value input.  Close the mutator.
7. Click on the join block to select it.
8. Press `m` to initiate move.
9. Press right arrow.  Observe that join block moves to second print statement.
11. Press right arrow again.  Observe a full-screen error pop up (via webpack dev server hot reload tooling), or error message in browser console.
    * If reproducing without webpack, observe that second move causes insertion marker to disappear without block moving.

### Stack trace

```shell
blockly_compressed.js:1326 Uncaught Error: Block not present in workspace's list of top-most blocks.
removeTopBlock	@	blockly_compressed.js:1326
removeTopBlock	@	blockly_compressed.js:1383
setParent	@	blockly_compressed.js:983
setParent	@	blockly_compressed.js:1044
connect_	@	blockly_compressed.js:935
connect_	@	blockly_compressed.js:962
connect	@	blockly_compressed.js:937
previewMarker	@	blockly_compressed.js:1530
previewConnection	@	blockly_compressed.js:1529
updateConnectionPreview	@	blockly_compressed.js:1025
drag	@	blockly_compressed.js:1024
drag	@	keyboard_drag_strategy.ts:59
drag	@	blockly_compressed.js:1074
moveDraggable	@	blockly_compressed.js:1581
onDrag	@	blockly_compressed.js:1580
moveConstrained	@	mover.ts:270
callback	@	move.ts:72
onKeyDown	@	blockly_compressed.js:589
globalShortcutHandler$$module$build$src$core$common	@	blockly_compressed.js:117
f	@	blockly_compressed.js:70
```

### Screenshots

[...]

Issue #707, opened 2025-08-28
Constrained move of value block visit order wrong when external inputs present (zelos only)


cpcallen:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

A constrained move of a value block, using left/right arrow keys to visit all possible connection candidates in sequence, is expected to visit compatible connection candidates in the same order that they would be visited if navigating left/right (assuming there was something focusable like a shadow block connected there already).

This appears to work correctly when using the geras and thrasos renderers, and works correctly in _most_ cases when using the zelos renderer, but if using zelos, the visitation order is wrong when a block with external inputs (e.g. `text_join`) is encountered.

Note that while zelos is typically used with block definitions that mainly have inline inputs, it does supports external inputs, rendering the connections as a vertical stack of inputs rather than the more typical  horizontal row.

### Reproduction steps

1. Use zelos.
2. Add a block with multiple value inputs, and elsewhere a value block that can be connected to those inputs.
3. Select the value block.
4. Use constrained keyboard move (`m` + right arrow) to move the value block to and past the block with multiple value inputs.
5. Observe that the visitation order is correct if the block with inputs has them inline inputs, but not if they are external.

### Screenshots

Constrained move, visiting a `text_join` block with inline inputs and another of the same type with external inputs (zelos renderer):

https://github.com/user-attachments/assets/b22c2455-37cf-46c0-964b-b5fb72e5067e

Note inconsistent visitation order between the two `text_join` blocks.  (N.B. the shadow block does not affect visitation order.)

Constrained move, same block configuration, geras renderer:

https://github.com/user-attachments/assets/d65dc0fc-ba38-486f-acfa-9896043257d2

Note move order is consistent between the two `text_join` blocks (and also consistent with navigation order).

gonfunko:
This was fixed in https://github.com/RaspberryPiFoundation/blockly/pull/9593

PR #710, opened 2025-08-29
test(`Mover`): Add test for moving value blocks right/left

cpcallen:
More work on move tests, including some tweaks to the stack block move test code added by PR #704:

* Add a `// @ts-ignore` to suppress an Unintended Comparison error that can occur when `PAUSE_TIME` is set to a value other than `0`—and then add an `// eslint-disable-next-line` to suppress a lint error caused by use of `@ts-ignore`.
* Modify the `moveTest` helper: by assuming that the moving block always finishes at the same location it starts, it's possible to considerably simplify the tests that call it.
  * Tests to verify correct reconnection after end of move (e.g. healing) will be done in a separate suite as part of #699.

as well as new work:

* Create `moveValueTestBlocks`, a set of test blocks for testing correct behaviour of constrained moves of value blocks.
* Add tests for moving the following things right and left:
  1. A simple reporter block (value block with one field and no inputs—in this case a `text` block).
  2. A unary expression: a value block with one input and one output.
  3. A two-block row, formed by connecting 1 to 2.
* Refactor both existing stack block and new value block move tests to make better use of nested suites.
  * For the value block tests, add an additional layer of nested suites so that these tests will be run in each of geras, thrasos and zelos, since some of the move code behaves differently depending on which renderer is used.
* Some code to skip certain tests that fail due to #707 and #709.
* Remove a debugging `console.log` from `test_setup.ts`.

Part of #696.
Supersedes #706.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -114,6 +114,9 @@
                 <option value="moveStatementTestBlocks">
                   move statement test blocks
                 </option>
+                <option value="moveValueTestBlocks">
+                  move value test blocks
+                </option>
                 <option value="comments">comments</option>
               </select>
             </div>
--- a/test/loadTestBlocks.js
+++ b/test/loadTestBlocks.js
@@ -1000,10 +1000,19 @@ const moveValueTestBlocks = {
         },
       },
       {
+        'type': 'text',
+        'id': 'unattached',
+        'x': 75,
+        'y': 200,
+        'fields': {
+          'TEXT': 'unattached value',
+        },
+      },
+      {
         'type': 'p5_draw',
         'id': 'p5_draw',
         'x': 75,
-        'y': 200,
+        'y': 260,
         'deletable': false,
         'inputs': {
           'STATEMENTS': {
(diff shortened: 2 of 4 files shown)


cpcallen:
It may be useful when reviewing to refer to this screenshot of the `moveValueTestBlocks` introduced in d432ade and tweaked in eeb5bf8 and 07028e5:

Screenshot 2025-08-29 at 21 45 10

RoboErikG (approved):
Overall LGTM. Approving to unblock but still needs the update to PAUSE_TIME.

cpcallen:
> Overall LGTM. Approving to unblock but still needs the update to PAUSE_TIME.

Hmm.  I am not sure what you mean.

* In relation to the discussion about tsc / eslint errors from comparing `PAUSE_TIME === 0`, this PR modifies the comparison in `test_setup.ts` to read:
  ```JS
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore: Unintentional comparison error
    if (PAUSE_TIME === 0) { // …
  ```
  which appears to mollify both tsc and eslint, for `PAUSE_TIME` values both zero and non-zero.

* I also considered whether I might have accidentally included a debugging change of `const PAUSE_TIME = 500;` in the PR but I'm pretty sure I did not.

cpcallen:
OK, although I'm only 60% confident that I have not overlooked something related to `PAUSE_TIME` I'm 95% confident that this PR does not break something as a result of such an oversight, so I'm going to go ahead and squash it.

PR #711, opened 2025-09-03
test(`Mover`): Add test for moving complex expression blocks right/left move tests

### LINKED ISSUE #696: Move tests: constrained movement (of connected block)

cpcallen:
More work on move tests, including some tweaks to the stack block move test code added by PR #710:

- Introduce a new `appendBlock` test helper, to make it easy to attach several different blocks in sequence to the same test block layout.
- Refactor the existing statement and value block move tests to use `appendBlock`.

followed by completing the testing of (well-defined) constrained block movement by adding adding tests for moving  complex expression blocks (ones with multiple inputs) right and left.

Fixes #696.

### DIFF
--- a/test/loadTestBlocks.js
+++ b/test/loadTestBlocks.js
@@ -870,9 +870,10 @@ const moveStartTestBlocks = {
   },
 };
 
-// A bunch of statement blocks.  The blocks with IDs simple_mover and
-// complex_mover will be (constrained-)moved up, down, left and right
-// to verify that they visit all the expected candidate connections.
+// A bunch of statement blocks.  It is intended that statement blocks
+// to be moved can be attached to the next connection of p5_canvas,
+// and then be (constrained-)moved up, down, left and right to verify
+// that they visit all the expected candidate connections.
 const moveStatementTestBlocks = {
   'blocks': {
     'languageVersion': 0,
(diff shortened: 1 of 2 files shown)


RoboErikG (approved):
Changes LGTM. Is the failing test (Start moving value blocks) the flaky one?

cpcallen:
> Changes LGTM. Is the failing test (Start moving value blocks) the flaky one?

Yep.

PR #718, opened 2025-09-05
chore: Add the advanced playground to the demo/test page.

gonfunko:
This PR adds the advanced playground, with its debug options/extra blocks/block persistence across reloads, to the demo/test page available through `npm run start`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "devDependencies": {
         "@blockly/dev-scripts": "^4.0.8",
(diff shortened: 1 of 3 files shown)


maribethb (approved):
it's weird that this PR contains already-merged commits but otherwise lgtm.

I do wonder if we should run this by matt since they use this page for internal testing, but we can always revert it if they hate it.

nit: this should be a chore since it isn't a feature that should cause a new minor version of the plugin

gonfunko:
Fixed the extra commits and commit type, thanks!

maribethb (approved):

PR #720, opened 2025-09-05
chore: fix a bug in flaky move test

### LINKED ISSUE #719: flaky test: move start test on ubuntu
### Check for duplicates

### Description

This test fails often causing us to have to re-run CI or ignore the failure. https://github.com/google/blockly-keyboard-experimentation/blob/344c1bf97a896762b2cb5334f0b5fd34bd115346/test/webdriverio/test/move_test.ts#L100

### Reproduction steps

_No response_

### Stack trace

```shell
1) Move start tests
       Start moving value blocks:
     AssertionError: moving block should have no parent block
      at Context.<anonymous> (file:///home/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js:92:22)
      at fulfilled (file:///home/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js:9:58)
```

### Screenshots

_No response_

### Browsers

_No response_

maribethb:
Fixes #719 hopefully?

There was a bug in this test, it both opened the context menu and pressed 'm' to start move mode. But that wasn't enough to fix the flaky test.

I changed some of the assertions to be more specific. using `chai.assert.isNull` gives a better error message than `chai.assert` as chai knows more about what you're intending to check. the old error didn't tell you what the value was, which could help with debugging. the new message will say something like "expected null, got 123". That did give better error messages but still didn't fix anything (it wasn't supposed to)

Next I tried adding an assertion to make sure the problem wasn't with the context menu showing up. I added a check to make sure that the "Move" option was found, and that assertion never failed. Yet, the test flaked on.

I tried adding some pause time and still flaky, so removed it again.

I tried triggering the move with the 'm' key instead of the context menu and it was still flaky, so I reverted those commits.

Finally I added a "wait for exist" on the move icon, so that the test doesn't continue until move mode is actually activated. This required adding a class to the actual source code. All other changes are to the test code only.

At this point I don't know if it will solve the flaky test but even if it doesn't I think it's worth merging at this point as it does improve the tests.

### DIFF
--- a/test/webdriverio/test/move_test.ts
+++ b/test/webdriverio/test/move_test.ts
@@ -98,8 +98,6 @@ suite('Move start tests', function () {
   // When a move of a value block begins, it is expected that block
   // and all blocks connected to its inputs will be moved - i.e., that
   // a stack heal (really: unary operator chain heal) will NOT occur.
-  //
-  // Also tests initiating a move via the context menu.
   test('Start moving value blocks', async function () {
     for (let i = 1; i < 7; i++) {
       // Navigate to statement_<i>.
@@ -122,20 +120,8 @@ suite('Move start tests', function () {
         'selected block should have child value block',
       );
 
-      // Start move using context menu (using keyboard nav).
-      await sendKeyAndWait(this.browser, [Key.Ctrl, Key.Return]);
-
-      // Find how many times to press the down arrow
-      const index = (await contextMenuItems(this.browser)).findIndex(({text}) =>
-        text.includes('Move'),
-      );
-      chai.assert.isAbove(
-        index,
-        -1,
-        'expected Move to appear in context menu items',
-      );
-      await keyDown(this.browser, index);
-      await sendKeyAndWait(this.browser, Key.Return);
+      // Start move
+      await sendKeyAndWait(this.browser, 'm');
 
       // Check that the moving block has nothing connected it its
       // next/previous connections, and same thing connected to value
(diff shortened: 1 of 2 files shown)


maribethb:
FWIW I've been rerunning the tests and the latest failure was a *different* flaky test, not this one :)

test/webdriverio/test/move_test.ts
```diff
@@ -106,34 +109,48 @@ suite('Move start tests', function () {
...
+        index,
+        -1,
+        'expected Move to appear in context menu items',
       );
+      await keyDown(this.browser, index);
       await sendKeyAndWait(this.browser, Key.Return);
 
+      // Wait for the move icon to appear so we know we're in move mode.
+      const bubble = this.browser.$('.blocklyMoveIndicatorBubble');
+      await bubble.waitForExist();
```
cpcallen: If `bubble` isn't used again then why save it?  (Unless line length?)

```suggestion
      this.browser.$('.blocklyMoveIndicatorBubble').waitForExist();
```

cpcallen (approved):
Good catch on the unwanted `m`—and thanks for educating me about the existence of these various fancy assertions.  This is definitely an improvement.

test/webdriverio/test/move_test.ts
maribethb: Good point, I had remembered running into issues with doing this previously but it was when I needed to assert something about `bubble` which I'm not doing in this case. Fixed (had to add `await` so didn't commit your suggestion but fixed it locally and pushed)

BenHenning:
Apologies for the delay--thanks for reviewing this @cpcallen!

PR #726, opened 2025-09-12
chore: merge main into add-screen-reader branch

maribethb:
Merges main in to keep feature branch up to date and pull in new features & fixes.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -2,14 +2,16 @@
 name: Deploy static content to Pages
 
 on:
-  # Runs on pushes targeting the default branch
+  # Runs on pushes targeting the default and experimentation branches.
   push:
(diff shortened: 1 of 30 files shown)


.github/workflows/build.yml
```diff
@@ -6,14 +6,14 @@ on:
   pull_request:
   push:
     branches:
-      - add-screen-reader-support-experimental
+      - main
```
RoboErikG: Should the branch names get changed?

RoboErikG (approved):
It looks like the workflows were changed to run against main instead of the local branch. Should that piece be reverted?

maribethb:
> It looks like the workflows were changed to run against main instead of the local branch. Should that piece be reverted?

Thanks, this is fixed now that Christopher updated main again.

PR #730, opened 2025-09-15
chore: remove unused ts expect error

maribethb:
Note: this is against the `add-screen-reader-support-experimental` branch.

Removes a ts-expect-error declaration because in the corresponding branch of blockly, that property is public. The unnecessary expect causes tests to fail to compile.

### DIFF
--- a/test/webdriverio/test/move_test.ts
+++ b/test/webdriverio/test/move_test.ts
@@ -692,7 +692,6 @@ function getConnectionCandidate(
     const dragStrategy =
       block.getDragStrategy() as Blockly.dragging.BlockDragStrategy;
     if (!dragStrategy) throw new Error('no drag strategy');
-    // @ts-expect-error connectionCandidate is private.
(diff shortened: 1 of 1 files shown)


maribethb:
Unfortunately the rest of the test failures are because blockly is currently broken in this branch for renderers other than zelos. Prior to merging `main` into this branch, there were no tests in this repo that tested other renderers, and apparently the tests in core did not catch the fact that dragging blocks is broken (in non-zelos renderers), or perhaps the browser tests have not been run on this branch.

maribethb:
Filed https://github.com/google/blockly/issues/9367

I think it's useful to go ahead and merge this despite the remaining failures since the problem is in core, not this repo, tests will automatically pass again when that issue is fixed.

maribethb:
The problems with geras/thrasos were fixed in core so this now passes.

BenHenning (approved):

PR #745, opened 2025-09-25
fix: Improve node ID check robustness & WebdriverIO test CI stability

BenHenning:
https://github.com/google/blockly/pull/9384 revealed that one of the Block comment tests was a bit fragile in directly validating a specific node ID. Since node IDs can change if new elements on the page require a generated ID, the test could easily fail with changes to core Blockly.

This PR fixes the issue by using a regex check instead of a direct ID match. The 'comment_textarea' part is _probably_ sufficiently unique for the test to be considered correct (though it is possible for the wrong comment to be focused here, but it seems really unlikely with how simple the test is being set up and, if that actually happened, it's sort of incidentally still correct as there would be _some_ comment that has focus).

This PR also includes a number of attempts to improve test stability (particularly in CI):
- It introduces a CI-specific command that adds a 30 second Mocha test timeout to provide a lot more time for tests to fail (since CI tends to run on less performant machines than used for development).
- It introduces a synchronization between the Mocha timeout and the timeout used for WebdriverIO's `waitFor*` functions. Note that this actually will usually be 60 seconds regardless of the configured Mocha timeout due to the way the hook is set up. However, this is still far better than the 5 second default as at least 1 test seemed to fairly regularly fail against that limit.
- It adds a few missing 'wait' calls for helpers that are performing side effects which ensures better synchronization when using `PAUSE_TIME` (and possibly better stability when running in CI, too, though it's unclear of `pause()` actually does anything when the provided timeout is 0).
- See https://github.com/google/blockly-keyboard-experimentation/pull/745#issuecomment-3336166191 attempts to further stabilize OS X tests only to yield that there are actual test failures.

This will be merged into the screen reader experimental branch to unblock the core Blockly PR as well.

#746 has been filed to track the discovered test failures.

### DIFF
--- a/test/webdriverio/test/block_comment_test.ts
+++ b/test/webdriverio/test/block_comment_test.ts
@@ -23,7 +23,10 @@ suite('Block comment navigation', function () {
 
   // Clear the workspace and load start blocks.
   setup(async function () {
-    this.browser = await testSetup(testFileLocations.NAVIGATION_TEST_BLOCKS);
+    this.browser = await testSetup(
+      testFileLocations.NAVIGATION_TEST_BLOCKS,
+      this.timeout(),
+    );
     await this.browser.execute(() => {
       Blockly.getMainWorkspace()
         .getBlockById('p5_canvas_1')
--- a/test/webdriverio/test/block_comment_test.ts
+++ b/test/webdriverio/test/block_comment_test.ts
@@ -37,7 +37,7 @@ suite('Block comment navigation', function () {
     await sendKeyAndWait(this.browser, Key.Enter);
     const focusedNodeId = await getCurrentFocusNodeId(this.browser);
     chai.assert.isDefined(focusedNodeId);
-    chai.assert.match(focusedNodeId, /blockly\-\w+_comment_textarea_/);
+    chai.assert.match(focusedNodeId, /blockly-\w+_comment_textarea_/);
   });
 
   test('Escape from a focused comment focuses its bubble', async function () {
--- a/test/webdriverio/test/block_comment_test.ts
+++ b/test/webdriverio/test/block_comment_test.ts
@@ -36,7 +36,8 @@ suite('Block comment navigation', function () {
     await keyRight(this.browser);
     await sendKeyAndWait(this.browser, Key.Enter);
     const focusedNodeId = await getCurrentFocusNodeId(this.browser);
-    chai.assert.equal(focusedNodeId, 'blockly-2s_comment_textarea_');
+    chai.assert.isDefined(focusedNodeId);
+    chai.assert.match(focusedNodeId, /blockly\-\w+_comment_textarea_/);
   });
 
   test('Escape from a focused comment focuses its bubble', async function () {
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+            BLOCKLY_WDIO_PAUSE_TIME_MS: donotoverride
+          - os: macos-latest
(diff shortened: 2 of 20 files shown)


gonfunko (approved):
LGTM once the lint failure is resolved.

test/webdriverio/test/block_comment_test.ts
```diff
@@ -36,7 +36,8 @@ suite('Block comment navigation', function () {
     await keyRight(this.browser);
     await sendKeyAndWait(this.browser, Key.Enter);
     const focusedNodeId = await getCurrentFocusNodeId(this.browser);
-    chai.assert.equal(focusedNodeId, 'blockly-2s_comment_textarea_');
+    chai.assert.isDefined(focusedNodeId);
+    chai.assert.match(focusedNodeId, /blockly\-\w+_comment_textarea_/);
```
BenHenning:
```suggestion
    chai.assert.match(focusedNodeId, /blockly-\w+_comment_textarea_/);
```

Addresses lint failure.

BenHenning:
> LGTM once the lint failure is resolved.

Should be fixed now. Thanks for quick review!

BenHenning:
I've seen the following tests consistently fail in CI now, even with the robustness improvements:

```
  1) Move start tests
       Start moving value blocks:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)

  2) Scrolling into view
       Insert scrolls new block into view:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (file:///Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/scroll_test.js:75:25)
      at fulfilled (file:///Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/scroll_test.js:9:58)
```

This very likely suggests that these are real failures, though oddly they don't seem consistently reproducible in local OS X environments.

gonfunko (approved):

BenHenning:
I attempted to slow the OS X tests _way_ down by forcing a 100 millisecond `PAUSE_TIME` for them (which would improve reliability if this is a timing issue per the change in this PR that adds more missing waits) in ae92cf5abd2fc0a8f0db716c3ad01c8dd573a092. However, it still had the same failures which strongly suggests a logical issue that's specific to the CI OS X environment for unknown reasons.

BenHenning:
Worth noting there were 2 new flakes in the latest run:

```
  2) Value expression move tests
       using geras
         "before each" hook for "moving right":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
  3) Value expression move tests
       using thrasos
         "before each" hook for "moving right":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
```

I'm a bit surprised that's happening since there should be a 30 second timeout here, not a 10 second one. Either way I can't dig into this more today so going ahead and merging with the failures.

PR #747, opened 2025-09-25
fix: Fix the scenario picker with the advanced playground.

gonfunko:
This PR rolls forward the change that enabled the advanced playground on the demo page, and fixes the scenario picker. Now, when a scenario is chosen, the contents of the workspace are replaced with the blocks from that scenario. Reloading the page persists the workspace's configuration. The renderer and toolbox options have been removed from below the scenario picker in favor of configuring those options in the advanced playground.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "devDependencies": {
         "@blockly/dev-scripts": "^4.0.8",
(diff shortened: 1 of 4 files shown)


maribethb (approved):
one lint error to fix then lgtm

gonfunko:
Fixed the unused import and updated to the latest devtools that doesn't trap command enter.

maribethb (approved):

PR #750, opened 2025-09-29
fix: Fix bug that caused blocks to be inserted into an immovable stack.

### LINKED ISSUE #725: Insertion from Toolbox incorrectly replaces immovable blocks
### Check for duplicates

### Description

It's possible to insert a block from the flyout into an invalid next connection if the child block is immovable. This results in the original child block getting deleted. 

In this case, I would expect that we would the next available valid connection, such as below the child block, or fall back to the workspace if none can be found. Instead, you are allowed to insert into this spot, which causes the original block to get deleted.
Blockly.getMainWorkspace().getAllBlocks().forEach(b=>b.setMovable(false))

### Reproduction steps

1. Open the test page and set a child block to immovable. 
2. Select a parent of the immovable block.
3. Open the toolbox and press enter on a block- you can now place the block into a position you should not be able to (and cannot drag it into)

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
This PR fixes #725. In cases where a statement block was selected and a new statement block was inserted, the new block would be proposed for connection to the selected block. If the selected block's next block is immovable however, this is invalid, because to splice the new block into the middle of an immovable block stack would move the subsequent blocks as a side effect. This PR verifies that the next connection is actually capable of connecting to the new block, and keeps looking down the stack until it finds an eligible connection or gives up.

### DIFF
--- a/src/navigation.ts
+++ b/src/navigation.ts
@@ -414,19 +414,33 @@ export class Navigation {
       const inputType = movingHasOutput
         ? Blockly.inputs.inputTypes.VALUE
         : Blockly.inputs.inputTypes.STATEMENT;
-      const compatibleInputs = stationaryNode.inputList.filter(
-        (input) => input.type === inputType,
-      );
-      const input = compatibleInputs.length > 0 ? compatibleInputs[0] : null;
-      let connection = input?.connection;
-      if (connection) {
+      const compatibleConnections = stationaryNode.inputList
+        .filter((input) => input.type === inputType)
+        .map((input) => input.connection);
+      for (const connection of compatibleConnections) {
+        let targetConnection: Blockly.Connection | null | undefined =
+          connection;
         if (inputType === Blockly.inputs.inputTypes.STATEMENT) {
-          while (connection.targetBlock()?.nextConnection) {
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            connection = connection.targetBlock()!.nextConnection!;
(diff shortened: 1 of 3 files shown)


src/navigation.ts
```diff
@@ -429,9 +429,28 @@ export class Navigation {
...
-        return stationaryNode.nextConnection;
+        let nextConnection: Blockly.RenderedConnection | null =
+          stationaryNode.nextConnection;
+        while (nextConnection) {
+          if (
+            movingBlock.workspace.connectionChecker.canConnect(
+              movingBlock.previousConnection,
+              nextConnection,
+              true,
+              Infinity,
```
BenHenning, line 461: Could this just be omitted instead? It seems like an optional argument.

If it needs to logically be infinity then it may be worth dropping a quick comment explaining why since it seems slightly unexpected (at least to me).

BenHenning (approved):
Thanks @gonfunko! Makes sense to me!

src/navigation.ts
gonfunko: No, it's needed because otherwise if the blocks are too far it'll return false, because this method is normally used for checking whether a block being manually dragged can connect, and there's a limited required radius for that. Added a comment to clarify.

gonfunko:
@RoboErikG pointed out that this bug still existed for input blocks. I've added another commit to address that scenario. PTAL @BenHenning

BenHenning (approved):
Latest changes still make sense, thanks!

Issue #697, opened 2025-08-21
Move tests: unconstrained movement


cpcallen:
Test unconstrained movement of:

* [ ] Simple statement block (with no statement input).
* [ ] C-shaped statement block (with statement input).
* [ ] Non-stack C-shaped statement block (with only statement input).
* [ ] Simple reporter block (only output)
* [ ] Value expression block (with unconnected input)
* [X] Workspace comments
* [ ] Bubbles

when moving left / right / up / down, and verify that:

* [ ] Movement occurs in consistent steps in the specified direction (only).
* [ ] Moving any unconnected connection within range of a compatible connection results in a candidate connection.

Additionally, test that constrained movement of:

* [X] Workspace comments
* [ ] Bubbles

behaves identically to unconstrained movement.

PR #759, opened 2025-10-10
test: Unconstrained movement tests of statement and value blocks.

johnnesky:
Addresses some of the cases in: https://github.com/google/blockly-keyboard-experimentation/issues/697

Added unit tests for a simple and complex statement blocks and simple and "unary" value blocks, moving each block several steps in the cardinal directions via unconstrained movement.

To help initialize the tests, I added helper functions to get the position of an existing block or one of its inputs, and to move the target block to a position near the existing block. I also refactored the `moveTest` helper function to be usable as part of a larger unit test function.

### DIFF
--- a/test/webdriverio/test/move_test.ts
+++ b/test/webdriverio/test/move_test.ts
@@ -537,9 +537,549 @@ suite(`Value expression move tests`, function () {
   }
 });
 
+suite(`Statement expression unconstrained move tests`, function () {
+  // Increase timeout to 10s for this longer test (but disable
+  // timeouts if when non-zero PAUSE_TIME is used to watch tests) run.
+  this.timeout(PAUSE_TIME ? 0 : 10000);
+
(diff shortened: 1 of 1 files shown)


test/webdriverio/test/move_test.ts
```diff
@@ -537,9 +537,553 @@ suite(`Value expression move tests`, function () {
...
+    type: 'controls_if',
+    id: 'complex_mover',
+    extraState: {hasElse: true},
+  };
+
+  /**
+   * Expected connection candidates when moving STATEMENT_SIMPLE, after
+   * pressing ArrowDown n times.
+   */
+  const EXPECTED_STATEMENT_SIMPLE_DOWN = [
```
RoboErikG, line 563: How does this behave across renderers since the blocks will have different sizes and shapes in different renderers?

RoboErikG (approved):

PR #770, opened 2025-11-20
fix: CI tests on OS X

### LINKED ISSUE #746: OS X WebdriverIO tests have legitimate failures
### Check for duplicates

### Description

See #745. With a bunch of attempts at trying to stabilize the OS X tests, it seems two tests are legitimately failing.

### Reproduction steps

Not sure as I don't have a Mac. These tests seem to fail 100% of the time on CI from what I can observe now that other timing-related failures have been mitigated in #745.

### Stack trace

```shell
1) Move start tests
       Start moving value blocks:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)

  2) Scrolling into view
       Insert scrolls new block into view:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (file:///Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/scroll_test.js:75:25)
      at fulfilled (file:///Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/scroll_test.js:9:58)
```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
Fixes #746

This PR does a bunch of things, but most importantly it fixes repeatedly failing OS X tests that ran in CI. These fell into two categories:
- A single scrolling test that would always fail due to a dropped rendering frame in headless Chrome at an inopportune point in the test.
- Two `alert` tests that would reliably cause a _different_ test to timeout.

These two fixes required completely different solutions and underwent different investigations. A long discussion of the investigation can be found in the conversation thread in this PR.

### Fixing `alert` dialog problems

These failures were a bit odd. Two flyout tests verify callback logic which is set up in the test to open an alert dialog. These tests pass without an issue. However, when either one of these tests is run exactly before move start tests (both of them). The first move start test will pass, the second will reliably time out in CI on OS X. Sometimes WebdriverIO will include errors about focus issues, but this didn't always happen.

The best guess here is that there's an actual issue in Chrome and/or WebdriverIO specifically with alert dialogs that is putting the framework or browser into a bad state. Blockly shares no state across its tests due to page reloads and no cookies or local storage being leveraged, but forcing a session reload between tests (i.e. a full browser reopen) fixes the issue which seems to confirm that it's an issue in the framework environment and not Blockly. This seems like a reasonable fix.

Note that this is sort of discouraged due to session reloading increasing test runtime, but the CI completions do not seem much slower than before even with the extra reloads. Locally it takes 2-3x longer to run when using headless mode. When using non-headless mode it encounters actual failures, but fortunately this needs to be forced since the setup currently is such that session reloading will only ever happen with headless runs. It also seems likely that forcing session reloads like this will actually improve other behavioral inconsistencies or flakes in the tests, as well.

Session reloading has only been enabled for CI runs of tests since it shouldn't be needed locally, at least per observation (though it can be demonstrated locally using the command: `CI=true npm run test:ci`).

### Fixing rendering synchronization issues

The rendering issue came down to the fact that Blockly can get into a temporarily broken state if its rendering queue is not run. Since Blockly relies on `requestAnimationFrame` for this to occur, dropped frames (which are legitimate according to the spec and something we can observe specifically in Mac CI with headless Chrome) will lead to block positioning not updating when the test expects, leading to a failing test. This can be easily reproduced even on Linux locally by linking against a version of Blockly that forces immediate rendering (e.g. pretending JavaFX is enabled).

The fix was to force the tests to try and trigger a re-render of the workspace at select moments. During investigation it was easiest just to do this everywhere there's a moment to pause execution since that's effectively a time when the test actually wants to wait for things to 'settle', so this is what's actually changed in the PR. All previous moments of pausing execution now route through a specific utility function that also will try to force rendering of Blockly, bypassing any dropped frames that the browser might introduce and ensuring tests are a lot more stable. It seems possible (though it hasn't been proven) that other tests may have also been flaky due to this behavior.

It's important to note that this _might_ be a bug in core Blockly, but the reality is it seems nearly impossible to ever actually occur. Even if a browser dropped or delayed a rendering frame, eventual consistency should kick in since Blockly does deterministically render itself and thus will self-correct on the next render (which is fundamentally why the solution above works as well as it does).

Note also that this fix had a compatibility issue with the dialog fix above and required a special setting to disable synchronization since it's not valid in WebdriverIO to execute browser-side code while a dialog is open (and attempting to detect this results in a significant slowdown in CI runtime and a large increase in console warnings).

### Other test infrastructure improvements

As part of the investigation it was useful to try and see what the tests were doing when they failed. WebdriverIO supports this with a built-in `saveScreenshot` function. All test suites have been updated in this PR to automatically check for failures when they finish each test and, if there's a failure, the current screen state will be snapshotted and uploaded to a zip file (specific to the OS) in GitHub Actions.

This actually did help with determining the root rendering issue because the screenshot showed a correct block layout even though debug logs showed incorrect positions. At first this hinted toward inconsistencies between the data model and actual rendering, but continued investigation eventually revealed the rendering frame being dropped (and the screenshot function was likely forcing rendering and 'correcting' the broken state).

Failures will also be screenshotted for local tests and saved in `test/webdriverio/test/failures/` with a filename corresponding to the test's name. These are automatically ignored from being added to Git via `.gitignore`.

Finally, a couple other miscellaneous changes:
- One additional pause was added to the scroll test. This probably isn't needed, but more pauses generally help with stability.
- Ditto for `sendKeyAndWait` though this was more of a contract correction: if `PAUSE_TIME` was zero then it wasn't actually waiting anymore which seemed incorrect. That's been fixed.
- There are some new string casts that seem necessary after some NPM upgrades, and they're generally innocuous.

And one final note: this PR mainly ensures all tests generally pass, not that there are no flakes. From observation tests may still flake on occasion, but they can be restarted and generally expected to pass. The OS X tests ran 14 times in a row without a failure before running into the 6 hour GitHub Actions runtime limit (it seems they slowed down each subsequent run, perhaps due to some sort of memory or resource leak). It may be possible to add a "try 3 times" approach to the WebdriverIO tests in CI in the future to significantly reduce flakes and increase the trustworthiness of failures to be actual problems, but this PR does not address that.

### DIFF
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -305,10 +305,10 @@ suite.only('Toolbox and flyout test', function () {
       // First thing in the toolbox is the first button
       // Press Enter to activate it.
       await keyRight(this.browser);
-      // await sendKeyAndWait(this.browser, webdriverio.Key.Enter);
+      await sendKeyAndWait(this.browser, webdriverio.Key.Enter);
 
       // This errors if there is no alert present
-      // await this.browser.getAlertText();
+      await this.browser.getAlertText();
     });
 
     test.skip('callbackKey is activated with enter', async function () {
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -25,7 +25,7 @@ import {
   setSynchronizeCoreBlocklyRendering,
 } from './test_setup.js';
 
-suite('Toolbox and flyout test', function () {
+suite.only('Toolbox and flyout test', function () {
   // Disable timeouts when non-zero PAUSE_TIME is used to watch tests run.
   if (PAUSE_TIME) this.timeout(0);
 
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -302,6 +302,8 @@ suite('Toolbox and flyout test', function () {
       }
     });
     test('callbackkey is activated with enter', async function () {
+      // Rendering synchronization must be disabled since this test opens an
+      // alert dialog. See the function's documentation for more specifics.
       setSynchronizeCoreBlocklyRendering(false);
       await tabNavigateToToolbox(this.browser);
       await pause(this.browser);
@@ -316,6 +318,8 @@ suite('Toolbox and flyout test', function () {
     });
 
     test('callbackKey is activated with enter', async function () {
+      // Rendering synchronization must be disabled since this test opens an
+      // alert dialog. See the function's documentation for more specifics.
       setSynchronizeCoreBlocklyRendering(false);
       await tabNavigateToToolbox(this.browser);
       await pause(this.browser);
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -25,7 +25,7 @@ import {
   setSynchronizeCoreBlocklyRendering,
 } from './test_setup.js';
 
-suite.only('Toolbox and flyout test', function () {
+suite('Toolbox and flyout test', function () {
   // Disable timeouts when non-zero PAUSE_TIME is used to watch tests run.
   if (PAUSE_TIME) this.timeout(0);
 
@@ -39,7 +39,7 @@ suite.only('Toolbox and flyout test', function () {
     await checkForFailures(this.browser, this.currentTest!.title, this.currentTest?.state);
   });
 
-  test.skip('Tab navigating to toolbox should open flyout', async function () {
+  test('Tab navigating to toolbox should open flyout', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped).
     await tabNavigateForward(this.browser);
 
@@ -50,7 +50,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isTrue(flyoutIsOpen);
   });
 
-  test.skip('Tab navigating to flyout should auto-select first block', async function () {
+  test('Tab navigating to flyout should auto-select first block', async function () {
     // Three tabs should navigate to the flyout (initial element is skipped).
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -63,7 +63,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'if_block');
   });
 
-  test.skip('Tab navigating to toolbox then right arrow key should auto-select first block in flyout', async function () {
+  test('Tab navigating to toolbox then right arrow key should auto-select first block in flyout', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped). One
     // right arrow key should select a block on the flyout.
     await tabNavigateForward(this.browser);
@@ -77,7 +77,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'if_block');
   });
 
-  test.skip('Keyboard nav to different toolbox category should auto-select first block', async function () {
+  test('Keyboard nav to different toolbox category should auto-select first block', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped),
     // then keys for a different category with a tab to select the flyout.
     await tabNavigateForward(this.browser);
@@ -91,7 +91,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'text_block');
   });
 
-  test.skip('Keyboard nav to different toolbox category and block should select different block', async function () {
+  test('Keyboard nav to different toolbox category and block should select different block', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped),
     // then keys for a different category with a tab to select the flyout and
     // finally keys to select a different block.
@@ -107,7 +107,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'append_text_block');
   });
 
-  test.skip('Tab navigate away from toolbox restores focus to initial element', async function () {
+  test('Tab navigate away from toolbox restores focus to initial element', async function () {
     // Two tabs should navigate to the toolbox. One tab back should leave it.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -122,7 +122,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(activeElementId, 'focusableDiv');
   });
 
-  test.skip('Tab navigate away from toolbox closes flyout', async function () {
+  test('Tab navigate away from toolbox closes flyout', async function () {
     // Two tabs should navigate to the toolbox. One tab back should leave it.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -135,7 +135,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Tab navigate away from flyout to toolbox and away closes flyout', async function () {
+  test('Tab navigate away from flyout to toolbox and away closes flyout', async function () {
     // Three tabs should navigate to the flyout, and two tabs back should close.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -149,7 +149,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Tabbing to the workspace should close the flyout', async function () {
+  test('Tabbing to the workspace should close the flyout', async function () {
     await tabNavigateToWorkspace(this.browser);
     await idle(this.browser);
 
@@ -158,7 +158,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Tabbing to the workspace after selecting flyout block should close the flyout', async function () {
+  test('Tabbing to the workspace after selecting flyout block should close the flyout', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped),
     // then use right key to specifically navigate into the flyout before
     // navigating to the workspace.
@@ -174,7 +174,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Tabbing to the workspace after selecting flyout block via workspace toolbox shortcut should close the flyout', async function () {
+  test('Tabbing to the workspace after selecting flyout block via workspace toolbox shortcut should close the flyout', async function () {
     await tabNavigateToWorkspace(this.browser);
 
     await sendKeyAndWait(this.browser, 't');
@@ -187,7 +187,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Tabbing back from workspace should reopen the flyout', async function () {
+  test('Tabbing back from workspace should reopen the flyout', async function () {
     await tabNavigateToWorkspace(this.browser);
 
     await tabNavigateBackward(this.browser);
@@ -198,7 +198,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isTrue(flyoutIsOpen);
   });
 
-  test.skip('Navigation position in workspace should be retained when tabbing to flyout and back', async function () {
+  test('Navigation position in workspace should be retained when tabbing to flyout and back', async function () {
     // Navigate to the workspace and select a non-default block.
     await tabNavigateToWorkspace(this.browser);
 
@@ -214,7 +214,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'p5_draw_1');
   });
 
-  test.skip('Clicking outside Blockly with focused toolbox closes the flyout', async function () {
+  test('Clicking outside Blockly with focused toolbox closes the flyout', async function () {
     // Two tabs should navigate to the toolbox. One click to unfocus.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -227,7 +227,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Clicking outside Blockly with focused flyout closes the flyout', async function () {
+  test('Clicking outside Blockly with focused flyout closes the flyout', async function () {
     // Two tabs should navigate to the toolbox. One key to select the flyout.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -242,7 +242,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test.skip('Clicking on toolbox category focuses it and opens flyout', async function () {
+  test('Clicking on toolbox category focuses it and opens flyout', async function () {
     const elemId = await findToolboxCategoryIdByName(this.browser, 'Loops');
     const elem = this.browser.$(`#${elemId}`);
     await elem.click();
@@ -311,7 +311,7 @@ suite.only('Toolbox and flyout test', function () {
       await this.browser.getAlertText();
     });
 
-    test.skip('callbackKey is activated with enter', async function () {
+    test('callbackKey is activated with enter', async function () {
       setSynchronizeCoreBlocklyRendering(false);
       await tabNavigateToToolbox(this.browser);
       await idle(this.browser);
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -308,7 +308,7 @@ suite.only('Toolbox and flyout test', function () {
       await sendKeyAndWait(this.browser, webdriverio.Key.Enter);
 
       // This errors if there is no alert present
-      await this.browser.getAlertText();
+      // await this.browser.getAlertText();
     });
 
     test.skip('callbackKey is activated with enter', async function () {
--- a/test/webdriverio/test/flyout_test.ts
+++ b/test/webdriverio/test/flyout_test.ts
@@ -39,7 +39,7 @@ suite.only('Toolbox and flyout test', function () {
     await checkForFailures(this.browser, this.currentTest!.title, this.currentTest?.state);
   });
 
-  test('Tab navigating to toolbox should open flyout', async function () {
+  test.skip('Tab navigating to toolbox should open flyout', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped).
     await tabNavigateForward(this.browser);
 
@@ -50,7 +50,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isTrue(flyoutIsOpen);
   });
 
-  test('Tab navigating to flyout should auto-select first block', async function () {
+  test.skip('Tab navigating to flyout should auto-select first block', async function () {
     // Three tabs should navigate to the flyout (initial element is skipped).
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -63,7 +63,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'if_block');
   });
 
-  test('Tab navigating to toolbox then right arrow key should auto-select first block in flyout', async function () {
+  test.skip('Tab navigating to toolbox then right arrow key should auto-select first block in flyout', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped). One
     // right arrow key should select a block on the flyout.
     await tabNavigateForward(this.browser);
@@ -77,7 +77,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'if_block');
   });
 
-  test('Keyboard nav to different toolbox category should auto-select first block', async function () {
+  test.skip('Keyboard nav to different toolbox category should auto-select first block', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped),
     // then keys for a different category with a tab to select the flyout.
     await tabNavigateForward(this.browser);
@@ -91,7 +91,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'text_block');
   });
 
-  test('Keyboard nav to different toolbox category and block should select different block', async function () {
+  test.skip('Keyboard nav to different toolbox category and block should select different block', async function () {
     // Two tabs should navigate to the toolbox (initial element is skipped),
     // then keys for a different category with a tab to select the flyout and
     // finally keys to select a different block.
@@ -107,7 +107,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(blockId, 'append_text_block');
   });
 
-  test('Tab navigate away from toolbox restores focus to initial element', async function () {
+  test.skip('Tab navigate away from toolbox restores focus to initial element', async function () {
     // Two tabs should navigate to the toolbox. One tab back should leave it.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -122,7 +122,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.strictEqual(activeElementId, 'focusableDiv');
   });
 
-  test('Tab navigate away from toolbox closes flyout', async function () {
+  test.skip('Tab navigate away from toolbox closes flyout', async function () {
     // Two tabs should navigate to the toolbox. One tab back should leave it.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
@@ -135,7 +135,7 @@ suite.only('Toolbox and flyout test', function () {
     chai.assert.isFalse(flyoutIsOpen);
   });
 
-  test('Tab navigate away from flyout to toolbox and away closes flyout', async function () {
+  test.skip('Tab navigate away from flyout to toolbox and away closes flyout', async function () {
     // Three tabs should navigate to the flyout, and two tabs back should close.
     await tabNavigateForward(this.browser);
     await tabNavigateForward(this.browser);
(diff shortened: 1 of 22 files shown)


BenHenning:
Looking at this scroll test first: `Insert scrolls new block into view`. It's failing with an error that the `controls_if` block isn't in the viewport. Comparing my local Linux run of the test vs. the CI run makes it pretty clear:

Linux local:
```
block bounds: { bottom: 1016, left: 53.34375, right: 213.34375, top: 904 }
viewport: { bottom: 1265.5, left: -10, right: 561, top: 814.5 }
```

Mac on CI:
```
block bounds: {
  bottom: 1882,
  left: 130.41592407226562,
  right: 290.4159240722656,
  top: 1770
}
viewport: { bottom: 1166.5, left: -10, right: 571, top: 705.5 }
```

The block is obviously way below the viewport for some reason. Also it's interesting to see that the viewport sizes are different which suggests discrepancies in test behaviors. Lack of cross-platform or cross-environment determinism will make this more difficult.

BenHenning:
Oh hey. It actually passed when using viewport instead of window size. Cool--I didn't expect that. I was guessing there could be some differences but I wasn't expecting there to be enough of one to cause _that much_ of a coordinate discrepancy. I'm surprised but I don't care about the why quite enough to dig on it.

BenHenning:
For reference, here are the results running the changed test on each platform.

Linux (local):

```
block bounds: { top: 904, bottom: 1016, left: 53.34375, right: 213.34375 }
viewport: { top: 740, bottom: 1340, left: -10, right: 570 }
```

Mac (CI):

```
block bounds: {
  top: 904,
  bottom: 1016,
  left: 54.20796203613281,
  right: 214.2079620361328
}
viewport: { top: 740, bottom: 1340, left: -10, right: 571 }
```

They are basically identical now which is great. :D

Unfortunately other Mac failures are likely a different problem entirely. Will dig on those next.

BenHenning:
Aha--enabling BiDi caused a bunch of new failures! Interesting.

BenHenning:
Updating to latest Webdriver seems to fix a few more issues with BiDi and bring us back to basically where we were at before for failures (I think), plus a new failure affecting Linux (maybe only?).

Edit: Actually interestingly the previously fixed test is either failing again, or is failing when run in conjunction with other tests.

BenHenning:
Either it's failing again after the upgrade or, maybe more likely, it was flaky and incidentally passed earlier. Latest dimensions with the failure:

```
block bounds: {
  top: 1770,
  bottom: 1882,
  left: 130.41592407226562,
  right: 290.4159240722656
}
viewport: { top: 636, bottom: 1236, left: -10, right: 571 }
```

This is more or less what we saw earlier. It seems that the BiDi change is probably not needed (and we could downgrade back if I can find fixes for everything without it). The flake seems to be that, for some reason, the block is being put in the wrong place.

BenHenning:
Huh. This is surprisingly difficult to break. It seems _much_ more consistent now, but it obviously can still fail per the two repros above.

BenHenning:
It passed 50 times in a row. Okay...this might be really tricky actually. My current working theory is that the extra browser waits to perform the debugging are actually adding a pause that allows scrolling to stabilize, thus fixing the test.

BenHenning:
The extra pause _seems_ to fix it. Removing all debug logs successfully re-failed the test (though we did see a failure earlier w/o needing to remove the later logs in the test, but oh well).

BenHenning:
Attempting a 50x run of all tests except for the move tests since those time out. I think this might take a while. Will follow up with a rough estimate for completion if none fail.

Edit: Approximately 65-70 seconds per run so I'm guessing about an hour to run through 50 times without failure.

Edit 2: Though that's based on WebdriverIO's reported times. The runner actually took 2.5 minutes to fail so I think there's a lot of overhead not being reported here. That likely means this will take more than 2 hours to run through. I think GitHub has a 6 hour limit on runners so we should be below that.

BenHenning:
Interestingly running the whole suite actually caused the earlier test to fail again. Will try re-adding all of the logs and see if we can glean anything interesting.

BenHenning:
Digging into this a bit more with more log statements. It's very odd. The inserted block is being added somewhere completely nonsensical and I'm not sure why. It's position isn't anywhere near its parent.

Edit: latest logs:

```
just moved block bounds: { top: 832, bottom: 936, left: 0, right: 160 }
workspace scroll position before scroll: [ 10, -346 ]
workspace scroll position after scroll: [ 10, -65 ]
current focused node before insert: blockly-u
current focused node after insert: blockly-36
workspace scroll position after insert: [ 10, -636 ]
block bounds: {
  top: 1770,
  bottom: 1882,
  left: 130.41592407226562,
  right: 290.4159240722656
}
block's parent bounds: { top: 832, bottom: 936, left: 0, right: 160 } id: blockly-u
viewport: { top: 636, bottom: 1236, left: -10, right: 571 }
```

The inserted block's surround parent's bounds are outside where the block is. This is a completely broken insertion--the workspace is _definitely_ in a bad state here and it's completely unclear why. I'm actually curious if adding an up/down arrow action during the insertion might 'correct' its placement, but that hack should absolutely not be required. This might be catching a real bug.

BenHenning:
I added support for uploading screenshots of failing tests at their point of failure.

The plot thickens because the metrics do not match the rendering here:

Insert scrolls new block into view

BenHenning:
Logs from a Linux run of the failing test:

```
just moved block bounds: { bottom: 936, left: 0, right: 160, top: 832 }
workspace scroll position before scroll: [ 10, -485 ]
workspace scroll position after scroll: [ 10, -65 ]
current focused node before insert: blockly-u
current focused node after insert: blockly-36
workspace scroll position after insert: [ 10, -809.5 ]
block bounds: { bottom: 1016, left: 53.34375, right: 213.34375, top: 904 }
block position: { x: 53.34375, y: 904 }
block's parent bounds: { bottom: 1040, left: 0, right: 160, top: 832 } id: blockly-u
viewport: { bottom: 1270.5, left: -10, right: 570, top: 809.5 }
```

Edit: failure from latest OS X run (without BiDi):

```
just moved block bounds: { bottom: 936, left: 0, right: 160, top: 832 }
workspace scroll position before scroll: [ 10, -485 ]
workspace scroll position after scroll: [ 10, -65 ]
current focused node before insert: blockly-u
current focused node after insert: blockly-36
workspace scroll position after insert: [ 10, -705.5 ]
block bounds: {
  bottom: 1882,
  left: 130.41592407226562,
  right: 290.4159240722656,
  top: 1770
}
block position: { x: 130.41592407226562, y: 1770 }
block's parent bounds: { bottom: 936, left: 0, right: 160, top: 832 } id: blockly-u
viewport: { bottom: 1166.5, left: -10, right: 571, top: 705.5 }
```

BenHenning:
Latest run (with relative position logs).

Edit: added additional details.

Linux:

```
just moved block bounds: { bottom: 936, left: 0, right: 160, top: 832 }
workspace scroll position before scroll: [ 10, -485 ]
workspace scroll position after scroll: [ 10, -65 ]
current focused node before insert: blockly-u
current focused node after insert: blockly-36
workspace scroll position after insert: [ 10, -809.5 ]
block bounds: { bottom: 1016, left: 53.34375, right: 213.34375, top: 904 }
block position: { x: 53.34375, y: 904 } relative: { x: 0, y: 48 } null null translate(0, 48) null
block's parent bounds: { bottom: 1040, left: 0, right: 160, top: 832 } id: blockly-u
viewport: { bottom: 1270.5, left: -10, right: 570, top: 809.5 }
matched blocks to controls_if: [ 'blockly-36' ]
```

OS X:

```
just moved block bounds: { bottom: 936, left: 0, right: 160, top: 832 }
workspace scroll position before scroll: [ 10, -485 ]
workspace scroll position after scroll: [ 10, -65 ]
current focused node before insert: blockly-u
current focused node after insert: blockly-36
workspace scroll position after insert: [ 10, -705.5 ]
block bounds: {
  bottom: 1882,
  left: 130.41592407226562,
  right: 290.4159240722656,
  top: 1770
}
block position: { x: 130.41592407226562, y: 1770 } relative: { x: 76.20796203613281, y: 914 } null null translate(76.20796203613281, 914) null
block's parent bounds: { bottom: 936, left: 0, right: 160, top: 832 } id: blockly-u
viewport: { bottom: 1166.5, left: -10, right: 571, top: 705.5 }
matched blocks to controls_if: [ 'blockly-36' ]
```

BenHenning:
Still pretty stuck on this issue. It seems that the newly inserted block's position is very far down (its actual translation), but it's not being rendered like that per the screenshot. Part of me is wondering if more than one block is being matched with `getBlocksByType`.

BenHenning:
Interesting. I added a save screenshot line before the logs to see if perhaps there's some weird delay happening with the block layout that's not actually being represented in the screenshot (i.e. it happens after the test finishes). Adding this line fixes the test, yet the 5 second delay doesn't. Something very odd is happening here.

BenHenning:
Not sure if that actually had the desired effect, but I tried to force the user agent over to pretending to be JavaFX to try and force Blockly into immediate rendering mode.

Right now I am suspicious of `requestAnimationFrame`. My limited understanding of this function is that it's simply a request for the browser to perform some action at the next rendering frame. However, the browser may choose to not call this in certain cases. I'm wondering right now if headless mode is affecting how `requestAnimationFrame` behaves (oddly differently on Mac--not sure about that yet) which might explain why snapping a screenshot fixes the problem since it would force the browse to perform a full render frame (and thus resolve any pending animation frame requests, thus allowing Blockly to render).

BenHenning:
I have found success! I was trying to use `headless=new` from some rough searching, but this seems to be the default in Chrome now (as far as I can tell) so it expectedly did nothing.

I wanted to test my theory about `requestAnimationFrame` in more detail. I was having some difficultly overriding core Blockly's `render_management.ts` behaviors in plugin tests, so I linked against a version of Blockly where I essentially disabled everything `queueRender` does. Voila--I get the _exact_ same failure for this test when run locally. I think this essentially confirms the theory. It doesn't explain why it's happening specifically with this test and so consistently (to that I can only guess), but it may actually explain some of the other instability being seen (possibly also on Linux, though I am also suspecting Chrome's rendering engine may simply behave differently on OS X--though the fact that can actually affect DOM behavior isn't great but alas).

BenHenning:
Okay, the plot thickens further. I don't have a deep understanding of Blockly rendering but I may need to develop it to make further progress here. Essentially from what I can tell the newly inserted block gets some bogus coordinates set if `requestAnimationFrame` isn't called as expected. This is especially odd because even forcing rendering with `Blockly.getMainWorkspace().render()` never corrects the position--the block remains forever displaced.

This can be very easily reproduced by updating core Blockly here:

https://github.com/RaspberryPiFoundation/blockly/blob/91ef951e92576a090aea7c675bafd37a93ac369a/core/render_management.ts#L48-L66

And replacing that with the following version:

```ts
export function queueRender(block: BlockSvg): Promise<void> {
  queueBlock(block);
  return Promise.resolve();
}
```

This simulates `requestAnimationFrame` not being called. This will break a few tests though that can be worked around by adding some force workspace rendering calls in select places in the plugin tests. Doing that still causes exactly the scroll test to continue to fail and exactly in the same way as in CI (this repros consistently on Linux with this setup).

My best guess until I can dig further is that rendering is ultimately non-deterministic and has the following effects:
- It somehow impacts block layouting.
- Block layouting likely feeds back into rendering or later logical decisions (probably because core Blockly sources block position from the SVG itself).
- Further attempts to re-render never fix the situation because the block itself got into a logically bad place due to a missed frame.

This almost certainly indicates a real bug in core Blockly. It's valid for browsers to delay frames which could, theoretically, result in Blockly logic executing on not-yet-updated rendering properties (which is ultimately what's happening in this test and breaking core Blockly state).

BenHenning:
It. Passes.

I realized when digging deeper into the drag logic that my stop-gap measure here of trying to force rendering wasn't actually synchronizing the rendering queue, it was only populating it. Forcing that extra bit actually causes the rendering to happen which _does_ correct the bad layout.

I'm happy to see that Blockly rendering and layouting does seem to be deterministic, though there's still some quirkiness happening to cause this situation in the first place. My rough (but slightly better than before) understanding of what's happening here:
- A user starts a drag event (as part of inserting a new block)
- The user finalizes their location decision (which has been correctly updating the new block's position all along), and commits the decision by ending the drag
- The process of ending the drag includes this long process which basically is as follows:
  - End drag ->
  - Hide the connection previewer ->
  - The insertion marker previewer hides the insertion marker ->
  - Hiding the insertion marker unplugs it from the target block ->
  - Block's `unplug` then calls `unplugFromStack` ->
  - `unplugFromStack` disconnects the block's previous connection ->
  - `disconnect` calls `disconnectInternal` which is overridden in `RenderedConnection` ->
  - `RenderedConnection.disconnectInternal` queues a render ->
  - When rendering happens (e.g. via `requestAnimationFrame`) each of the queue blocks (which seems to come out to all of them) go through a `renderEfficiently` pass ->
  - `renderEfficiently` eventually calls down to `tightenEfficiently` which calls `translate` to correct all block positions relative to their (now finalized) connections
- When `requestAnimationFrame` doesn't call the callback then the tighten passes never happen and block positions end up being completely wrong

This all amounts to forcing a render pass to not actually being a bad strategy when 'idling' between steps in the tests. However this is a possible real bug because of the fact that `requestAnimationFrame` isn't guaranteed to be called. It's very unlikely to ever actually happen in a modern browser, but it does mean the Blockly's immediate rendering mode isn't working correctly (which subsequently means its JavaFX support is likely broken).

I will need to file a core Blockly issue to track this but I think we probably want a way to correctly handle immediate rendering and also be able to enable it for more than just JavaFX (or at least have ways to force rendering that aren't hacky workarounds). This will require considerations from the rest of the team, though.

BenHenning:
The elephant in the room is that I have had the move tests disabled this entire investigation and have no clue if they have the same root cause. I'll try to enable those and see how they're looking, plus I want to try and undo some of the extra changes on this branch and then perform a long 50x run to verify stability.

Some tests are failing but they may be new failures from the WDIO upgrade. I'll see if they still happen when undoing some of the changes (and reenable Ubuntu tests since I think those were also failing with similar errors).

BenHenning:
Unfortunately the move tests don't fail when set up with the local hack to simulate the CI breakage, so they may be caused by an unrelated thing. We still see timeout failures occasionally:

```
  1) Statement move tests
       "before each" hook for "moving down":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
  2) Value expression move tests
       using geras
         "before each" hook for "moving right":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
  3) Value expression move tests
       using thrasos
         "before each" hook for "moving right":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
  4) Mutator navigation
       "before each" hook for "Enter opens mutator":
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/mutator_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
```

These are what we see with move tests so I see no reason to not reenable those and continue digging on the timeout side of things (which is shaping up to be caused by an entirely different thing, anyway).

BenHenning:
Failure this time:

```
  1) Move start tests
       Start moving value blocks:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/blockly-keyboard-experimentation/blockly-keyboard-experimentation/main/test/webdriverio/test/dist/move_test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7)
```

Interesting remarks:
- The move tests consistently time out which is why I disabled them for the previous investigation. They don't seem dependent on overall test runtime, they just outright fail (or, more correctly, the move start test).
- The most recent run was 2 minutes and 20 seconds for build & run (non-WDIO tests are disabled). The move test was the only failure this time.
- The previous run (when move tests were disabled) had 4 failures and the overall runtime was 5 minutes and 31 seconds.

It seems very clear that the other flakes observed above had something to do with the CI environment itself. Maybe it was a high-load machine or perhaps something about the environment triggered a non-deterministic slowdown in the tests that caused failures.

These are very likely environmental flakes and the correct mitigation is to increase the timeout thresholds but I don't want to do that until after the move test is actually passing.

FWIW I haven't seen those other failures as far as I can remember on this thread, but I have seen transient failures that looked like those in other PRs so I don't think these are super uncommon occurrences.

BenHenning:
The test passes in isolation so it's apparently not 100% deterministic, but perhaps it fails with a specific set of tests running before it?

BenHenning:
Ah I just re-realized the interesting error logs that crop up when this test fails:

```
2025-12-05T01:47:50.697Z WARN webdriver: WebDriverError: javascript error: nothing focused
  (Session info: chrome=142.0.7444.176) when running "execute/sync" with method "POST" and args "() => {
        var _a, _b, _c, _d, _e, _f, _g;
        const focused = Blockly.getFocusManager().getFocusedNode();
        if (!focused)
            throw new Error('nothing focused');
        if (!(focused instanceof Blockly.BlockSvg)) {
            throw new TypeError('focused node is not a BlockSvg');
        }
        const block = focused; // Inferred as BlockSvg.
        const parent = block === null || block === void 0 ? void 0 : block.getParent();
        // N.B. explicitly converts any undefined values to null because
        // browser.execute does this implicitly and so otherwise this
        // function would return values that were not compliant with its
        // own inferred type signature!
        return {
            parentId: (_a = parent === null || parent === void 0 ? void 0 : parent.id) !== null && _a !== void 0 ? _a : null,
            parentIndex: (_b = parent === null || parent === void 0 ? void 0 : parent.getConnections_(true).findIndex((conn) => conn.targetBlock() === block)) !== null && _b !== void 0 ? _b : null,
            nextId: (_d = (_c = block === null || block === void 0 ? void 0 : block.getNextBlock()) === null || _c === void 0 ? void 0 : _c.id) !== null && _d !== void 0 ? _d : null,
            valueId: (_g = (_f = (_e = block === null || block === void 0 ? void 0 : block.inputList[0].connection) === null || _e === void 0 ? void 0 : _e.targetBlock()) === null || _f === void 0 ? void 0 : _f.id) !== null && _g !== void 0 ? _g : null,
        };
    }"
2025-12-05T01:47:51.211Z WARN webdriver: WebDriverError: javascript error: nothing focused
  (Session info: chrome=142.0.7444.176) when running "execute/sync" with method "POST" and args "() => {
        var _a, _b, _c, _d, _e, _f, _g;
        const focused = Blockly.getFocusManager().getFocusedNode();
        if (!focused)
            throw new Error('nothing focused');
        if (!(focused instanceof Blockly.BlockSvg)) {
            throw new TypeError('focused node is not a BlockSvg');
        }
        const block = focused; // Inferred as BlockSvg.
        const parent = block === null || block === void 0 ? void 0 : block.getParent();
        // N.B. explicitly converts any undefined values to null because
        // browser.execute does this implicitly and so otherwise this
        // function would return values that were not compliant with its
        // own inferred type signature!
        return {
            parentId: (_a = parent === null || parent === void 0 ? void 0 : parent.id) !== null && _a !== void 0 ? _a : null,
            parentIndex: (_b = parent === null || parent === void 0 ? void 0 : parent.getConnections_(true).findIndex((conn) => conn.targetBlock() === block)) !== null && _b !== void 0 ? _b : null,
            nextId: (_d = (_c = block === null || block === void 0 ? void 0 : block.getNextBlock()) === null || _c === void 0 ? void 0 : _c.id) !== null && _d !== void 0 ? _d : null,
            valueId: (_g = (_f = (_e = block === null || block === void 0 ? void 0 : block.inputList[0].connection) === null || _e === void 0 ? void 0 : _e.targetBlock()) === null || _f === void 0 ? void 0 : _f.id) !== null && _g !== void 0 ? _g : null,
        };
    }"
```

BenHenning:
I've been able to partially isolate it. The move start test seems to consistently fail when run after the flyout/toolbox tests. None of the other suites seems to matter, but that one does.

BenHenning:
I've isolated it to the flyout tests opening an alert dialog and not closing it. For some reason the regular 'Start moving statement blocks' test needs to run and complete after the dialog has been opened in a previous test and before starting the 'Start moving value blocks' test begins in order to trigger the failure.

BenHenning:
I'm trying to reduce both of the move tests to just a single case rather than a loop and see if that still triggers the failure (in conjunction with the alert dialog).

This is a rather bizarre situation and it's not at all clear to me what's happening since the tests should be largely isolated between runs. My best guess is that webdriver or Chrome is being put into a semi-broken state here.

Edit: Isolating both move tests to 1 iteration seems to work. That means it's failing on the first iteration of the failing test and only 1 iteration is needed to get things into a broken state.

BenHenning:
The screenshot for the latest failure doesn't seem particularly interesting or noteworthy:

image

BenHenning:
Forcing session reloading fixes it. I may never fully understand the nuances that led to this particular issue, but I'm going to try reenabling the whole suite and also upping the timeout threshold for the Mac tests since there are a few that get close to 10s.

We also are going to have a bit slowdown with this change since there's a lot of overhead in reloading the entire browser for every suite, but it should tremendously improve stability especially on Mac.

BenHenning:
Looks like most timeouts are already 30s so only needed to make a few adjustments there.

BenHenning:
Looks like everything is passing, and honestly the runtime isn't terrible. I'm going to try kicking off 50 runs and see how far both Linux & Mac get to see how stable they are now.

BenHenning:
Funnily the Ubuntu tests failed immediately with a flake but the Mac tests lasted 14 runs before hitting the 6 hour actions timeout (interestingly they took longer and longer to run each iteration so presumably there's some sort of bad memory or resource leak happening here). I don't think we can assume this means the Mac tests are more stable than the Ubuntu ones, but 14x runs in a row without failing is a pretty great improvement over before.

BenHenning (commented):
Quick self-review pass.

BenHenning:
I want to try and fix the lint warnings before sending this out.

BenHenning (commented):
Self-reviewed lint fixes.

test/webdriverio/test/flyout_test.ts
```diff
@@ -291,8 +302,9 @@ suite('Toolbox and flyout test', function () {
       }
     });
     test('callbackkey is activated with enter', async function () {
+      setSynchronizeCoreBlocklyRendering(false);
```
maribethb, line 307: we could potentially change these tests to have the callback do something other than open the alert if that's going to be problematic.

but if not, i would add a comment above this line saying that it needs to be disabled since it opens an alert, otherwise i fear we will lose sight of when this needs to be set to false

test/webdriverio/test/test_setup.ts
```diff
@@ -119,8 +121,16 @@ export async function testSetup(
...
+  // Reset back to default state between tests.
+  synchronizeCoreBlocklyRendering = true;
+
   if (!driver) {
     driver = await driverSetup(wdioWaitTimeoutMs);
+  } else if (process.env.CI) {
+    // If running in CI force a session reload to ensure no browser state can
+    // leak across test suites (since this can sometimes cause complex combined
+    // failures in CI).
+    await driver.reloadSession();
```
maribethb, line 133: Is this run between each test or between suites? The PR description says between suites but this I think this is run between each test, though I may be mistaken. Not sure which one you intended so just wanted to double check.

test/webdriverio/test/test_setup.ts
```diff
@@ -192,6 +224,64 @@ export async function getSelectedBlockId(browser: WebdriverIO.Browser) {
...
+    // potential bugs when running in headless mode that can cause
+    // requestAnimationFrame to not call back (and cause state inconsistencies
+    // in block positions and sizes per #770).
+    await browser.execute(() => {
+      const workspace = Blockly.getMainWorkspace() as Blockly.WorkspaceSvg;
+      // Queue re-rendering all blocks.
+      workspace.render();
+      // Flush the rendering queue (this is a slight hack to leverage
+      // BlockSvg.render() directly blocking on rendering finishing).
+      const blocks = workspace.getTopBlocks();
```
maribethb, line 252: You could await the `finishQueuedRenders()` promise instead, which is designed for this purpose

maribethb (approved):

test/webdriverio/test/flyout_test.ts
BenHenning: Yeah, I don't think changing the test would be a problem but it's also fortunately not necessary (plus I'd imagine we may run into alert problems again in the future depending on how we address the create variables issue). Went ahead and added comments to clarify the need for this as suggested.

test/webdriverio/test/test_setup.ts
BenHenning: Great call. You're correct, it is per test. I've found some new data on how it behaves, too, and I've updated the PR description and the comment here accordingly.

test/webdriverio/test/test_setup.ts
BenHenning: I don't think that's true unless I'm mistaken. `finishQueuedRenders` waits on `requestAnimationFrame` to execute and the whole reason this issue exists is because the browser isn't running `requestAnimationFrame` callbacks (which means the render manager will never finish the promise and this will end up blocking indefinitely).

The pathway here, while very hacky, synchronously flushes the queue directly to ensure that it actual renders rather than hoping it renders per the `requestAnimationFrame` callback.

BenHenning (commented):
Self-reviewed comment fixes.

test/webdriverio/test/test_setup.ts
maribethb: Gotcha, that makes sense. I think `triggerQueuedRenders` might work then, as it cancels the animation frame and manually renders all the blocks in the queue. So it's possible this would potentially only re-render blocks that are "stuck" in the queue from the `requestAnimationFrame` callback not being called. But if that doesn't make sense to you then this hack is fine.

PR #771, opened 2025-11-21
Update test page link in README.md

kmcnaught:
To point at rasp pi deployment.


BenHenning (dismissed):
Ah, I fixed this in samples but forgot to check here. Thanks!

BenHenning (approved):
Thanks again for this fix @kmcnaught!

PR #773, opened 2025-12-03
fix: Support more complex 'simple reporter' blocks (experimental)

### LINKED ISSUE #9456: [Screen reader] Missing aria details for shadow blocks in some cases
### Check for duplicates

### Description

There is currently a [check for whether a block is a simple reporter with a full block field](https://github.com/google/blockly/blob/add-screen-reader-support-experimental/core/block_svg.ts#L235-L238), however, this code only checks the first field in the input. There are instances in MakeCode where it is not the first field that is the full block field. Please see the example below where the first field is just the text "melody" and the second field is the full block field.

Image

I think that at least for shadow blocks, the check needs to extend to all of the fields. This same change would also have to be applied to [`getFocusableElement` where the same check is made](https://github.com/google/blockly/blob/add-screen-reader-support-experimental/core/block_svg.ts#L1965-L1971).

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9456

This updates the 'enter' shortcut to work correctly for blocks that can have more than field but still be considered 'single input' (due to only having one mutable field). The base support for this change was done in the experimental branch in core Blockly via #9484.

### DIFF
--- a/src/actions/enter.ts
+++ b/src/actions/enter.ts
@@ -350,7 +350,7 @@ export class EnterAction {
    * @returns True if we showed the editor, false otherwise.
    */
   private tryShowFullBlockFieldEditor(block: Block): boolean {
-    if (block.isSimpleReporter()) {
+    if (block.isSimpleReporter(true, true)) {
       for (const input of block.inputList) {
         for (const field of input.fieldRow) {
           if (field.isClickable() && field.isFullBlockField()) {


BenHenning:
PTAL @gonfunko.

@microbit-robert if you can, could you please verify this finishes the functionality per https://github.com/RaspberryPiFoundation/blockly/issues/9456 that's needed for micro:bit?

microbit-robert:
> PTAL @gonfunko.
> 
> @microbit-robert if you can, could you please verify this finishes the functionality per [RaspberryPiFoundation/blockly#9456](https://github.com/RaspberryPiFoundation/blockly/issues/9456) that's needed for micro:bit?

I've tested this change locally and it looks good, thanks.

gonfunko (approved):

BenHenning:
Thanks @microbit-robert & @gonfunko. Going ahead and merging (which should auto-deploy the test page on GH Pages).

PR #777, opened 2025-12-08
fix: Don't loop around when navigating

gonfunko:
This PR fixes https://github.com/RaspberryPiFoundation/blockly/issues/9496 by disabling looping when using the keyboard to navigate through the workspace, and playing an error tone when the end of the navigable tree is reached.

### DIFF
--- a/src/actions/arrow_navigation.ts
+++ b/src/actions/arrow_navigation.ts
@@ -66,7 +66,10 @@ export class ArrowNavigation {
             if (
               !this.navigation.defaultWorkspaceCursorPositionIfNeeded(workspace)
             ) {
-              workspace.getCursor().in();
+              const newNode = workspace.getCursor().in();
+              if (!newNode) {
+                workspace.getAudioManager().beep(260);
+              }
             }
             isHandled = true;
           }
@@ -99,7 +102,10 @@ export class ArrowNavigation {
             if (
               !this.navigation.defaultWorkspaceCursorPositionIfNeeded(workspace)
             ) {
-              workspace.getCursor().out();
+              const newNode = workspace.getCursor().out();
+              if (!newNode) {
+                workspace.getAudioManager().beep(260);
+              }
             }
             isHandled = true;
           }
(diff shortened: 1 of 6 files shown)


microbit-robert:
FYI I noticed yesterday that there is still a call to create a [`LineCursor` in index.ts](https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/blob/main/src/index.ts#L63), but it's never used.

gonfunko:
Oh nice catch, went ahead and cleaned that up.

test/webdriverio/test/basic_test.ts
```diff
@@ -49,11 +49,11 @@ suite('Keyboard navigation on Blocks', function () {
     await tabNavigateToWorkspace(this.browser);
     await this.browser.pause(PAUSE_TIME);
 
-    await keyDown(this.browser, 22);
+    await keyDown(this.browser, 13);
```
maribethb, line 52: is this a keycode? please use Blockly.utils.KeyCodes for this

src/actions/arrow_navigation.ts
```diff
@@ -66,7 +66,10 @@ export class ArrowNavigation {
             if (
               !this.navigation.defaultWorkspaceCursorPositionIfNeeded(workspace)
             ) {
-              workspace.getCursor().in();
+              const newNode = workspace.getCursor().in();
+              if (!newNode) {
+                workspace.getAudioManager().beep(260);
```
maribethb, line 71: the magic number makes me twitch, but i think the ideal place for this to live is probably an `errorBeep` function or something in core, so it can stay for now ;p

maribethb (approved):

test/webdriverio/test/basic_test.ts
gonfunko: No, this is the number of times to trigger a keydown.

maribethb: ohhh that makes more sense, i wasn't interpreting keyDown as being the down arrow. cool, thanks

src/actions/arrow_navigation.ts
gonfunko: Yeah, agreed that would be nicer and should be dealt with when we finalize/merge back.

PR #788, opened 2026-02-04
fix: Export registerFlyoutCursor as a static method

### LINKED ISSUE #756: `registerFlyoutCursor` is not exported
### Check for duplicates

### Description

[This PR](https://github.com/google/blockly-keyboard-experimentation/pull/671) introduced a couple of functions to remove the remaining side-effects associated with loading the keyboard navigation plugin which need to be called by consuming applications before instantiating the plugin. However, the `registerFlyoutCursor` functions is not exported from index.ts and can't be called by consuming applications. The test files in this repo get around this via relative imports.

The result is a `LineCursor` instead of a `FlyoutCursor` in the flyout which skips over flyout labels and buttons. Note that this doesn't happen on the initial move from toolbox to flyout (the `LineCursor` will happily focus a button or label if it's the first item in the flyout), only on subsequent navigation in the flyout itself.

As @cpcallen notes on the linked PR, this is a breaking change, but no reference is made to this in the release notes for [the version](https://github.com/google/blockly-keyboard-experimentation/releases/tag/%40blockly%2Fkeyboard-navigation%403.0.1) that contains the change. 

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

microbit-robert:
Fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/756

Since v3.0.1 the flyout has been using a `LineCursor` which cannot navigate to flyout labels or buttons. If the button or label is the first item in the flyout, it is focussed by default when moving focus from the toolbox to the flyout, but cannot be navigated to otherwise.

I've added this as a static method following existing convention, but just `export {registerFlyoutCursor} from './flyout_cursor';` would work if it needs to be a separate function.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import * as Blockly from 'blockly/core';
 import {NavigationController} from './navigation_controller';
 import {enableBlocksOnDrag} from './disabled_blocks';
 import {registerHtmlToast} from './html_toast';
+import {registerFlyoutCursor} from './flyout_cursor';
 
 /** Plugin for keyboard navigation. */
 export class KeyboardNavigation {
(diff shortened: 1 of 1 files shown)


maribethb:
This is needed sooner because this is affecting a live issue in makecode

we should do a patch release of this after merging, we were already aiming to do one for core next week ish so we should test/release this as well

gonfunko (approved):

